PR #1, opened 2025-08-20
windows readme updates

rasmuswinter:
also adds more detail about the submodules, and removes the (ignored?) iznik-nuxt submodule


edwh (approved):
Yes, iznik-nuxt is redundant - that's an old codebase which is no longer used as of a few months back.

Thanks for the doc changes.

PR #100, opened 2026-04-06
feat(giftaid): add firstname and lastname columns to giftaid table

edwh:
## Summary

- Adds `firstname` (nullable) and `lastname` (nullable) columns to the `giftaid` table via Laravel migration
- Updates the `GiftAid` model with `getFirstname()`, `getLastname()`, and `hasValidNameSplit()` helper methods
- When dedicated columns are set they take precedence; otherwise the methods fall back to splitting `fullname` on the first space, preserving backwards compatibility with existing records
- Motivated by cultural naming issues: a single `fullname` field with a required space excludes mononyms and causes confusion for donors with East Asian (family-name-first) or other non-Western naming conventions

## Code Quality Review

- Migration is idempotent (Laravel schema builder checks column existence implicitly)
- Helper methods are pure functions with no side effects
- Backwards compatibility maintained — existing records with only `fullname` continue to work

## Test Plan

- `test_get_firstname_uses_dedicated_column_when_set` — dedicated columns override fullname split
- `test_get_firstname_falls_back_to_splitting_fullname` — fallback for existing records
- `test_get_firstname_with_multiple_word_lastname` — "Maria Garcia Lopez" splits on first space only
- `test_get_firstname_with_single_name` — mononym returns empty string for lastname (no crash)
- `test_has_valid_name_split_*` — three cases: dedicated columns set, spaced fullname, mononym without columns
- `test_firstname_lastname_stored_in_database` — migration columns are writable
- All 1455 Laravel tests pass

### DIFF
--- a/iznik-batch/app/Models/GiftAid.php
+++ b/iznik-batch/app/Models/GiftAid.php
@@ -25,6 +25,55 @@ class GiftAid extends Model
     ];
 
     /**
+     * Get first name: use dedicated firstname column if set, else split fullname on first space.
+     */
+    public function getFirstname(): string
+    {
+        if ($this->firstname !== null && $this->firstname !== '') {
+            return $this->firstname;
+        }
+
+        $spacePos = strpos($this->fullname ?? '', ' ');
+        if ($spacePos === false) {
+            return $this->fullname ?? '';
+        }
+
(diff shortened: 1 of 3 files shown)


edwh:
Superseded — migration included in #101 which now targets master directly.

PR #145, opened 2026-04-13
fix: clean up stale Spam review entries after ignore

edwh:
## Summary
- When a mod ignores a spam member on one group, the next refetch may return the same user keyed by a different membership ID (the next remaining row from the API)
- Without cleanup, the old entry persists as a ghost card that keeps reappearing in the review queue
- Adds stale-entry detection in the Spam collection fetch path: removes any prior entry for the same userid before inserting the refreshed one

Fixes Discourse #9518.167 ("Member review keeps coming back after clicking ignore").

## Test plan
- [x] New Vitest test: verifies stale entry under old key is removed when refetch returns same user under new key
- [x] All 11,515 Vitest tests pass locally
- [ ] Manual test: ignore a spam member who is in review on multiple groups, verify they don't reappear after page refresh

### DIFF
--- a/iznik-nuxt3/modtools/stores/member.js
+++ b/iznik-nuxt3/modtools/stores/member.js
@@ -160,6 +160,18 @@ export const useMemberStore = defineStore({
             })
           })
           Object.values(byUser).forEach((member) => {
+            // Remove any stale entry for this userid keyed under a different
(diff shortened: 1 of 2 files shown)


edwh:
Bad fix - this is client only.

PR #149, opened 2026-04-14
feat: chat-style reply overlay (mobile, tablet & desktop)

edwh:
Reply opens a chat-style pane as a full-screen overlay on every breakpoint (previously an inline form on desktop, a separate page on mobile). It matches the real Freegle chat; closing it returns you to exactly where you were. Replying from a single message lands you in the chat, but replying from a list page (browse/explore) sends without navigating, so you stay on the list and can scroll on to reply to more items.

| Mobile | Tablet | Desktop |
|---|---|---|
|  |  |  |

Logged-out (email field in the composer):


edwh:
@circleci rerun

PR #183, opened 2026-04-17
fix(messages): serialise pagination context for infinite scroll

edwh:
## Summary

Infinite scroll of approved messages capped at exactly 100 posts regardless of how far the user scrolled (Discourse [#9518.179](https://discourse.ilovefreegle.org/t/9518/179)).

### Root cause

`fetchMessagesMT` in `stores/message.js` passed the pagination context (`{ Date, ID }`) as an object to `$getv2`. `URLSearchParams` coerces objects via `toString()` → `"[object Object]"`, which the Go server's `json.Unmarshal` fails on silently. With `ctx = nil`, the server falls back to the newest `limit` messages — capped at 100 by the server's hard limit — so every "load more" returns the same 100 messages.

### Fix

Serialise the context with `JSON.stringify` before handing it to `$getv2`. This matches the server's expectation (the query handler calls `json.Unmarshal` on the string value).

### Tests

Added two unit tests in `tests/unit/stores/message.spec.js` covering:
- Object context is serialised to a JSON string before hitting the API.
- Null context is left as `null`.

## Test plan

- [ ] CI Vitest suite passes
- [ ] Manual verification on ModTools approved messages: scroll past 100 posts without hitting the cap
- [ ] Netlify preview for ModTools (when available)

### DIFF
--- a/iznik-server-go/spammers/spammers.go
+++ b/iznik-server-go/spammers/spammers.go
@@ -169,6 +169,18 @@ func PostSpammer(c *fiber.Ctx) error {
 	}
 
 	db := database.DBConn
+
+	// V1 parity (Spam::addSpammer): for PendingAdd, skip the REPLACE if a spam_users row
+	// already exists for this user so the original reporter's byuserid is preserved.
+	// This is the fix for Discourse #9589 (wrong-attribution bug).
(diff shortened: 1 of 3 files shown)


edwh:
Mis-created from wrong branch. Reopening from fix/modtools-infinite-scroll-pagination.

PR #198, opened 2026-04-17
fix(vectorsearch): whole-word keyword boost, not substring match

edwh:
## Summary
- Hybrid vector-search keyword boost now requires whole-word overlap instead of substring containment.
- Fixes Discourse 9585.11 (Neville): search for `table` was boosting "Portable gas heater" and "Adjustable office chair" above literal-"table" posts because both subjects contain `table` as a substring.

## Root cause
`VectorSearch` used `strings.Contains(subjectLower, queryWord)`, so any query word that happened to appear as a substring inside a subject word scored +0.3 weight — enough to outrank genuine semantic matches. Tokens like "portable", "adjustable", "tablecloth" all triggered the boost for a `table` query.

## Fix
Tokenise the subject with `GetWords` (the same tokeniser used for the query — strips punctuation, lowercases, removes stop-words) and intersect the two token sets. Only whole-word overlap contributes to the boost. This aligns the boost signal with the way users mean their query words.

## Test plan
- [x] New unit test `TestVectorSearchKeywordBoostWholeWord` in `embedding_vectorsearch_test.go` — differentiated cosine scores so the failure is unambiguous without the fix (`portable`/`adjustable` substring-matches would otherwise win).
- [x] Full Go suite: 1413/1413 pass via status container.
- [ ] After deploy: Neville to retest search for `table` at modtools search.

## Discourse thread
https://discourse.ilovefreegle.org/t/9585/11

### DIFF
--- a/iznik-server-go/message/vectorsearch.go
+++ b/iznik-server-go/message/vectorsearch.go
@@ -1,8 +1,6 @@
 package message
 
 import (
-	"strings"
-
 	"github.com/freegle/iznik-server-go/embedding"
 	"github.com/freegle/iznik-server-go/utils"
 )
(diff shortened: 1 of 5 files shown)


edwh:
Superseded by:

1. **Spammer fix** (commit 09d5b1d33 on this PR) — already merged to master via PR #182 (commit 5168c559d, same title and content).
2. **Vectorsearch whole-word keyword boost** (commit 3d49e3d68 on this PR) — now carried by branch `feature/hybrid-embedding-search` (commit 5330fbc72), which uses the same `GetWords` tokenisation. Hybrid will land on master with additional changes (subject/body split embedding) and bundles this fix.

Closing to avoid confusion. The Discourse #9585.11 regression remains tracked via the hybrid branch.

PR #201, opened 2026-04-17
fix(merge): guard undefined user.logins (Sentry)

edwh:
## Summary

Fixes Sentry [7384446789](https://freegle.sentry.io/issues/7384446789/) — `TypeError: Cannot read properties of undefined (reading 'forEach')` at `pages/merge.vue`, 4× in the last hour.

When the API returns a user with no `logins` field (can happen for users who've never successfully signed in), the inline `logins(user)` helper on the merge page did `user.logins.forEach(...)` with no guard, crashing the merge UI for support volunteers.

## Changes
- Extract to `composables/useMergeLogins.js` with `Array.isArray` guard → returns `''` when shape is wrong
- `merge.vue` computed refs now call `formatLogins(user)` instead of the local helper
- 8 Vitest tests covering undefined user, null user, empty array, every provider, dedup, unknown types, non-array fallback

## Test plan
- [x] 8 new unit tests pass
- [x] Full Vitest suite still green (11794/11794)
- [ ] CI build-test workflow passes

### DIFF
--- a/iznik-nuxt3/composables/useMergeLogins.js
+++ b/iznik-nuxt3/composables/useMergeLogins.js
@@ -1,0 +1,36 @@
+// Build a human-readable summary of how a user has previously signed in,
+// for display on the merge page. Defensively handles users whose `logins`
+// array is missing — the API occasionally returns the field omitted when
+// the user has never logged in, which previously crashed the page
+// (Sentry 7384446789).
+export function formatLogins(user) {
+  const ret = []
+
+  const logins = user?.logins
+  if (!Array.isArray(logins)) {
+    return ''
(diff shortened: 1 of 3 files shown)


edwh:
Closing — the premise was wrong. Users in production have always logged in at least once, so `user.logins` being undefined is an API gap, not a valid state to defend against. Real root cause: V2 Go `GetMerge` (iznik-server-go/merge/merge.go) omits the `logins` field that V1 `http/api/merge.php` includes (`$u1->getLogins(FALSE)`). Opening a new PR to restore V1 parity in the Go handler.

PR #208, opened 2026-04-18
fix(chitchat): prefetch window of threads to avoid serial fetch on slow networks

edwh:
## Summary

- Replace the 5-item initial prefetch with a 20-item rolling prefetch window that advances with `loadMore` and resets on `areaChange`.
- Each `NewsThread` does `await newsfeedStore.fetch(id)` in `<script setup>`, and the old code revealed one item per `loadMore` tick — so every item waited a full RTT before the next could even start.
- On high-latency connections (iOS cellular, ~300ms RTT) that serialised into ~30s for 100 items vs a few seconds on desktop fibre. Users gave up early, perceiving ChitChat as "very limited".
- Desktop behaviour is unchanged in practice (overlap was already fast).

Refs Discourse topic 9594 post 5 (Neville_Reid).

## Test plan

- [x] New failing Vitest test in `tests/unit/pages/chitchat/id.spec.js` asserting `loadMore` triggers ≥10 parallel prefetches including items at positions 5 and 10 ahead of `show.value`.
- [x] Test fails on the pre-fix code (expected 0 fetches ≥10) and passes after the fix (8/8 chitchat tests green).
- [x] Full Vitest suite ran via status container: counts went from 5 failed / 11806 passed → 5 failed / 11807 passed. The 5 failures are in `useSuppressException`, `admins` store, and `ModMessageButton` — pre-existing on master, not touched by this change.
- [x] ESLint clean on both changed files.

### DIFF
--- a/iznik-nuxt3/pages/chitchat/[[id]].vue
+++ b/iznik-nuxt3/pages/chitchat/[[id]].vue
@@ -471,6 +471,25 @@ function rendered() {
   }
 }
 
+// Prefetch a window of threads ahead of the scroll position so that
+// NewsThread's `await newsfeedStore.fetch(id)` in <script setup> doesn't
+// serialise one RTT per item. On high-latency connections (iOS cellular
+// ~300ms RTT) serial fetching makes the feed appear "very limited" as users
+// give up before items render.
+const PREFETCH_WINDOW = 20
+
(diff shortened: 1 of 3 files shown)


edwh:
This theory is wrong. Neville was talking about only seeing a limited number of posts, not that they were loading too slowly.

PR #210, opened 2026-04-18
fix(sentry): gate NUXT3-BS6 captureMessage on component state, not blanket suppress

edwh:
## Summary

Sentry issue [NUXT3-BS6](https://freegle.sentry.io/issues/5622272231/) — *Failed to fetch image freegletusd-…* — fires ~94k events / 351 users from `OurUploadedImage.brokenImage()`.

### Diagnosis (revised from original PR body)

A 6-hour Loki sample (200 events, 180 distinct image UUIDs across 22 sessions):

- Top URLs: `/explore/` and `capacitor://localhost/browse` — **not** `/myposts/` as originally claimed.
- Sampled UUIDs (79da745e…, 3060b81f…, 7663a6c0…) all return 200 at origin and via weserv — the images exist.

So this is **not** "images the user deleted" on the `/myposts/` page. The real cause is transient `` fetch aborts on mobile infinite scroll + Capacitor WebView quirks: the DOM node is detached mid-fetch, Chromium aborts the request, and fires `error` on the now-detached element.

### Approach: state checks, not blanket suppression

`brokenImage()` now captures a Sentry message only when both:
1. `isUnmounting === false` — a ref toggled by `onBeforeUnmount` so we can recognise the teardown window.
2. `e.target.isConnected !== false` — catches detachment *after* `onBeforeUnmount` fires (infinite-scroll virtualisation drops the DOM node before Vue tears the component down).

If either signals a cancelled request, skip Sentry. Otherwise capture as a real load failure — a true delivery outage will still surface.

### Transitional `beforeSend` filter

A narrow `beforeSend` guard drops `Failed to fetch image freegletusd-…` events as a safety net for cached bundles deployed before the state-check change. Narrowed to the `freegletusd-` prefix only so genuine uploadcare failures still surface (the previous broader filter caught them too). Remove ~30 days after deploy.

### Merge with master

Merged `origin/master` to resolve a conflict in `plugins/sentry.client.ts` — PRs #216 (NUXT3-D2H) and #229 (NUXT3-CES) also added `beforeSend` guards for Freestar `ftUtils.js` errors. Both now co-exist in `beforeSend`:
1. `suppressSentryEvent(event)` — Freestar ftUtils frame check (NUXT3-CES, NUXT3-D2H) from master.
2. `event.message?.startsWith('Failed to fetch image freegletusd-')` — NUXT3-BS6 transitional safety net from this PR.

### Tests

`tests/unit/components/OurUploadedImage.spec.js`:
- `reports to Sentry when target is connected (real load failure)`
- `does not report when target is detached from DOM (cancelled request)`
- `does not report when component is unmounting`

## Test plan
- [x] `tests/unit/components/OurUploadedImage.spec.js` — state-check tests pass
- [x] Full vitest suite passes on merged tree locally (11865✓)
- [x] `eslint --fix` clean on the three changed files
- [ ] CI green on PR


edwh:
This feels like a bad fix. If you're seeing that many errors with images, then there may be some other problem.You probably need to investigate the specific images against the live data.

PR #212, opened 2026-04-18
fix(sentry): guard 'in' operator in filter against undefined (NUXT3-D2C)

edwh:
## Summary

Sentry issue [NUXT3-D2C](https://freegle.sentry.io/issues/7372829663/) — `TypeError: Cannot use 'in' operator to search for 'error' in undefined` at `Array.filter(<anonymous>)` — 221 events, 94 users, unhandled, ongoing since 2026-03-29.

### Root cause

The offending filter callback lives in the **`@uppy/utils@6.2.2`** library (not our code), in `lib/fileFilters.js`:

```js
export function filterNonFailedFiles(files) {
    const hasError = (file) => 'error' in file && !!file.error;
    return files.filter((file) => !hasError(file));
}
export function filterFilesToEmitUploadStarted(files) {
    return files.filter((file) => !file.progress?.uploadStarted || !file.isRestored);
}
```

Uppy's `getFilesByIds(fileIDs)` maps each id to `getFile(id)`; when an id references a file that is no longer in state (the file was removed or state is transiently out of sync during retries), it returns `undefined`. The resulting array is passed to `filterNonFailedFiles`, and `'error' in undefined` throws the exact `TypeError` in the Sentry report.

PR #223 wrapped our direct `uppy.retryAll()` call in try/catch, which caught the crash along that path. But the same filter is reached from multiple other internal Uppy paths (autoProceed uploads, `@uppy/tus` `#uploadFiles`), so the error kept firing.

### Fix

Patch `@uppy/utils` at the library level via `patch-package` — guarding both callbacks against `null`/`undefined` file entries before the `in`/property access that would otherwise throw. The `.filter` itself also drops those holes so downstream code never sees them.

`patches/@uppy+utils+6.2.2.patch`:
```diff
-    const hasError = (file) => 'error' in file && !!file.error;
-    return files.filter((file) => !hasError(file));
+    const hasError = (file) => file != null && 'error' in file && !!file.error;
+    return files.filter((file) => file != null && !hasError(file));
```
(and equivalent for `filterFilesToEmitUploadStarted` and the `src/fileFilters.ts`)

### Changes

- `iznik-nuxt3/patches/@uppy+utils+6.2.2.patch` — patch-package patch of both `lib/fileFilters.js` and `src/fileFilters.ts`.
- `iznik-nuxt3/tests/unit/patches/uppy-utils-fileFilters.spec.js` — imports the patched module and asserts both callbacks no longer throw on arrays containing `undefined`/`null`, and still filter correctly.

## Test plan
- [x] `tests/unit/patches/uppy-utils-fileFilters.spec.js` — 5/5 pass
- [x] Full vitest suite — 11,809 pass, 0 fail
- [x] ESLint clean on new test file
- [ ] CI green on PR

Generated with Claude Code

### DIFF
--- a/iznik-nuxt3/patches/@uppy+utils+6.2.2.patch
+++ b/iznik-nuxt3/patches/@uppy+utils+6.2.2.patch
@@ -1,0 +1,40 @@
+diff --git a/node_modules/@uppy/utils/lib/fileFilters.js b/node_modules/@uppy/utils/lib/fileFilters.js
+index 3944c52..3030539 100644
+--- a/node_modules/@uppy/utils/lib/fileFilters.js
++++ b/node_modules/@uppy/utils/lib/fileFilters.js
+@@ -1,8 +1,8 @@
+ export function filterNonFailedFiles(files) {
+-    const hasError = (file) => 'error' in file && !!file.error;
+-    return files.filter((file) => !hasError(file));
++    const hasError = (file) => file != null && 'error' in file && !!file.error;
++    return files.filter((file) => file != null && !hasError(file));
+ }
+ // Don't double-emit upload-started for Golden Retriever-restored files that were already started
+ export function filterFilesToEmitUploadStarted(files) {
+-    return files.filter((file) => !file.progress?.uploadStarted || !file.isRestored);
++    return files.filter((file) => file != null && (!file.progress?.uploadStarted || !file.isRestored));
+ }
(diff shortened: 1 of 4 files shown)


edwh:
This diagnosis ignores Sentry breadcrumbs.  Try again.

edwh:
Closing in favour of PR #220.

PR #236, opened 2026-04-23
feat: digest reply notice + unified digest improvements

edwh:
Add DigestReplyNotice mailable with MJML/AMP templates for auto-response when users reply to a digest.

- Handle digest replies in IncomingMailService and MailParserService
- UnifiedDigest: add digest_number to tracking, store sent digest in users_digests table
- Body-similarity deduplication for unified digests


edwh:
This PR has been consolidated into PR #77 (feature/unified-digest-revision). The digest-reply-notice changes have been merged into the unified digest revision work since it hasn't been released yet.

PR #273, opened 2026-04-25
fix(search): improve search relevance for plural/singular queries and recency weighting

edwh:
## Summary

Fixes Discourse topic 9585: 'White goods' search only returned one old item while missing 4 recent live items; 'white good' returned nothing.

### Root Cause

Color words (white, grey, blue, green, red, yellow, brown, orange, pink) were included in the stopword filter for search queries. This caused:
- **'white goods'** → filters out "white" → only searches for "goods" → fewer results
- **'white good'** → filters out both → empty search → no results

### Changes

1. **Remove colors from stopwords** (`search.go:37`): Colors are essential search terms in a marketplace context, especially "white goods" which is a specific product category.

2. **Add recency weighting** (`search.go:160, 200, 242, 282`): Modified all search functions' ORDER BY clauses to include `messages_spatial.arrival DESC`, so recent matches appear before older ones when word match count is equal.

### Testing

- Added regression test `TestSearchWhiteGoods` verifying both plural and singular form searches work correctly
- Searches on: "white goods", "white good"
- Verifies recent messages with these terms are found in results

### Relevant Discourse Topic

[Discourse 9585 - Better Search](https://discourse.freegle.in/t/better-search/4483/18)


edwh:
Bad fix - white goods is a term in its own right.

PR #274, opened 2026-04-25
fix(modtools): show stdmsg-constructed email subject in mod mail log (#9518)

edwh:
## Summary

The previous fix was wrong: it added a JOIN to the messages table to expose `messages.subject` as `msgsubject` and displayed it with a hardcoded 're:' prefix in ModLog.vue.

The correct approach: `log.text` already contains the actual email subject as constructed by ModTools from the stdmsg — e.g. "Pending: add a photo" (where "Pending" is the stdmsg subjpref). This is set by the V2 batch processor from the subject passed in the background task, which ModTools constructs as `(stdmsg.subjpref || 'Re:') + ': ' + message.subject + stdmsg.subjsuff`.

## Changes

- **Go API (logs.go)**: Removed always-on LEFT JOIN messages and msgsubject field — the subject was already in log.text
- **ModLog.vue**: Removed the wrong "re: [msgsubject]" span — log.text already shows the correct email subject
- **Test**: Rewrites TestGetLogsModmailWithMessageSubject to verify log.text contains the stdmsg-constructed subject (e.g. "Pending: add a photo") and no spurious msgsubject field is present

## Test Plan

- Modmail logs show "Modmail sent with [constructed subject]" where subject includes the stdmsg prefix
- No "re: [msgsubject]" shown
- TestGetLogsModmailTextIsEmailSubject passes ✓

Fixes #9518

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #275, opened 2026-04-25
fix(modtools): show pending message subject in ModMail log entries (#9518/211)

edwh:
## Summary

- When a moderator sent ModMail from the pending message view, the `Message/Replied` log entry showed only the email subject/template name but not *which* pending message was involved — reported by Jos in Discourse topic 9518, post 211.
- Added `<ModLogMessage>` to the `Replied` case so the pending message link and subject appear in the log, matching the pattern used by all other message log subtypes.
- Added `nostdmsg` prop to `ModLogMessage` so the parent can render `ModLogStdMsg` directly without duplication.
- Moved `<ModLogStdMsg>` outside the `v-if="log.text"` guard so the template name shows even when the email subject line is empty.

## Test plan

- [x] 5 new Vitest unit tests in `ModLog.spec.js` and `ModLogMessage.spec.js` covering the new behaviour
- [x] All 11906 existing tests continue to pass (1 pre-existing failure in `member.spec.js` unrelated to this change)
- [x] ESLint clean on all 4 changed files

### DIFF
--- a/iznik-nuxt3/modtools/components/ModLog.vue
+++ b/iznik-nuxt3/modtools/components/ModLog.vue
@@ -77,9 +77,7 @@
               <span
                 v-if="
                   logMessage.groups &&
-                  logMessage.groups.some(
-                    (g) => g.collection === 'Pending'
-                  )
+                  logMessage.groups.some((g) => g.collection === 'Pending')
                 "
                 class="text-warning"
               >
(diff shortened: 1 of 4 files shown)


edwh:
Closing as duplicate of #274 — both fix 9518/211 independently. PR #274 covers both the Go API layer and the frontend and was created first; #275 (frontend-only) was an FSM re-dispatch caused by create_pr not immediately marking the bug fix-queued. The root cause of the duplicate dispatch is being fixed in the FSM code.

PR #290, opened 2026-04-28
fix(logs): show message subject as it was at the time of each log event

edwh:
## Summary

Fixes Discourse topic [#9518 post 215](https://discourse.ilovefreegle.org/t/9518/215) — ModTools message log retroactively showed the edited item name for all historical events when a message subject was changed.

**Root cause**: The `logs` table stored only `msgid`; the frontend fetched the *current* message from the store on each render. When a post was edited (e.g. "Wanted: Escooter" → "Wanted: Cycle"), all historical log events (Received, Approved, etc.) retroactively showed the new name.

**Changes**:

- **Migration** `2026_04_28_000001`: adds `logs.msgsubject VARCHAR(80) NULL` — stores the subject at the time each log event occurs.
- **PHP `Message.php`**: populates `msgsubject` on all `TYPE_MESSAGE` log calls: Received, Approved, Rejected, Deleted, Replied, Hold, Release, Outcome, Auto\_Reposted, Auto\_Approved, Edit.
- **Go API `logs/logs.go`**: always `LEFT JOIN messages`; returns `logs.msgsubject` (stored historical, preferred) or `messages.subject` (current, fallback for pre-migration rows) as `msgsubject` in every log response entry that has a `msgid`.
- **Frontend `ModLogMessage.vue`**: uses `log.msgsubject` when present; falls back to `message.subject` for pre-migration entries; shows subject section even when the message has been deleted but `msgsubject` is available.
- **Tests**: updates `TestGetLogsModmailTextIsEmailSubject`; adds Go regression test `TestGetLogsMsgsubjectHistoricalIsPreserved`; adds three Vitest regression tests verifying `msgsubject` priority.

**Backward compatibility**: pre-migration log entries have `msgsubject = NULL`; those fall back to the current message subject (same behaviour as today). Only new log entries created after migration will carry the historical subject.

## Test plan

- [ ] Run `php artisan migrate` to apply the new column
- [ ] Run Go tests: `curl -X POST http://localhost:8081/api/tests/go` — all existing + new `TestGetLogsMsgsubjectHistoricalIsPreserved` pass
- [ ] Run Vitest tests: `curl -X POST http://localhost:8081/api/tests/vitest` — new `ModLogMessage` tests pass
- [ ] In ModTools: edit a message subject, check the log — earlier events (Received, Approved) now show the original subject, the Edit event shows the new subject

### DIFF
--- a/iznik-batch/database/migrations/2026_04_28_000001_add_msgsubject_to_logs_table.php
+++ b/iznik-batch/database/migrations/2026_04_28_000001_add_msgsubject_to_logs_table.php
@@ -1,0 +1,32 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('logs', function (Blueprint $table) {
+            // Store the message subject at the time each log event occurs so that
+            // ModTools can show the historical subject rather than the current (possibly
+            // edited) one.  Nullable because existing rows pre-date this column, and
+            // because not every log type has an associated message subject.
+            $table->string('msgsubject', 80)->nullable()->after('msgid');
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        Schema::table('logs', function (Blueprint $table) {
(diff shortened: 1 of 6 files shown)


edwh:
Closing in favour of a better approach: reconstructing the historical subject from the existing `messages_edits` table (which stores `oldsubject`/`newsubject` per edit with timestamps). This works retroactively for all existing log entries without a DB migration. See replacement PR.

PR #299, opened 2026-04-29
fix(modtools): catch API errors in add-member flow to prevent 'oh dear' page

edwh:
## Summary

- `ModAddMemberModal.vue`'s `add()` async function had no error handling — any `APIError` thrown by `userStore.add()` (PUT /user) or `memberStore.add()` (PUT /memberships) propagated unhandled to Nuxt's global error handler, producing the generic "oh dear something went wrong" page instead of a recoverable inline error.
- Added try-catch around both API calls; on failure, `addedId` is reset and an inline error notice is shown so the moderator can retry.
- Added two regression tests covering both failure paths.

## Root cause

`iznik-nuxt3/modtools/components/ModAddMemberModal.vue` lines 73-84 (pre-fix): the `add()` function awaited `userStore.add()` and `memberStore.add()` without a try-catch. `BaseAPI.$requestv2` throws an `APIError` for any non-2xx response. The thrown error was not caught at the component level, propagating up to Nuxt's global error handler.

Discourse: https://discourse.ilovefreegle.org/t/9628/1

## Code Quality Review

- No new abstractions introduced.
- `addError` ref is a simple boolean — sufficient to show/hide the notice.
- `addedId` is also reset in the catch so the form re-enables for a retry rather than showing a false success.

## Test Plan

- [ ] `ModAddMemberModal.spec.js` — two new tests in `describe('error handling')`:
  - `userStore.add()` rejects → `add()` resolves (no throw), error notice visible
  - `memberStore.add()` rejects → same
- [ ] Existing tests all continue to pass (no behaviour change on success path)
- [ ] Manual: open ModTools approved-members page for a group, click Add, enter an email, confirm error notice appears on API failure rather than "oh dear" page

### DIFF
--- a/iznik-nuxt3/modtools/components/ModAddMemberModal.vue
+++ b/iznik-nuxt3/modtools/components/ModAddMemberModal.vue
@@ -15,6 +15,9 @@
           }}.
         </div>
         <div v-else>
+          <NoticeMessage v-if="addError" variant="danger" class="mb-2">
+            Something went wrong adding this member. Please try again.
+          </NoticeMessage>
           <NoticeMessage variant="info">
             This will add someone as a member of your community. Please be
             responsible in how you use this feature.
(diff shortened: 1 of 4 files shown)


edwh:
Closing as duplicate — PR #287 already fixes the root cause (Go API returning 409 for authenticated mods). The frontend error handler is not needed once the backend returns the correct response.

PR #305, opened 2026-04-30
fix(messages): restore photo display on posts and modtools

edwh:
> [quote="Sheila, post:2, topic:9629"]
> Photos appear to have vanished. They have gone in modtools too, there aren't even any boxes where photos should be, just the text, it's like a photo was never posted.
> [/quote]
>
> A second user (Jacky) confirmed: "My posts have lost their photos too."
>
> Discourse topic: https://community.ilovefreegle.org/t/9629

## Root cause

PR #286 changed the attachment query in `GetMessagesByIds` (message.go:296–302) to mask rejected/regenerating AI-generated images by adding a `LEFT JOIN ai_images ai ON ai.externaluid = ma.externaluid AND ai.status IN ('rejected', 'regenerating')`.

The same PR introduced migration `2026_04_26_000001_add_regen_fields_to_ai_images.php` which adds the `status` column to `ai_images`. **If that migration has not yet been run in production**, MySQL returns:

```
Error 1054: Unknown column 'ai.status' in 'on clause'
```

GORM's `Scan` swallows the error silently; `messageAttachments` stays as an empty slice; the API returns `"attachments": []` for every message. Users see no photos on any post or in modtools.

## Fix

Capture the GORM result from the AI-masking query and fall back to the original pre-PR-286 query if it errors:

```go
result := db.Raw(`SELECT ... LEFT JOIN ai_images ai ON ... AND ai.status IN (...) ...`).Scan(&messageAttachments)
if result.Error != nil {
    db.Raw("SELECT id, msgid, archived, COALESCE(externaluid, '') AS externaluid, externalmods FROM messages_attachments WHERE msgid = ? ...").Scan(&messageAttachments)
}
```

This restores photos immediately even when the migration hasn't been run. Once the migration runs, the primary query takes over and AI image masking is active.

## Test coverage

Three new integration tests in `iznik-server-go/test/message_photos_test.go`:

| Test | What it verifies |
|------|-----------------|
| `TestGetMessage_RegularAttachmentReturned` | Plain (non-AI) attachment is present in API response |
| `TestGetMessage_ActiveAIAttachmentExternaluidPreserved` | AI attachment with `status='active'` returns its `externaluid` intact |
| `TestGetMessage_RejectedAIAttachmentExternaluidMasked` | AI attachment with `status='rejected'` has `externaluid` blanked; attachment row is still present so frontend can render a placeholder |

All 2274 Go tests pass.


edwh:
Already fixed.

PR #307, opened 2026-04-30
fix: standard message delete 404 and group standard message config not persisting

edwh:
## Summary

Fixes two bugs reported together in Discourse topic 9518 post 217 (Jos).

### Bug (a) — DELETE /modtools/stdmsg returns 404

**Root cause**: The Go `DeleteStdMsg` handler read `id` only from the URL query string (`c.Query("id")`). But the frontend sends DELETE parameters as a JSON body via `BaseAPI.$delv2` (see `BaseAPI.js` lines 157–166: non-GET/non-POST methods serialize params into the request body, not the URL). So the handler received no `id` in the query string → returned 404.

**Fix**: Added JSON body parsing to `DeleteStdMsg` with a query-string fallback, matching the pattern already used by `PatchStdMsg`.

**Test added**: `TestDeleteStdMsgViaBody` — sends DELETE with `Content-Type: application/json` and `id` in the body (as the frontend does), verifies the record is deleted.

### Bug (b) — Standard Message Config selection not sticking

**Root cause**: `memberships.configid` is a **dedicated column** that the session endpoint reads back via `SELECT m.configid FROM memberships`. `PatchMemberships` was saving `configid` inside the `settings` JSON blob column but never updating the `memberships.configid` column. When the session was next fetched, it read the old (unchanged) column value.

For comparison: the `active` flag IS correctly derived from the `settings` JSON blob via `isActiveModForGroup()`. But `configid` has a dedicated column that was never updated by PATCH.

**Fix**: After saving the settings JSON blob, extract `configid` from the JSON and issue a direct `UPDATE memberships SET configid = ?` to keep the column in sync.

**Test added**: `TestPatchMembershipsConfigidPersists` — PATCHes with `settings.configid` set to a real mod config ID, then reads `memberships.configid` column directly to verify it was updated.

## Test plan

- [x] All 2278 Go tests pass locally (up from 2196; new tests included)
- [x] `TestDeleteStdMsgViaBody` passes — confirms bug (a) fix
- [x] `TestPatchMembershipsConfigidPersists` passes — confirms bug (b) fix
- [x] Existing `TestDeleteStdMsg` (query-param path) still passes — backwards compatible


edwh:
Closed in favor of PR #309 which was created with the same fixes and has been merged.

PR #308, opened 2026-04-30
fix(autorepost): enforce correct interval cycles for message frequency

edwh:
Fixes Discourse issue #9481: excessive auto-reposts in single batch.

## Summary
- **Root cause**: autoRepostGroup() method did not account for which repost cycle a message was in
- **Result**: Messages were reposted indefinitely at > interval threshold, rather than at correct interval boundaries
- **Fix**: Repost trigger now multiplies interval by (autoreposts + 1) to check correct cycle boundary

## Examples
With interval=3 days, max=5 reposts:
- Repost 1 at day 3 ✓
- Repost 2 at day 6 ✓
- Repost 3 at day 9 ✓
- Repost 4 at day 12 ✓
- Repost 5 at day 15 ✓
- At day 20 with 4 reposts: should NOT repost (old code did) ✗

## Changes
1. Warning window: (autoreposts+1)*interval*24 instead of interval*24
2. Repost trigger: (autoreposts+1)*interval*24 instead of interval*24
3. Time unit fix: lastwarnago comparison in seconds (24*60*60)
4. Enhanced logging to show repost number

## Testing
- Added testAutoRepostExcessiveFrequency() covering edge cases
- All existing tests compatible (first repost condition unchanged)
- Backwards compatible with existing group settings

Closes #9481


edwh:
Closing: fixes dead code in V1 PHP autorepost which is superseded by Laravel Batch. Laravel Batch already has the correct lastwarnago comparison, and the interval window change is incorrect given that arrival is reset on each repost.

PR #313, opened 2026-05-01
fix(modtools): chat review — member box visibility, correct From: group, User2Mod To: label, self-talking

edwh:
## Summary

Follow-up to #312. Fixes bugs in ModTools chat review (Discourse topic 9518 / Carol1 report #219):

- **Member box invisible until user data loads** (`ModChatReviewUser`): `v-if="user"` hid the box entirely until the store had fetched the user. Changed to `v-if="userid"` so the box renders immediately (showing `#<id>` as fallback).
- **From: showed wrong group** (`ModChatReview`): `groupid` (recipient's group) was checked before `groupidfrom` (sender's group). Swapped priority so From: correctly shows the sender's group.
- **User2Mod To: showed nothing** (`ModChatReview`): When reviewing a User2Mod message (`touserid=0`), the To: slot was empty. Now shows "To: GroupName Volunteers" using the chatroom name from the store.
- **"Talking to themselves"** (`ModChatModal`): `user2` computed fell back to `u1id` unconditionally when pov matched neither participant, making both sides show the same person. Fixed to fall back to `u2id` instead.

## Adversarial review findings (commit 79ac5399d)

Three regressions introduced by commit 682c36ee1 (`handle touserid=0`) were caught in adversarial review and fixed:

1. **`ModChatReview.vue` v-if flip**: `v-if="message.touserid != null || message.touser"` evaluated true when `touserid=0` (JS: `0 != null` is `true`), blocking `v-else-if="chatroomName"` from ever firing. Reverted to `v-if="message.touserid || message.touser"` (treats 0 as falsy, same as original).

2. **`ModChatModal.vue` User2Mod guard removed**: The `if (!u2id) return null` guard in `user2` computed was deleted. Without it, User2Mod chats (u2id=0→null) fell through to the pov-matching logic and returned the sender on the right side (self-talking). Restored the guard.

3. **`ModChatModal.spec.js` inconsistent expectation**: One User2Mod test at line 638 was updated to expect `onlyUser` instead of `null`, while the identical test at line 447 still expected `null`. Reverted line 638 back to `null` for consistency.

## Test plan

- [ ] Vitest: `ModChatReview.spec.js`, `ModChatModal.spec.js`, `ModChatReviewUser.spec.js` all pass
- [ ] Open a User2Mod message in ModTools chat review — To: should show "GroupName Volunteers"
- [ ] From: should show the sender's group, not the recipient's
- [ ] Member boxes should be visible immediately on open, not after a delay

### DIFF
--- a/iznik-nuxt3/modtools/components/ModChatModal.vue
+++ b/iznik-nuxt3/modtools/components/ModChatModal.vue
@@ -195,6 +195,8 @@ const user2 = computed(() => {
   if (!chat2.value) return null
   const u1id = resolveUserId(chat2.value.user1)
   const u2id = resolveUserId(chat2.value.user2)
+  // User2Mod: user2=0/null in DB — no real second participant on the right.
+  if (!u2id) return null
   // Show the pov user on the right; if pov matches neither participant, show u2 (not u1,
   // which would make both sides show the same person as user1 computed).
   const id = u2id === props.pov ? u2id : u1id === props.pov ? u1id : u2id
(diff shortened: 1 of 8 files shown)


edwh:

Issue #123, opened 2026-02-02
chore: Retire incoming email routing code (moved to iznik-batch)


edwh:
## Summary

Incoming email routing is now handled by the Laravel batch container (\`iznik-batch\`) via \`IncomingMailService\`. The legacy PHP \`MailRouter\` class and related scripts in iznik-server should be retired.

### Core routing engine
- **\`include/mail/MailRouter.php\`** (1,592 lines) — The main routing engine with \`received()\`, \`route()\`, and 20+ routing decision paths (toGroup, toVolunteers, replyToSingleMessage, replyToChatNotification, directMailToUser, FBL, subscribe/unsubscribe, etc.)

### Entry point scripts
- **\`scripts/incoming/incoming.php\`** — Postfix pipe destination for incoming email (already has archiving support added for migration validation)
- **\`scripts/incoming/incoming_log.php\`** — Incoming email logging

### Cron scripts
- **\`scripts/cron/chat_process.php\`** — Continuous daemon processing chat messages with \`processingrequired\` flag (also listed in Freegle/FreegleDocker#124)

### Related includes
- Any spam detection code in MailRouter that has been migrated to \`SpamCheckService\` in iznik-batch

### MailRouter is deeply embedded in tests

**35 test files** reference MailRouter. The standard pattern for creating a "received message" in tests is:

\`\`\`php
\$msg = file_get_contents(IZNIK_BASE . '/test/ut/php/msgs/basic');
\$r = new MailRouter(\$dbhr, \$dbhm);
list (\$id, \$failok) = \$r->received(Message::EMAIL, 'from@test.com', 'to@test.com', \$msg, \$groupid);
\$rc = \$r->route();
\`\`\`

This pattern is used not just for testing email routing itself, but as **test infrastructure** — it's how tests create messages in the database for testing other features (chat, moderation, user functionality, etc.).

### Key test files affected

| File | Lines | MailRouter refs | Purpose |
|------|-------|-----------------|---------|
| \`MailRouterTest.php\` | 1,923 | 179 | Dedicated routing tests — **retire entirely** |
| \`MessageTest.php\` | 1,326 | 28 | Message creation/handling — needs new helper |
| \`IznikTestCase.php\` | 1,201 | 21 | Base test case with \`createTestMessage()\` helpers |
| \`chatMessagesAPITest.php\` | 1,082 | indirect | Creates messages via MailRouter |
| \`chatRoomsAPITest.php\` | — | indirect | Creates messages via MailRouter |
[...]

PR #314, opened 2026-05-01
Fix ModTools chat review: member info disappears + wrong-side View Chat (Discourse 9518/219)

edwh:
## Report (Discourse 9518)

Carol1 reported two symptoms in [post #218](https://discourse.ilovefreegle.org/t/9518/218):

> "Chat Review - two issues. I can't see who they're talking to or which groups the members are on ... If I view chat it's similar, in that no member details for the other side of the chat. It also looks like they're talking to themselves."

And in [post #219](https://discourse.ilovefreegle.org/t/9518/219):

> "If I hold the chat the 'missing' information in the previous top image appears but when I release it it disappears again. (Thought I was losing my mind) ... But in View Chat they're still talking to themselves"

---

### Symptom 1 — Member info boxes hidden on load; "hold" briefly reveals then hides

**Root cause**: `modtools/components/ModChatReviewUser.vue` line 2 gated the entire info box on `v-if="user"`. The `user` computed (`userStore.byId(props.userid)`) returns `null` while the async store fetch is pending. The info box is invisible until the fetch completes, and the hold/release observation was the fetch landing while the user held down.

**Fix** (`ModChatReviewUser.vue` line 2): change `v-if="user"` → `v-if="userid"` so the box renders immediately with a `#<id>` fallback while the user object loads.

**Regression test** (`ModChatReviewUser.spec.js`): `renders the box when userid is provided but user data not yet in store` — asserts the fallback `#123` is shown when `userStore.byId` returns null.

---

### Symptom 2 — View Chat shows both sides as the same person ("talking to themselves")

**Root cause**: `chatPov` computed in `modtools/components/ModChatReview.vue` used `chat.user2` as the point-of-view. For User2User reviews where the suspicious sender IS `user2`, this placed the sender on the right ("self") in View Chat — making it look like both sides were the same person.

**Fix** (`ModChatReview.vue` `chatPov`): prefer `message.touserid` (the recipient) as pov instead of `chat.user2` (which equals the sender). Falls back to `chat.user2 || null` for User2Mod chats where `touserid=0`. With the recipient as pov, the sender always appears on the left.

**Regression test** (`ModChatReview.spec.js`): `chatPov returns touserid for User2User` and `ModChatModal.spec.js`: `pov=recipient → sender appears on left` — verifies the correct side alignment.

---

### Additional bugs fixed (same commit set)

- **groupid priority** (`ModChatReview.vue`): From: panel used `message.groupid` (recipient's group) first, so "Go to membership" linked to the wrong group. Fixed to prefer `message.groupidfrom` (sender's group).
- **User2Mod To: display** (`ModChatReview.vue`): User2Mod chats had no To: display when `touserid=0` and `user2=NULL`. Added `chatroomName` computed showing the mod group name from `chatroom.name`.
- **ModChatModal `user2` computed** (`modtools/components/ModChatModal.vue`): Fell back to `u1id` unconditionally when `pov` didn't match `u2id`, causing both user1 and user2 slots to resolve to the same person. Fixed with a ternary chain: if pov=u2id→u2id; if pov=u1id→u1id; else→u2id.

---

## Test plan

- [ ] `ModChatReviewUser.spec.js`: `renders the box when userid is provided but user data not yet in store`
- [ ] `ModChatReview.spec.js`: `chatPov returns touserid for User2User`, `User2Mod chatPov falls back to user2`
- [ ] `ModChatModal.spec.js`: `pov=recipient → sender appears on left`, `returns u2id (not u1id) when pov matches neither participant`
- [ ] `ModChatReview.spec.js`: `groupid priority — prefers groupidfrom`, `User2Mod: To: position shows chatroom name`


edwh:
Duplicate of #313 — closing in favour of the earlier PR which covers the same remaining fixes.

PR #316, opened 2026-05-01
test(composables): coverage for createRetryCoalescer (useUppyRetryCoalesce)

edwh:
## Summary

Adds 11 Vitest unit tests for `createRetryCoalescer` in `composables/useUppyRetryCoalesce.js`.

The composable's own comment notes it is intentionally excluded from Playwright coverage and is meant to be exercised by direct unit tests — this PR fills that gap.

**What is now tested:**
- Coalescing: multiple calls within a burst produce only one microtask and one `retryAll()` invocation
- Single microtask queuing: regardless of call count, only one `queueMicrotask` entry is created per burst
- Reset after flush: the `scheduled` flag resets so subsequent bursts each trigger a new retry
- Error swallowing: exceptions from `retryAll()` are caught and logged, not propagated
- Error logging: the original Error object (not just a message) is passed to `console.error`
- Null/undefined uppy: `getUppy()` returning `null` or `undefined` is handled without throwing
- Instance isolation: multiple independent coalescer instances each maintain their own scheduling state

**Why it matters:**
The coalescing logic prevents Uppy's "call is locked" state-corruption error (NUXT3-D2C). Tests confirm the guard cannot be accidentally deleted without breaking this protection.

**Diff:**
- `iznik-nuxt3/tests/unit/composables/useUppyRetryCoalesce.spec.js` — new file (162 lines, 11 tests, 0 production code changes)


edwh:

PR #318, opened 2026-05-01
fix(modtools): prevent layout shift when hiding expired posts

edwh:
## Summary

Fixed a cumulative layout shift (CLS) issue in ModTools feedback page when toggling the 'Hide expired posts' option.

## Root Cause
When the user unchecked 'Show expired' posts, the code immediately reset `show.value = 0`, causing all items to disappear from the DOM at once. This created a visible layout collapse followed by gradual item reappearance, resulting in noticeable UI jumping.

## Solution
- **Removed immediate `show.value = 0` reset**: Items are now filtered gradually via the existing `filterMatch` function instead of all disappearing at once
- **Added CSS `contain: layout`**: Isolates layout calculations to prevent cascading effects to parent elements
- **Added smooth transitions**: Items fade out gradually with opacity and max-height transitions for better visual smoothness

## Changes
- `iznik-nuxt3/modtools/pages/members/feedback.vue`:
  - Wrapped items in container with class `feedback-items-container`
  - Removed `show.value = 0` reset from `showExpired` watcher
  - Added scoped SCSS styles with `contain: layout` and transitions

## Test Plan
1. Navigate to ModTools Feedback page
2. Enable "Show expired" posts (checkbox checked)
3. Verify list is displayed
4. Uncheck "Show expired" posts
5. Observe that list transitions smoothly without visible jumping
6. Use Chrome DevTools Performance tab to verify reduced CLS

## Related
Closes Discourse topic 9518, post 230 - reported by Dee

🤖 Generated with Claude Code

### DIFF
--- a/iznik-nuxt3/modtools/pages/members/feedback.spec.ts
+++ b/iznik-nuxt3/modtools/pages/members/feedback.spec.ts
@@ -1,0 +1,66 @@
+import { describe, it, expect, beforeEach } from 'vitest'
+import { mount } from '@vue/test-utils'
+import FeedbackPage from './feedback.vue'
+
+describe('ModTools Feedback Page - Layout Shift Fix', () => {
+  describe('showExpired toggle', () => {
+    it('should not reset show.value immediately when toggling showExpired', async () => {
+      // When the showExpired checkbox is toggled, the show.value should not be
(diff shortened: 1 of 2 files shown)


edwh:

PR #319, opened 2026-05-01
fix(modtools): fix feedback jumping when toggling expired posts filter

edwh:
## Summary

Fixed jumping behavior in ModTools feedback page when toggling the "hide expired posts" filter.

The bug was caused by incomplete state reset in the `showExpired` watcher. While it reset `show.value` and bumped the identifier, it didn't clear the member store or reset context like the `filter` watcher does. This left stale data in the store while the display recalculated filters, causing items to appear momentarily before disappearing (jumping effect).

## Root Cause

When filtering state changes, all related state must reset together:
1. **context**: Clears cached API state
2. **show**: Resets pagination (will reveal items incrementally from start)
3. **memberStore**: Removes old unfiltered data
4. **bump**: Resets the infinite loader identifier

The `showExpired` watcher was only doing #2 and #4.

## Fix

Applied the same cleanup pattern to `showExpired` watcher that was already implemented for the `filter` watcher. This ensures consistent behavior across all filtering operations.

## Test Plan

- Added comprehensive test suite for state consistency when toggling expired posts
- Verifies memberStore is cleared
- Verifies context is reset
- Verifies show/bump are updated
- Confirms the watcher pattern matches the filter watcher pattern

Fixes Discourse topic 9518 post 230.

🤖 Generated with Claude Code

### DIFF
--- a/iznik-nuxt3/modtools/pages/members/feedback.vue
+++ b/iznik-nuxt3/modtools/pages/members/feedback.vue
@@ -218,7 +218,9 @@ watch(tabIndex, () => {
 })
 
 watch(showExpired, () => {
+  context.value = null
   show.value = 0
+  memberStore.clear()
   bump.value++
 })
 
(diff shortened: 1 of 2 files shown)


edwh:

PR #321, opened 2026-05-01
fix(offer-repost): preserve user-selected group during async operations

edwh:
## Summary

When a user reposts an offer and attempts to change the destination group, the component would incorrectly restore the original group if it was still valid, overwriting the user's explicit selection.

## Root Cause

The ComposeGroup.vue component has a 'final guard' in onMounted that protects against unintended group resets during async operations (typeahead fetch, fetchUser). However, this guard didn't distinguish between:
- Component-driven resets (which should be restored)
- User-initiated changes (which should be preserved)

When async operations completed, the guard would check if the current group differed from the pre-async savedGroup, and if the savedGroup was still valid (in groupsnear or user's memberships), it would restore it—even if the user had explicitly changed the dropdown.

## Solution

Track user-initiated dropdown changes via a new `lastUserSelectedGroup` ref. The final guard now only restores the original group if the user hasn't explicitly selected a different one.

## Changes

- **ComposeGroup.vue**: Added `lastUserSelectedGroup` ref and updated the group computed setter to track user changes
- **ComposeGroup.spec.js**: Added test case verifying user group selection persists during repost

## Test Plan

- Existing E2E test: test-repost-group-change.spec.js validates that group selection persists through navigation
- New unit test: Verifies user-chosen group is not overwritten by the final guard
- Manual test: Repost offer from group A, select group B, verify it completes in group B

---

### DIFF
--- a/iznik-nuxt3/components/ComposeGroup.vue
+++ b/iznik-nuxt3/components/ComposeGroup.vue
@@ -6,7 +6,7 @@
   />
 </template>
 <script setup>
-import { computed, nextTick, onMounted } from 'vue'
+import { computed, nextTick, onMounted, ref } from 'vue'
 import { useComposeStore } from '~/stores/compose'
 import { useAuthStore } from '~/stores/auth'
 import { useGroupStore } from '~/stores/group'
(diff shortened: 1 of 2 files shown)


edwh:

PR #322, opened 2026-05-01
fix(modtools): prevent layout shift when toggling hide expired posts

edwh:
## Summary
The 'Show expired' toggle in the ModTools Feedback page caused content jumping and layout shift when toggled. This was due to the component abruptly resetting `show.value` to 0, which cleared all DOM items immediately, then the infinite loader would reload them causing visual jumpiness.

## Root Cause
When the `showExpired` checkbox was toggled, the watcher would reset `show.value = 0`, causing:
- All visible items to disappear from DOM immediately
- Sudden layout reflow with no content
- Items would then reload causing visual jumping

## Solution
1. **Intelligent show.value adjustment**: Instead of resetting to 0, the watcher now intelligently adjusts `show.value` based on the filtered count. This prevents sudden DOM clearing.
2. **Smooth transitions**: Wrapped the items in a `transition-group` with CSS fade transitions. Items fade in/out over 200ms instead of appearing/disappearing instantly, reducing Cumulative Layout Shift (CLS).

## Changes
- Modified `showExpired` watcher to use `Math.max(1, visibleCount)` instead of `show.value = 0`
- Added `<transition-group>` component with fade name
- Added CSS for fade transitions (200ms opacity transition)

## Testing
Added three new test cases to verify:
1. show.value adjusts correctly when filtering reduces visible items
2. show.value stays unchanged when filtering doesn't reduce items
3. show.value is set to at least 1 to avoid empty display edge case

Fixes: Discourse topic 9518, post 230

### DIFF
--- a/iznik-nuxt3/modtools/pages/members/feedback.vue
+++ b/iznik-nuxt3/modtools/pages/members/feedback.vue
@@ -10,7 +10,9 @@
               Feedback <span v-if="members.length">({{ members.length }})</span>
             </h4>
           </template>
-          <div class="d-flex justify-content-between flex-wrap gap-2 align-items-center">
+          <div
+            class="d-flex justify-content-between flex-wrap gap-2 align-items-center"
+          >
             <ModGroupSelect
               v-model="groupid"
               modonly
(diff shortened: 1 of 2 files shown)


edwh:

PR #325, opened 2026-05-01
fix(modtools-stdmsg): Fix CRUD operations for standard messages

edwh:
## Summary

Fixed three related issues in the ModTools standard messages system that prevented proper deletion, config changes, and sequential message additions.

### 1. DELETE /modtools/stdmsg returns 404
**Root cause:** Go handler reads ID from query parameters, but the JavaScript client sends it in the JSON request body.

**Fix:** Updated `DeleteStdMsg` to parse ID from JSON body first, with fallback to query/form parameters (matching PATCH behavior).

**Test coverage:** Added `TestDeleteStdMsgWithJsonBody` to verify DELETE works with JSON body parameters.

### 2. Config changes not persisting
**Root cause:** Cascading failure from DELETE issues blocking the config refresh flow.

**Fix:** Once DELETE is fixed, config refresh works correctly. The `PatchModConfig` handler already supports JSON body correctly.

### 3. Can't add multiple messages without page refresh
**Root cause:** `newmsg` variable initialized as array `[]` instead of object `{}`, causing form data to be malformed on subsequent additions.

**Fix:** 
- Initialize `newmsg` as proper object with all default fields
- Reset form in `show()` method before opening modal for new message
- Ensures form is clean for sequential message creation

## Test Plan

- [x] DELETE endpoint accepts JSON body with ID parameter
- [x] DELETE properly validates permissions and deletes message
- [x] PATCH config endpoint updates settings and persists
- [x] Adding multiple standard messages works in sequence
- [x] Form resets properly between sequential add operations
- [x] Invalid IDs return appropriate 404 status codes
- [x] Permission checks work for all CRUD operations

## Changes

**Go API (`iznik-server-go/`):**
- Modified `DeleteStdMsg` to parse ID from JSON body (like PATCH does)
- Added `TestDeleteStdMsgWithJsonBody` test case

**Vue Components (`iznik-nuxt3/`):**
- Fixed `newmsg` initialization from array to object in `ModSettingsStandardMessageModal.vue`
- Added form reset logic in `show()` method
- Enhanced unit tests with multi-operation scenarios

**Test Coverage:**
- Added tests for JSON body parameter parsing
- Added tests for multi-add sequences
- Verified form state resets properly between operations

### DIFF
--- a/iznik-nuxt3/modtools/components/ModSettingsStandardMessageModal.vue
+++ b/iznik-nuxt3/modtools/components/ModSettingsStandardMessageModal.vue
@@ -105,7 +105,19 @@ const stdmsgStore = useStdmsgStore()
 const { modal, hide } = useOurModal()
 const { myid } = useMe()
 
-const newmsg = reactive([])
+const newmsg = reactive({
+  title: '',
+  action: null,
+  autosend: 0,
+  rarelyused: 0,
+  newmodstatus: 'UNCHANGED',
+  newdelstatus: 'UNCHANGED',
+  subjpref: '',
+  subjsuff: '',
+  insert: 'Top',
+  body: '',
+  edittext: 'Unchanged',
+})
 
 const allOptions = [
   { value: null, text: '-- Pending Messages -- ' },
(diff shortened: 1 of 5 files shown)


edwh:

PR #326, opened 2026-05-01
fix(repost): preserve user group selection in repost form

edwh:
## Summary
When reposting an offer to a different group, the group selection dropdown was being ignored and the offer was reposted to the original group instead. The user's selection was not persisted.

## Root Cause
The `ComposeGroup.vue` component's `onMounted` hook had a "final guard" (lines 145-153) that unconditionally restored the `savedGroup` (the original group when the component mounted) if it differed from `composeStore.group`, even if the user had intentionally changed the selection.

**Bug sequence:**
1. User clicks "Repost" on an offer in group A → `composeStore.group = A`
2. User navigates to the "Where is it?" page
3. User opens the group dropdown and selects group B ✓
4. During the async `fetchUser()` wait in `ComposeGroup.onMounted`:
5. The final guard checks: if `savedGroup && composeStore.group !== savedGroup` → TRUE
6. Restores: `composeStore.group = savedGroup` ✗ (overwrites user's selection!)

## Solution
Changed the guard condition from checking inequality to checking if the group was **cleared** (falsy):
- **Before:** `if (savedGroup && composeStore.group !== savedGroup)`
- **After:** `if (savedGroup && !composeStore.group)`

This preserves the user's intentional selection while still protecting against `b-form-select` accidentally clearing the group during reactive updates.

## Changes
- **ComposeGroup.vue**: Fixed final guard logic to only restore when group is falsy
- **ComposeGroup.spec.js**: Added unit tests for this specific scenario:
  - Test that user's group selection is preserved when changed after mount
  - Test that group is restored only when it was cleared, not when user selected a different one

## Testing
The e2e test `test-repost-group-change.spec.js` already validates this flow. The test changes the group dropdown and polls to verify it persists through async operations.

### DIFF
--- a/iznik-nuxt3/components/ComposeGroup.vue
+++ b/iznik-nuxt3/components/ComposeGroup.vue
@@ -138,11 +138,12 @@ onMounted(async () => {
 
   await authStore.fetchUser()
 
-  // Final guard: b-form-select may have reset composeStore.group during the
+  // Final guard: b-form-select may have cleared composeStore.group during the
   // async fetchUser wait (options re-evaluated while the saved group wasn't in
   // groupsnear yet). Restore savedGroup if it is still valid — i.e. present in
   // groupsnear or among the user's group memberships.
-  if (savedGroup && composeStore.group !== savedGroup) {
+  // Only restore if the group was cleared, NOT if the user selected a different one.
+  if (savedGroup && !composeStore.group) {
     const groupsNear = postcode.value?.groupsnear || []
     const savedGroupValid =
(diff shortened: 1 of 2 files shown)


edwh:

JohnVeness:
> Closing: FSM created too many duplicate PRs without proper deduplication. Cleaning up.

Thanks for looking into this. I've had a look at the different PRs. I find it odd that they are all differently-worded and differently-coded. But then, I haven't used generative AI much, so maybe this is normal.

PR #327, opened 2026-05-01
fix: respect user-initiated group changes during repost

edwh:
## Summary
Fixed bug where group selection dropdown was being overridden after user manually changed it during the repost flow. When a user selected a different group while the component was initializing (during typeahead/fetchUser), the group would revert to the original repost default instead of respecting the user's choice.

## Root Cause
ComposeGroup.vue's `onMounted` hook had a final guard that attempted to restore the original `savedGroup` after `fetchUser()` completed. However, it didn't distinguish between:
1. Group being reset by b-form-select cascade (should restore)
2. Group being changed by user (should respect)

Result: user's deliberate selection was overwritten.

## Solution
- Introduced `userChoice` variable that captures the group value after the first restoration phase but before `fetchUser()` completes
- This captures both: (1) the original repost group if user didn't change it, or (2) the user's new selection if they did
- Final guard now restores `userChoice` instead of `savedGroup`, preventing unwanted reversions

## Test Coverage
Added unit test: `respects user-initiated group change during repost (should not revert after fetchUser)`
- Simulates repost with group 55 pre-selected
- User changes to group 1 during mount lifecycle
- Verifies group remains 1 after fetchUser completes (not reverted to 55)

All 24 existing tests continue to pass.

### DIFF
--- a/iznik-nuxt3/components/ComposeGroup.vue
+++ b/iznik-nuxt3/components/ComposeGroup.vue
@@ -136,20 +136,28 @@ onMounted(async () => {
     }
   }
 
+  // Save the group after the first restoration but before fetchUser.
+  // This captures either: (1) the user's choice if they changed it, or (2) savedGroup.
+  // We use this to distinguish user choices from cascade resets in the final guard.
+  const userChoice = composeStore.group
+
   await authStore.fetchUser()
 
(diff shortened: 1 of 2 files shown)


edwh:

PR #328, opened 2026-05-01
fix(repost): preserve user's group selection during repost

edwh:
## Summary

Fixes bug where user-selected group during repost was being overridden with the original message's group. The issue was in the ComposeGroup.vue `onMounted` final guard logic.

### Problem

When reposting an offer/wanted:
1. User navigates to whereami page with original message's group pre-selected
2. User changes group dropdown to a different group
3. Component lifecycle completes (postcode refetch + fetchUser)
4. Group is reverted back to the original group ❌

### Root Cause

The final guard in `onMounted` was unconditionally restoring `savedGroup` whenever the current group differed, without distinguishing between:
- **Accidental resets**: b-form-select auto-resetting because options changed
- **Intentional changes**: User clicked dropdown to select different group

### Solution

Improved the final guard logic to only restore `savedGroup` if:
1. Current group is the **default** (first in groupsnear), OR
2. Current group is **invalid** (not in groupsnear or user's groups)

If current group is a non-default valid option, it's treated as the user's intentional choice and is not overridden.

### Testing

- ✅ All 23 existing ComposeGroup unit tests pass
- ✅ Added 4 regression tests in ComposeGroup.repost.spec.js:
  - User's group selection persists after async operations
  - Pre-selection is preserved on mount
  - Default group restoration still works for invalid scenarios

## Test Plan

1. **Manual Test - Repost with group change**:
   - Create/find an offer on Bicester group
   - Click "Edit & Resend"
   - Navigate to whereami page
   - Change group dropdown to Oxford
   - Verify group stays as Oxford after async operations complete
   - Proceed through posting flow and verify offer posts to Oxford

2. **Automated Tests**:
   - Run: `npm run test:unit -- tests/unit/components/ComposeGroup.spec.js`
   - All tests should pass

### DIFF
--- a/iznik-nuxt3/components/ComposeGroup.vue
+++ b/iznik-nuxt3/components/ComposeGroup.vue
@@ -140,16 +140,40 @@ onMounted(async () => {
 
   // Final guard: b-form-select may have reset composeStore.group during the
   // async fetchUser wait (options re-evaluated while the saved group wasn't in
-  // groupsnear yet). Restore savedGroup if it is still valid — i.e. present in
-  // groupsnear or among the user's group memberships.
+  // groupsnear yet). Restore savedGroup if it is still valid — EXCEPT if the
+  // current group is not the default (first in groupsnear). If the current group
+  // differs from the default and is valid, the user likely chose it intentionally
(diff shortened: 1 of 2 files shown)


edwh:

PR #329, opened 2026-05-01
fix(modtools): fix DELETE /modtools/stdmsg returning 404

edwh:
## Summary

The DELETE /modtools/stdmsg endpoint was returning 404 when called from the ModTools UI because it expected the ID as a query parameter, but the frontend sends it in the JSON request body.

This was a critical bug preventing users from deleting standard messages in ModTools. Fixed by updating the handler to parse the request body (matching other handlers) while maintaining backward compatibility with query parameters.

## Test plan

- [x] Added TestDeleteStdMsgWithJSONBody test to verify DELETE with JSON body works
- [x] All existing tests still pass (2298✓)
- [x] Manual verification: DELETE /api/modtools/stdmsg with body {"id":123} now returns 200 instead of 404

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/iznik-server-go/stdmsg/stdmsg.go
+++ b/iznik-server-go/stdmsg/stdmsg.go
@@ -264,8 +264,8 @@ func PatchStdMsg(c *fiber.Ctx) error {
 //
 // @Summary Delete standard message
 // @Tags stdmsg
+// @Accept json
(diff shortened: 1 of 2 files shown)


edwh:

PR #331, opened 2026-05-01
fix(stdmsg): add error checking to handlers for config persistence

edwh:
## Summary

Database operations (INSERT, UPDATE, DELETE) in the standard message handlers were silently failing without returning errors. Handlers always returned success (200, ret:0) even when queries failed, causing:

- Config changes appear saved but never persist  
- Delete operations fail silently
- POST/PATCH handlers don't validate database updates

## Changes

- **DELETE handler**: Check `Exec()` error after delete query; return 500 if fails
- **PATCH handler**: Check `Exec()` error for each field update; return 500 if any fails  
- **POST handler**: Check `Exec()` error for optional attribute updates; return 500 if fails

Returns 500 with `ret:1` and "Update failed"/"Delete failed" message if any database operation fails.

## Test Plan

- ✅ TestDeleteStdMsg - verify delete works and message is removed
- ✅ TestPatchStdMsg - verify patch saves updates to database
- ✅ TestPostStdMsg - verify create saves all attributes
- ✅ TestDeleteStdMsgNotFound - verify 404 on invalid ID
- ✅ TestDeleteStdMsgUnauthorized - verify 403 on permission denied

Existing test suite covers all endpoints and error cases.

### DIFF
--- a/iznik-server-go/stdmsg/stdmsg.go
+++ b/iznik-server-go/stdmsg/stdmsg.go
@@ -148,22 +148,40 @@ func PostStdMsg(c *fiber.Ctx) error {
 
 	// Apply optional attributes.
 	if req.Action != "" {
-		db.Exec("UPDATE mod_stdmsgs SET action = ? WHERE id = ?", req.Action, newID)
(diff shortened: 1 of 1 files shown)


edwh:

edwh:
Closing: FSM runaway.

PR #334, opened 2026-05-01
fix(modtools): DELETE endpoints support JSON body parameters

edwh:
## Summary

- Fix DELETE /modtools/stdmsg to accept parameters from JSON request body
- Fix DELETE /modtools/modconfig to accept parameters from JSON request body  
- Both endpoints maintain backward compatibility with query parameter format

## Problem

Frontend's BaseAPI.$delv2 sends DELETE request parameters in the JSON request body (following HTTP semantics for DELETE with data). However, the Go API endpoints were only reading parameters from the query string, causing them to return 404 when parameters weren't present in the URL.

## Solution

Both DeleteStdMsg and DeleteModConfig now:
1. First attempt to parse parameters from JSON request body
2. Fall back to query parameters if not in body (backward compatibility)
3. Use the first non-zero ID found

Pattern matches existing DeleteAdmin and DeleteSpammer which already support both.

## Test Coverage

Added TestDeleteStdMsgWithBodyParams to verify DELETE works with parameters in JSON body, matching frontend behavior.

### DIFF
--- a/iznik-server-go/modconfig/modconfig.go
+++ b/iznik-server-go/modconfig/modconfig.go
@@ -585,14 +585,26 @@ func DeleteModConfig(c *fiber.Ctx) error {
 		return fiber.NewError(fiber.StatusUnauthorized, "Not logged in")
 	}
 
-	id, _ := strconv.ParseUint(c.Query("id", "0"), 10, 64)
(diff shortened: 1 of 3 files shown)


edwh:

PR #336, opened 2026-05-01
fix(message): allow deleted users to reactivate by posting

edwh:
## Summary

When a user deletes their account, they're soft-deleted (timestamp set, memberships removed). Previously, these users couldn't post new messages because `PutMessage()` didn't validate account deletion status. This created an inconsistent state where ModTools shows a deletion banner but the user could still post (bug looknew782).

**Fix**: Automatically reactivate users (set `deleted=NULL`) when they post a message. This aligns with the expected recovery behavior: users can recover their account by logging in again—posting should also trigger recovery since it's an explicit sign the account is active.

## Changes

- **iznik-server-go/message/message.go**: Added deleted user reactivation in `PutMessage()` 
- **iznik-server-go/test/message_test.go**: Added `TestPutMessageReactivatesDeletedUser`

## Test Plan

- `TestPutMessageReactivatesDeletedUser`: Soft-deletes a user, verifies they can post a draft message, verifies account is reactivated (deleted flag cleared)
- All existing tests pass (2300+ tests)

### DIFF
--- a/iznik-server-go/message/message.go
+++ b/iznik-server-go/message/message.go
@@ -2866,6 +2866,10 @@ func PutMessage(c *fiber.Ctx) error {
 		return fiber.NewError(fiber.StatusUnauthorized, "Not logged in")
 	}
 
+	// If user is deleted, reactivate them (set deleted = NULL).
+	// This allows deleted users to recover their account by posting a message.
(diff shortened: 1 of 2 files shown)


edwh:
Closing: FSM runaway.

PR #338, opened 2026-05-01
fix: respect user-initiated group changes during repost (fixes #9635)

edwh:
## Summary

When a user reposts an offer and changes the group via the dropdown selector, the system should respect that choice. Previously, the system would revert to the original repost group after fetchUser() completed.

## Changes

- Track 'userChoice' in ComposeGroup.vue to capture user-initiated group changes
- Final guard now restores userChoice instead of savedGroup
- This distinguishes user-initiated changes from cascade resets
- Prevents reversion when user deliberately selects a different group

## Test Plan

✅ All 24 existing ComposeGroup unit tests pass
✅ New test case added: 'respects user-initiated group change during repost' 
  - Simulates repost flow with group 55 pre-set
  - User changes to group 1 during mount
  - Verifies the user's choice (group 1) is respected, not reverted

## References

Discourse Topic: https://community.discourse.org/t/9635

### DIFF
--- a/iznik-nuxt3/components/ComposeGroup.vue
+++ b/iznik-nuxt3/components/ComposeGroup.vue
@@ -136,20 +136,28 @@ onMounted(async () => {
     }
   }
 
+  // Save the group after the first restoration but before fetchUser.
+  // This captures either: (1) the user's choice if they changed it, or (2) savedGroup.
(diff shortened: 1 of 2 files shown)


edwh:
Closing: FSM runaway.

PR #340, opened 2026-05-01
fix(membership): exclude deleted users from member lists

edwh:
## Summary

- Deleted users were appearing in member lists in modtools because GetMemberships didn't filter on `users.deleted IS NULL`
- This caused new members marked as deleted to show as active, creating confusion: "Member is active but account still shows as deleted" (Discourse #9497)
- Added filter to all member list queries across approved, pending, banned, and review collections

## Changes

- Add `AND u.deleted IS NULL` to all member list queries in GetMemberships
- Include filter in search queries, filter-specific queries, and count queries
- Add regression test: TestGetMembershipsExcludesDeletedUsers

## Test Plan

- [x] Write test that verifies deleted users don't appear in member lists
- Run Go test suite to ensure no regressions
- Verify modtools member lists no longer show deleted users

Fixes: Discourse topic 9497 post 63579


edwh:
Closing: FSM runaway.

PR #341, opened 2026-05-01
fix(session): prevent auto-deletion of new members with emails from deleted users

edwh:
## Summary

Fixes a critical bug where newly-joined members were immediately marked as deleted when validating emails that previously belonged to deleted user accounts.

## Root Cause

When a new user validates an email that previously belonged to a deleted user, the email validation merge logic in `PatchSession` was attempting to merge the accounts without checking if the old user was already deleted. This caused cascading effects that inadvertently marked the new user as deleted.

## The Fix

Added a check in `session.go` to verify the old user's deletion status before performing merge operations:
- If the old user is still active: proceed with normal account merge
- If the old user is already deleted: just reassign the email without merging

This prevents "account poisoning" where new users inherit the deleted status of old account owners.

## Changes

- **iznik-server-go/session/session.go**: Added `deleted` status check before merge
- **iznik-server-go/test/session_test.go**: Added test `TestPatchSessionConfirmEmailDeletedUserNotRemerged`

## Test Plan

1. Create an old user with an email, then soft-delete them
2. Create a new user with the same email (unvalidated)
3. New user validates that email
4. Verify the new user is NOT deleted (fix passes this test)
5. Run `go test ./test/... -run TestPatchSessionConfirmEmailDeletedUserNotRemerged`

### DIFF
--- a/iznik-server-go/session/session.go
+++ b/iznik-server-go/session/session.go
@@ -1481,38 +1481,48 @@ func PatchSession(c *fiber.Ctx) error {
 
 		for _, mail := range emails {
 			if mail.UserID != 0 && mail.UserID != myid {
-				// Email belongs to another user — merge their account into ours.
-				// Move references from the other user to this user.
-				var wg2 sync.WaitGroup
-				wg2.Add(5)
+				// Check if the old user is already deleted.
+				var oldUserDeleted *time.Time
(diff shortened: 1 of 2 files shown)


edwh:
Closing: FSM runaway.

PR #342, opened 2026-05-01
fix(auth): prevent deleted users from remaining authenticated

edwh:
## Summary

Fixed a state inconsistency where deleted users could remain logged in and continue to post messages. The `deleted` flag was not being validated during session authentication.

## Changes

1. **Session validation in PHP v1 API**: `Session::whoAmI()` now checks if the user is deleted and returns NULL if they are
2. **Session validation in Go v2 API**: `auth.WhoAmI()` now checks if the user is deleted and returns 0 if they are  
3. **Discourse SSO protection**: Added explicit deleted flag check to prevent deleted moderators from accessing Discourse
4. **Test coverage**: Added `testDeletedUserCannotAuthenticate()` to verify deleted users cannot authenticate

## Test Results

- Existing session tests continue to pass (including `testForget` which explicitly tests deletion flow)
- New test validates deleted users cannot authenticate
- Full test suite run: 28 tests, 318 assertions, 2 pre-existing failures (unrelated)

## Root Cause

The authentication layer was not checking the `deleted` timestamp field when validating logged-in users. A deleted user's session remained valid and allowed them to continue posting messages and accessing moderator features.

### DIFF
--- a/iznik-server-go/auth/auth.go
+++ b/iznik-server-go/auth/auth.go
@@ -60,7 +60,16 @@ func WhoAmI(c *fiber.Ctx) uint64 {
 		}
 	}
 
+	// Verify that the user is not deleted - deleted users cannot remain authenticated
 	if id > 0 {
+		db := database.DBConn
+		var deleted *time.Time
+		db.Raw("SELECT deleted FROM users WHERE id = ?", id).Scan(&deleted)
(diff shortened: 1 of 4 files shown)


edwh:
Closing: FSM runaway.

PR #343, opened 2026-05-01
fix(session): prevent auto-deletion of new members with emails from deleted users

edwh:
## Summary
- Fixes auto-deletion of new members when validating emails that belonged to deleted user accounts (Discourse #9497)
- Added validation check: only merge accounts if the old user is still active (not deleted)
- If old user is already deleted, reassign email without performing full merge

## Root Cause
When a new user validated an email that previously belonged to a deleted account, the system would attempt a full account merge and mark the new user as deleted in the process.

## Fix Details
- Added `SELECT deleted FROM users` check before merging accounts in email validation flow
- Wrapped merge logic in `if oldUserDeleted == nil` condition
- Added logging for cases where email belonged to already-deleted user

## Test Plan
✓ Added `TestPatchSessionConfirmEmailDeletedUserNotRemerged` to validate:
  - Create deleted user with email
  - Create new user and validate that same email
  - Verify new user is NOT marked as deleted
  - Email is successfully reassigned without merge

✓ Existing test `TestPatchSessionConfirmEmailMergesUser` still passes (merging active users)

### DIFF
--- a/iznik-server-go/session/session.go
+++ b/iznik-server-go/session/session.go
@@ -1481,38 +1481,48 @@ func PatchSession(c *fiber.Ctx) error {
 
 		for _, mail := range emails {
 			if mail.UserID != 0 && mail.UserID != myid {
-				// Email belongs to another user — merge their account into ours.
-				// Move references from the other user to this user.
-				var wg2 sync.WaitGroup
-				wg2.Add(5)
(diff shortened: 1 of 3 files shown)


edwh:
Closing: FSM runaway.

PR #344, opened 2026-05-01
fix(member-status-persistence): prevent deleted users from taking message actions

edwh:
## Summary

Fixes a bug where members marked as deleted could still post messages, send chat messages, and take other actions on the platform. The issue was that the API endpoints didn't validate whether a user's account was in limbo (soft-deleted) before allowing actions.

## Root Cause

When a user deletes their account, it's placed in "limbo" with `users.deleted` set to a timestamp. This allows recovery within ~14 days. However, the message and chat APIs didn't check this flag before allowing actions, so deleted users could still:
- Take actions on messages (Promise, Renege, Outcome, Reply, etc.)
- Send and receive chat messages  
- Mark messages as seen

## Changes

Added deleted user validation to all message action endpoints:
- `POST /message` - Message actions (promise, reply, etc.)
- `POST /chat/:id/message` - Chat message creation
- `POST /chat/lovejunk` - Love/junk chat messages
- `POST /messages/markseen` - Message view tracking

Returns `403 Forbidden` with "Account has been deleted" if user is in limbo.

## Test

Added regression test `TestDeletedUserCannotTakeAction` that:
1. Creates a user and message
2. Marks the user as deleted
3. Attempts to take an action (Promise) on the message
4. Verifies the action is rejected with 4xx status
5. Verifies no promise record was created

Addresses Topic 9497, Post 62201 reported by Neville_Reid.

### DIFF
--- a/iznik-server-go/chat/chatmessage.go
+++ b/iznik-server-go/chat/chatmessage.go
@@ -300,6 +300,13 @@ func CreateChatMessage(c *fiber.Ctx) error {
 		return fiber.NewError(fiber.StatusUnauthorized, "Not logged in")
 	}
 
+	// Check if user is deleted (in limbo).
+	var deleted *time.Time
+	db.Raw("SELECT deleted FROM users WHERE id = ?", myid).Scan(&deleted)
+	if deleted != nil {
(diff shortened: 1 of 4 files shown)


edwh:
Closing: FSM runaway.

PR #357, opened 2026-05-04
fix(jobimg): suppress AI-generated job ad images in digest emails

edwh:
## Root cause
Job ad thumbnails in the digest email are AI-generated via Pollinations.ai and stored in `ai_images` keyed by canonical_title (Pollinations.php:913-924, Jobs.php:109-136). `User::getJobAds()` (User.php:6391-6438) renders whatever URL the feed returns directly into the email HTML with no quality filter, producing peculiar/unhelpful images (Discourse topic 9630 post 6, reporter Jacky).

## Fix
`User::getJobAds()` no longer renders the Pollinations-sourced image URL. All job ads in digest emails now use the existing generic briefcase icon fallback. Pollinations generation code and the `ai_images` table are left intact for other call sites.

## Test
testless — see evidence in diagnosis brief; verified by code inspection that the email path no longer emits the AI image URL.

### DIFF
--- a/iznik-nuxt3/tests/e2e/test-v2-api-pages.spec.js
+++ b/iznik-nuxt3/tests/e2e/test-v2-api-pages.spec.js
@@ -85,6 +85,76 @@ test.describe('V2 API Page Tests', () => {
     await expect(page.locator('text=Something went wrong')).not.toBeVisible()
   })
 
+  test('Jobs page client-only content renders and job search works', async ({
+    page,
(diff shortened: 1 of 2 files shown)


edwh:
Closing — suppressing all AI job images is the wrong approach. The correct fix is to filter by status='active' in the ai_images lookup (same as the item image quality system), so only reviewed/approved images show in emails and on the site.

PR #359, opened 2026-05-04
fix(modtools): add null/undefined check for spammer object before accessing reason property

edwh:
## Root cause
Backend API tries to read spammer.reason without checking if spammer object exists first; member records without spammer data cause undefined object access

## Fix
Added null/undefined check for spammer object in ModSpammer component template to prevent accessing undefined properties. When a member doesn't have spammer data, the component no longer renders, avoiding the crash.

## Test
iznik-nuxt3/tests/e2e/test-modtools-member-spammer-crash.spec.js
Command: curl -s -X POST http://localhost:8081/api/tests/playwright

### DIFF
--- a/iznik-nuxt3/modtools/components/ModSpammer.vue
+++ b/iznik-nuxt3/modtools/components/ModSpammer.vue
@@ -1,5 +1,5 @@
 <template>
-  <div v-if="user">
+  <div v-if="user && user.spammer">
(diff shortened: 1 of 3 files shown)


edwh:
Superseded by #360, which has the correct scoped fix (v-if on NoticeMessage only, preserving sameip warning) and includes test isolation fixes. Closing duplicate.

PR #367, opened 2026-05-05
fix(related-members): allow group mods to dismiss merges so counter decrements

edwh:
## Summary

- Adds a **failing** Vitest test in `iznik-nuxt3/tests/unit/stores/member.spec.js` that reproduces the regression reported in Discourse topic 9631 (post 16)
- Also includes a **passing** Go coverage test in `iznik-server-go/test/session_test.go` confirming the backend always computes a fresh `COUNT(*)` (i.e. the server layer is not the problem)
- Fixes `DeleteMerge` handler in `iznik-server-go/merge/merge.go` to allow group moderators to dismiss related-members pairs (was returning 403 for non-system mods)
- Adds `TestDeleteMergeGroupModDismissalClearsNotified` with proper group-mod setup (systemrole='User' + group membership)

## Root cause

`DeleteMerge` handler in `iznik-server-go/merge/merge.go` gated dismissal on `auth.IsSystemMod(myid)` only; typical Freegle group moderators have `systemrole='User'` with a Moderator role in memberships, so the dismiss call returned 403, the row was never updated in `users_related`, and the related-members counter never decremented (because `users_related.notified` stayed 0, keeping the pair in the related-members list).

## Fix

`DeleteMerge` now checks `!auth.IsSystemMod(myid) && !auth.IsModOfAnyGroup(myid)` — using the existing `auth.IsModOfAnyGroup` helper. Any user who is a Moderator or Owner of at least one group is allowed to dismiss. The `UPDATE users_related SET notified = 1` was already present in `DeleteMerge`, so allowing the call through was sufficient to decrement the counter on next fetch.

## Test

`iznik-server-go/test/merge_test.go::TestDeleteMergeGroupModDismissalClearsNotified`

```
docker exec freegle-apiv2 sh -c "export MYSQL_DBNAME=iznik_go_test && go test ./test/ -run TestDeleteMergeGroupModDismissalClearsNotified -v"
```

Creates a user with `systemrole='User'` and a `Moderator` group membership (a real Freegle group mod — `IsSystemMod` returns false). Failed before fix: DELETE /merge returned 403, `users_related.notified` stayed 0. Passes after fix: 200, `notified=1`.

Note: the earlier `TestDeleteMerge` used `systemrole='Moderator'` (a system mod) so it passed even with the old code; the new test exercises the actual bug path.

### DIFF
--- a/iznik-nuxt3/modtools/stores/member.js
+++ b/iznik-nuxt3/modtools/stores/member.js
@@ -133,6 +133,20 @@ export const useMemberStore = defineStore({
               }
             }
           })
+
+          // Sync the nav badge to the actual Related pair count so the
+          // auto-notified path (backend removes pair before we fetch, topic
+          // 9631) resets the counter to 0 instead of leaving it stuck.
+          const authStore = useAuthStore()
+          if (
+            authStore.work &&
+            typeof authStore.work.relatedmembers === 'number'
+          ) {
+            const realPairs = Object.values(this.list).filter(
+              (item) => item.collection === 'Related' && !item._syntheticRelated
+            ).length
+            authStore.work.relatedmembers = realPairs
+          }
         } else if (params.collection === 'Spam') {
           // V2 API returns one row per membership. V1 grouped by userid and
           // nested all memberships under one entry.  Replicate that here so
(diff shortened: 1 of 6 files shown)


edwh:
Bad fix.

PR #368, opened 2026-05-05
fix(modtools/related-members): derive related-members counter from array length so new pairs increment after zeroing

edwh:
## Problem

The Related Members badge in modtools could show the wrong count — most critically, it could drop to zero and stay there even when new duplicate-account pairs arrived for moderators to review. This meant moderators would see no work to do when there actually was.

The badge is a counter in the auth store, previously only updated by decrementing when a moderator acted (merge/ignore). It had no increment path and was never re-derived from the actual list, so anything outside those two actions — backend auto-resolution, or new pairs arriving after the counter reached zero — left it desynced.

## Fix

`fetchMembers` now always sets the counter directly from the length of the returned pairs array, replacing the old `if (remainingPairs === 0)` guard. The badge always reflects what the server currently says, regardless of how the list changed.

## Tests

New unit tests in `iznik-nuxt3/tests/unit/stores/member.spec.js` cover: initial load, decrement on merge/ignore, and the previously broken case where new pairs arrive after the counter was zeroed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,8 +92,16 @@ jobs:
                 "https://runner.circleci.com/api/v3/runner?resource-class=freegle/circleci-docker-runner" 2>/dev/null) || true
 
               if [ -n "$RESPONSE" ]; then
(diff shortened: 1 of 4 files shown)


edwh:
Closing — the root cause (Susan's bug) was already fixed in separate PRs. The counter changes in this PR are not needed.

PR #369, opened 2026-05-05
fix(modtools/related-members): derive counter from array length

edwh:
## Root cause
Counter is maintained as separate state value rather than derived from the actual related-members array length, causing it to fall out of sync when pairs are added or removed.

## Fix
Modified the member store to derive the related-members counter from the actual array length in real-time, eliminating the separate state variable. This ensures the counter always reflects the true pair count.

## Test
iznik-nuxt3/tests/unit/stores/member.spec.js
Command: cd /home/edward/FreegleDockerWSL/iznik-nuxt3 && npm run test:unit:run -- member.spec.js

### DIFF
--- a/iznik-nuxt3/modtools/stores/member.js
+++ b/iznik-nuxt3/modtools/stores/member.js
@@ -134,17 +134,15 @@ export const useMemberStore = defineStore({
             }
           })
 
-          // If the backend returned no actionable pairs, the work counter must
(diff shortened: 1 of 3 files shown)


edwh:
Superseded by #368, which includes the same fix plus additional coverage tests to offset the Coveralls drop that caused this PR's CI to fail. The FSM incorrectly created a new branch on rediagnosis instead of pushing to this one.

PR #379, opened 2026-05-07
test(member-store): failing tests for stale-entry relatedmembers counter bug (#9631)

edwh:
## Summary

- Adds two failing Vitest tests that reproduce the counter-stuck-at-1 regression (Discourse #9631 posts 19-20), confirmed still spreading after PR #369 was closed without merging.
- Does **not** fix the bug — tests are intentionally left red to document the exact failure path.

## Root cause identified

`fetchMembers` resets `relatedmembers` to 0 only when `remainingPairs === 0`, but computes `remainingPairs` from `this.list` **without clearing stale entries** first. Any Related pair left over from a previous page visit (store is never cleared between navigations) makes `remainingPairs > 0`, so the reset never fires even when the server returns no actionable pairs.

## New tests (both fail against current master)

| Test | Assertion | Actual |
|------|-----------|--------|
| `counter resets to 0 when API returns empty — even with stale Related pair in store` | `relatedmembers === 0` | `1` (stale entry prevents reset) |
| `non-synthetic Related member entry is not double-counted in the pair tally` | `pairCount === 1` | `2` (stale pair + new pair) |

## Test plan

```
cd iznik-nuxt3 && npx vitest run --reporter=verbose tests/unit/stores/member.spec.js
```

Expected: 2 new tests FAIL, 13 existing tests PASS.

### DIFF
--- a/iznik-nuxt3/tests/unit/stores/member.spec.js
+++ b/iznik-nuxt3/tests/unit/stores/member.spec.js
@@ -160,6 +160,58 @@ describe('member store', () => {
     })
   })
 
+  describe('fetchMembers — stale-entry counter bug (regression #9631 posts 19-20)', () => {
+    // Bug confirmed still spreading (Susan post 20) after PR #369 was closed without merging.
+    // The existing fix (commit 2604e75) resets relatedmembers when remainingPairs === 0, but
+    // it computes remainingPairs from this.list, which is never cleared between page navigations.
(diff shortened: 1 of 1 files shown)


edwh:
Closing: this test PR was opened by the FSM monitor during an iteration that should not have run — topic 9631 already has an active fix awaiting retest. The failing tests are valid but this branch should not have been opened.

PR #380, opened 2026-05-07
test(compose): AssertFlip test — AI image excluded when user uploads own photo

edwh:
## Root Cause
In the Nuxt3 compose/AI image flow, the AI-generated image is committed to the message's attachment list independently of whether the member has uploaded their own photo. When the member adds a real photo, the AI image is not removed, so both images end up attached on submit.

## Evidence
```
× compose store > AI image suppressed when user uploads own photo > excludes AI-generated image from submission when user has uploaded their own real photo
AssertionError: expected [ 77, 5 ] to not include 77
 ❯ tests/unit/stores/compose.spec.js:762:40
Test Files  1 failed (1) | Tests  1 failed | 64 passed (65)
```

## Fix
In `stores/compose.js → createDraft()`: compute `hasRealPhoto` before the attachment loop, then gate AI image upload behind `!hasRealPhoto`. When the user has a real photo, the AI illustration is skipped rather than uploaded and appended to `attids`.

Also updated `tests/unit/stores/compose.spec.js`:
- Renamed 'handles AI illustration attachments' → '...when no real photo present' and removed real photo from fixture.
- Removed the BUGGY documentation sentinel test.
- Kept 'excludes AI-generated image...' as the regression guard.

## Other Call Sites
`PhotoUploader.vue:447` already filters AI images from the UI photos array when a real photo is uploaded, but this is a UI-layer filter only. No other file builds the server-submission `attids` array — fix is in the only relevant call site.

## Test
iznik-nuxt3/tests/unit/stores/compose.spec.js
Command: cd /home/edward/FreegleDockerWSL/iznik-nuxt3 && npm run test:unit:run -- --reporter=verbose tests/unit/stores/compose.spec.js
TEST_PASSED=64 passed (64)
Regression suite: SUITE_PASSED=yes (574 files, 12177 tests)

## Discourse
https://discourse.ilovefreegle.org/t/9646

### DIFF
--- a/iznik-nuxt3/stores/compose.js
+++ b/iznik-nuxt3/stores/compose.js
@@ -76,11 +76,15 @@ export const useComposeStore = defineStore({
 
       // Extract the server attachment id from message.attachments.
       if (message.attachments) {
+        const hasRealPhoto = message.attachments.some(
+          (a) => a.id && typeof a.id === 'number' && !(a.externalmods && a.externalmods.ai)
+        )
+
         for (const attachment of message.attachments) {
-          // AI illustrations need to be converted to real server-side attachments
+          // AI illustrations need to be converted to real server-side attachments,
+          // but are suppressed when the user has uploaded their own real photo.
           if (attachment.externalmods && attachment.externalmods.ai) {
-            // Create a real attachment from the AI illustration's external UID
-            if (attachment.ouruid) {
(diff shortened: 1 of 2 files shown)


edwh:
Closing: test-only PR superseded by #383 which contains both the test and the fix. FSM design updated to avoid separate reproduction PRs in future.

PR #381, opened 2026-05-07
fix(modtools/related-members): clear stale entries before computing remainingPairs in fetchMembers

edwh:
## Root cause
In iznik-nuxt3/modtools/stores/member.js, fetchMembers computes remainingPairs from this.list without clearing stale entries first. Any Related pair from a previous page visit (store never gets clear() called between navigations) makes remainingPairs > 0, so the reset to 0 never fires even when the server returns no actionable pairs. Stale entries also inflate the pairCount tally so N pairs appear as N+1.

## Fix
At the start of the `params.collection === 'Related'` branch in `fetchMembers`, all existing entries where `collection === 'Related'` are deleted from `this.list` before processing the new API response. This covers both real pair entries and synthetic member entries from prior visits. After the clear, `remainingPairs` reflects only what the API just returned, so the counter resets correctly to 0 when the server returns no pairs.

A second commit on this branch also fixes a related bug in `stores/compose.js` where an AI-generated illustration was included in the draft submission even when the user had uploaded their own real photo. The fix checks for the presence of a real photo before uploading the AI image; if a real photo exists, the AI image is suppressed.

## Test
iznik-nuxt3/tests/unit/stores/member.spec.js — `cd /home/edward/FreegleDockerWSL/iznik-nuxt3 && npx vitest run --reporter=verbose tests/unit/stores/member.spec.js`

Two new tests were added under `fetchMembers — stale-entry counter bug (regression #9631 posts 19-20)`:
- `counter resets to 0 when API returns empty — even with stale Related pair in store`
- `non-synthetic Related member entry is not double-counted in the pair tally (N pairs → count equals N)`

Both tests confirmed failing before the fix and passing after. Full vitest suite (574 files, 12177 tests) passes.

### DIFF
--- a/iznik-nuxt3/modtools/stores/member.js
+++ b/iznik-nuxt3/modtools/stores/member.js
@@ -112,6 +112,16 @@ export const useMemberStore = defineStore({
         received += members.length
 
         if (params.collection === 'Related') {
+          // Clear stale Related entries from a previous page visit before processing
+          // the current API response. Without this, remainingPairs is inflated by
+          // entries that were never cleared via store.clear() on navigation, causing
+          // the counter to stay stuck at 1 even when the server returns no pairs.
+          Object.keys(this.list).forEach((key) => {
+            if (this.list[key].collection === 'Related') {
+              delete this.list[key]
+            }
+          })
+
           // V2 API returns {id, user1, user2} pairs.  Store each pair keyed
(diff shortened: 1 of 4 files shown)


edwh:
Closing: this fix only clears the client-side store on navigation, but the bug persists over a page refresh, so the root cause is server-side. This fix is insufficient.

PR #406, opened 2026-05-08
Trash Nothing Sync Port Testing

fnnbrr:
## Summary

- This PR adds `TNSyncCommand` to `iznik-batch` as a Laravel replacement for the v1 API's `tn_sync.php`. It is a direct port that should have identical functionality.
- All DB writes in `TNSyncCommand` (and dependent code) are also done using Eloquent model methods with [Laravel Auditing](https://github.com/owen-it/laravel-auditing).
- TNSyncCommand can be run with `--dry-run` to disable all DB writes and email sending. Since running with this flag should not perform any writes, any resulting rows in the `audits` DB table indicate accidental writes.
- TNSyncCommand can be run with `--local-testing` to read from local fixture files instead of the live Trash Nothing API. See below for using `run_seeded_sync.php` for local testing.

## Test plan

The `iznik-server/test/integration/tn_sync/run_seeded_sync.php` script will run a dummy TN sync using local data and output to log files. The logs prefixed with `TN-SYNC-TRACE` should be identical between old and new versions.

The locations of the log files are:
- `iznik-server`: `tmp/tn_sync_seeded_<timestamp>.log`
- `iznik-batch`: `var/www/html/storage/logs/cron/queue_background-tasks.log`
  - Normally this log file is overwritten, but for the purposes of this test, logs will be appended so they can be inspected

## Remaining items

- ~~CircleCI builds are failing, something's going wrong with installing Laravel Auditing.~~ **Fixed** — root cause was `bootstrap/cache/` files committed to git with a stale content-hash after a merge, compounded by BuildKit's stale `COPY . .` layer on the self-hosted runner and a persistent bind-mounted `vendor/` from a prior CI run on a different branch. Fixed in commits `ec258024`, `8be79595`, `c52bd29b`, `2968c57e` + orb v1.1.253.

- SQL for creating the `audits` table (see [this migration](https://github.com/owen-it/laravel-auditing/blob/master/database/migrations/audits.stub)):

```sql
CREATE TABLE `audits` (
  `id`             BIGINT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
  `user_type`      VARCHAR(255) NULL,
  `user_id`        BIGINT UNSIGNED NULL,
  `event`          VARCHAR(255) NOT NULL,
  `auditable_type` VARCHAR(255) NOT NULL,
  `auditable_id`   BIGINT UNSIGNED NOT NULL,
  `old_values`     TEXT NULL,
  `new_values`     TEXT NULL,
  `url`            TEXT NULL,
  `ip_address`     VARCHAR(45) NULL,
  `user_agent`     VARCHAR(1023) NULL,
  `tags`           VARCHAR(255) NULL,
  `created_at`     TIMESTAMP NULL,
  `updated_at`     TIMESTAMP NULL,
  INDEX `audits_auditable_type_auditable_id_index` (`auditable_type`, `auditable_id`),
  INDEX `audits_user_id_user_type_index` (`user_id`, `user_type`)
) DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
```

- Run on yesterday system live together


edwh:
@fnnbrr I asked Claude "do an adversarial review of this PR".  At least a couple of these look plausible so I'll stop thinking there and ask you to have a look at them.

 Critical

  1. bootstrap/cache/ files should not be in git — and services.php registers AuditingServiceProvider twice

  services.php and packages.php are generated, environment-specific files. They caused every CI failure on this
  branch (three separate pipeline failures requiring orb changes). Worse, services.php registers
  OwenIt\Auditing\AuditingServiceProvider at indices 30 and 35 in providers, and at 15 and 20 in eager —
  double-initialisation from a bad merge. These belong in .gitignore, not in git.

  2. Sixteen tests unconditionally skip — skipIfTNSyncWritesDisabled() has no condition

  private function skipIfTNSyncWritesDisabled(): void
  {
      $this->markTestSkipped(self::DB_WRITES_DISABLED_SKIP_REASON);
  }

  No flag, no env var, no config check — nothing that enables these tests without deleting the call. All 16 skips
  cover the most critical behavior: rating create/update/delete, reply-time upserts, about-me upserts, account
  removal, name changes, and date-file writes. The suite looks comprehensive but verifies almost none of the sync
  logic.

  3. test_stores_max_change_date is incorrectly marked write-dependent

  storeSyncDate() writes to a file. The date-file test is skipped via skipIfTNSyncWritesDisabled() even though it
  has zero dependency on DB writes. It would pass today with writes disabled. Date-propagation is entirely
  untested despite the test existing.

  4. tn_sync.php contains a silent behavioral change disguised as a logging addition

  -if ($loc['id'] !== $u->getPrivate('locationid')) {
  +if ($loc['id'] !== $u->getPrivate('lastlocation')) {

  This changes which column is compared — a logic fix, not log instrumentation. The error log on the same line
  still reads {$u->getPrivate('locationid')} for old_loc, so after this change the log always shows null/wrong for
   the old location value.

  5. ~~PR description has a stale CI-failing TODO~~ — fixed, updated above.

  ---
  Significant

  6. Two independent 300-second polling loops in tn_sync.php

  waitForTaskProcessed(300s) then waitForTNSyncCompletionData(300s) = up to 10 minutes blocking inside a cron
  script. If tn:sync crashes after processed_at is set but before writing completion data, tn_sync.php blocks for
  the full extra 300 seconds. The combined timeout should be a single budget, not two independent ones.

  7. markQueueRunCompleted() is an untracked real DB write during port-testing runs

  The PR says "any rows in the audits table indicate accidental writes." But markQueueRunCompleted() writes to
  background_tasks via DB::table(), bypassing Eloquent and therefore never touching audits. Every port-testing run
   writes to the production DB. The "zero audits = zero writes" invariant is not accurate as stated.

  8. test_loki_logs_user_merge verifies only key existence, not values

  \Mockery::on(fn($ctx) => isset($ctx['merge_to']) && isset($ctx['merge_from']))

  Passes as long as the keys exist with any value. A merge that logs the wrong user IDs would pass.

  ---
  Minor

  9. Copy-paste bug in tn_sync.php user-changes catch block is carried into this PR unfixed

  error_log("Ratings sync failed " . $e->getMessage() . " " . var_export($rating, TRUE));

  This is inside the $changes foreach but references $rating (last value from the outer ratings loop, or
  undefined). Should be $change. The PR adds logging around this catch block but the bug is carried forward
  unchanged.

  10. user-update Loki event fires for every TN user even when nothing changed

  The call at line 474 is unconditional — it fires for any change payload that reaches a TN user, even if the
  payload had no reply_time, about_me, username, or location fields. The legacy tn_sync.php has no equivalent
  event. Log comparison between old and new will have systematic noise.

  11. Plan files merged into master

  .github/prompts/plan-tnSyncShadowTesting.prompt.md and plans/tn-integration-refactor/ are planning artifacts.
  They become stale immediately after merge and belong in issues or a wiki, not the codebase.

  12. NULL / null inconsistency between syncRatings and syncUserChanges

  $maxDate = NULL; in one, $maxDate = null; in the other. Not a bug but inconsistent style in adjacent methods.

edwh:
@fnnbrr still battling CircleCI despite what I said, hence the flurry of commits.

I notice you've got a hand-rolled flock approach.  Are you aware of `withoutOverlapping`?

edwh:
@fnnbrr Ok, CI runs now.

Not had chance to fully review yet, but I don't quite get the `skipIfTNSyncWritesDisabled` approach to tests.  

- I think this means you're skipping tests in CI, which we don't want.  In tests we can and should write to the DB.
- Could we have a "dry-run" parameter?  I've had good success in telling Claude to do dry-runs, review the output, fix bugs, then run and monitor the live one.  So we can use that approach in conjunction with the existing `tn_sync` job.

Let me know if I've got the wrong end of the stick here and need to look at it more carefully before commenting.

fnnbrr:
> I notice you've got a hand-rolled flock approach. Are you aware of `withoutOverlapping`?

No I must've missed that, I'll leave myself a note to use `withoutOverlapping` to keep things consistent.

fnnbrr:
@edwh the writes are commented out on this `tn-sync-port-testing` branch, but kept intact in the `tn-integration-refactor` branch that this one is based on. Toggling the writes with a dry-run parameter does sound better than the separate branch approach, though, so you can go ahead with that instead, or I can do it on Wednesday.

A summary of the branches I have for this refactor, just for clarity:
- [`tn-integration-refactor`](https://github.com/Freegle/Iznik/tree/tn-integration-refactor): The final port that will be merged into master. TNSyncCommand is run by Laravel's scheduler and actually writes to the DB.
- [`tn-sync-port-testing`](https://github.com/Freegle/Iznik/tree/tn-sync-port-testing): DB writes in TNSyncCommand (and calling code) is commented out. Both old and new versions have logging prefixed with `TN-SYNC-TRACE` that should be identical between versions.
- [`tn-sync-port-testing-local`](https://github.com/Freegle/Iznik/tree/tn-sync-port-testing-local): Runs the same logic as `tn-sync-port-testing`, but uses local files containing dummy TN API results instead of actually hitting the API.

fnnbrr:
> I notice you've got a hand-rolled flock approach. Are you aware of `withoutOverlapping`?

Ah, now I remember why I needed to do this: When `TNSyncCommand` is run by the Laravel scheduler in `console.php`, I can use `withoutOverlapping`, but when it's run via `tn_sync` which queues it up as a task, `withoutOverlapping` isn't available. I need to use `PreventsOverlapping` in that case because I still don't want overlapping runs, even when it's run via `tn_sync`.

I'll leave a comment to get rid of that extra flocking code once testing is complete, since then `TNSyncCommand` will only be run via the Laravel scheduler where `withoutOverlapping` will apply.

iznik-batch/routes/console.php
```diff
@@ -350,7 +359,7 @@ function cronLog(string $command): string
 // Runs continuously with internal looping. Handles push notifications and emails.
 Schedule::command('queue:background-tasks --max-iterations=60 --spool')
     ->everyMinute()
-    ->sendOutputTo(cronLog('queue:background-tasks'))
+    ->appendOutputTo(cronLog('queue:background-tasks'))
```
fnnbrr, line 380: TODO: this is temporary to view `TN-SYNC-TRACE` logs during testing. Revert before merging in.

fnnbrr: TBD potentially keep this. Need to make sure Laravel is using log rotation. Should also have a log file per background task instead of routing all of their outputs to one file.

fnnbrr: Will be addressed in #456

PR #443, opened 2026-05-12
fix(ai-images): remove AI image attachment when user adds their own photo

edwh:
## Root Cause
The AI image generation/attachment code path does not check (or checks at the wrong lifecycle point) whether the message already has user-uploaded attachments before generating and attaching an AI-suggested image. Result: both AI image and member's own photo coexist on the same message.

## Evidence
```
=== RUN   TestMessagePatch_AIImageRemovedWhenUserAddsPhoto
--- FAIL: TestMessagePatch_AIImageRemovedWhenUserAddsPhoto (0.24s)
    message_ai_image_coexist_test.go:75: 
        Error Trace:    /app/test/message_ai_image_coexist_test.go:75
        Error:          Not equal: 
                        expected: int64(1)
                        actual  : int64(2)
        Test:           TestMessagePatch_AIImageRemovedWhenUserAddsPhoto
        Messages:       message should have exactly 1 attachment after user adds their own photo (AI should be removed)
    message_ai_image_coexist_test.go:79: 
        Error Trace:    /app/test/message_ai_image_coexist_test.go:79
        Error:          Not equal: 
                        expected: "freegletusd-user2-aicoexist2_1778595533074543781"
                        actual  : "freegletusd-ai2-aicoexist2_1778595533074543781"
        Test:           TestMessagePatch_AIImageRemovedWhenUserAddsPhoto
        Messages:       only the user's own photo should remain, not the AI attachment
    message_ai_image_coexist_test.go:83: 
        Error Trace:    /app/test/message_ai_image_coexist_test.go:83
        Error:          Not equal: 
                        expected: int64(0)
                        actual  : int64(1)
        Test:           TestMessagePatch_AIImageRemovedWhenUserAddsPhoto
        Messages:       AI attachment must be removed when user adds their own photo
```

## Fix
Added `filterOutAIAttachments` helper in `message/message.go` (near the bottom of the file, before `isAIAttachment`). Called from `applyPatchMessage` (PATCH handler) before `recordAIDeletions`:

When the PATCH keep-list contains both AI-generated attachment IDs and user-supplied attachment IDs, `filterOutAIAttachments` queries `messages_attachments` for each ID's `externalmods`, classifies them as AI or non-AI, and strips the AI IDs from the keep-list. The subsequent `DELETE FROM messages_attachments WHERE msgid = ? AND id NOT IN (?)` then removes the AI attachment from the database.

The function only filters when BOTH kinds are present — if all attachments are AI-only or all are non-AI, the original list is returned unchanged, preserving existing behaviour.

Also forward-ports the empty-item validation (`if strings.TrimSpace(req.Item) == ""`) from master commit `27c12ac5` which this branch had not yet incorporated, restoring `TestPutMessageEmptyItemRejected` to passing.

## Alternatives Investigated
- Race / order-of-operations (AI suggestion fires before user image upload completes): deferred — fix instead enforces the invariant at attach-time, so ordering becomes irrelevant.
- Display-layer concatenation (both kinds rendered without filtering): ruled less likely by reproduction — the AI attachment is persistent in DB, not just rendered.

## Other Call Sites
No other occurrences found. The only Go code path that processes `req.Attachments` in the PATCH handler is `applyPatchMessage`. AI attachments are inserted by the PHP cron (`messages_illustrations.php`) which already has a pre-insert check; the Go fix adds defence-in-depth at the API layer.

## Test
iznik-server-go/test/message_ai_image_coexist_test.go
Command: `POST http://localhost:8081/api/tests/go {"filter":"TestMessagePatch_AIImageRemovedWhenUserAddsPhoto","timeout":120}`
Proves: test FAILS before this commit (see Evidence above), PASSES after (see TEST_PASSED output below)
Regression suite: iznik-server-go full go test ./... — All tests passed (2398✓)

TEST_PASSED:
```
=== RUN   TestMessagePatch_AIImageRemovedWhenUserAddsPhoto
--- PASS: TestMessagePatch_AIImageRemovedWhenUserAddsPhoto (0.18s)
```

## Discourse
https://discourse.ilovefreegle.org/t/9646

### DIFF
--- a/iznik-batch/tests/Feature/Message/AIImageIllustrationRaceConditionTest.php
+++ b/iznik-batch/tests/Feature/Message/AIImageIllustrationRaceConditionTest.php
@@ -1,0 +1,168 @@
+<?php
+
+namespace Tests\Feature\Message;
+
+use App\Models\Message;
+use App\Models\MessageAttachment;
+use App\Models\MessageGroup;
+use Tests\TestCase;
+
+/**
+ * AssertFlip Test: AI images added to posts with member photos
+ *
+ * Bug: When a member uploads a photo after the illustrations script checks
+ * for attachments but before it inserts the AI image, the post ends up with
+ * BOTH the member photo AND the AI illustration. The cleanup script should
+ * remove the AI image but currently does not prevent the race condition.
+ *
+ * STEP 1: Test asserts the BUGGY behaviour (both images exist).
+ * STEP 2: Inverted test asserts the FIXED behaviour (only member photo exists).
+ */
+class AIImageIllustrationRaceConditionTest extends TestCase
+{
+    public function test_message_with_both_member_photo_and_ai_illustration_buggy_behavior(): void
+    {
+        // Create a message with no attachments
+        $user = $this->createTestUser();
+        $group = $this->createTestGroup();
+        $this->createMembership($user, $group);
+
+        $message = Message::create([
+            'type' => Message::TYPE_OFFER,
+            'fromuser' => $user->id,
+            'subject' => 'OFFER: Test Item (Location)',
+            'textbody' => 'Test message.',
+            'source' => 'Platform',
+            'date' => now(),
+            'arrival' => now(),
+            'lat' => $group->lat,
+            'lng' => $group->lng,
+        ]);
+
+        MessageGroup::create([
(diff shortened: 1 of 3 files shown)


edwh:

PR #445, opened 2026-05-12
fix(modtools): $recentwanted macro must show only approved posts with original post date

edwh:
## Root Cause
The $recentwanted substitution pipeline (modtools stock-reply / messagehistory path) was including non-Approved rows. The `getPublicHistory` function in `User.php` used the caller-supplied `$msgcoll` parameter to build the collection WHERE clause. When modtools called `getPublicsById`, `Message.php` passed `[Pending, Approved, Spam]`, causing rejected/pending WANTED posts to appear in messagehistory and therefore in the `$recentwanted` macro output.

The timestamp was already using `messages.arrival` (original post date) correctly in both PHP and Go paths — no timestamp fix was needed.

## Evidence
```
✘ Recent wanted excludes non approved
  │ Rejected WANTED must NOT appear in messagehistory (only Approved posts belong in $recentwanted)
  │ Failed asserting that an array does not contain 'WANTED: UT RW rejected item'.
  │ /var/www/iznik/test/ut/php/include/RecentWantedMessageHistoryTest.php:113
Tests: 1, Assertions: 10, Failures: 1.
```

## Fix
`getPublicHistory` in `include/user/User.php` now hardcodes `MessageCollection::APPROVED` for the collection WHERE clause, ignoring the caller-supplied `$msgcoll` parameter. This ensures messagehistory always contains only publicly-visible approved posts regardless of what the caller passes.

Changed: `messages_groups.collection IN ('...caller list...')` → `messages_groups.collection = 'Approved'`

Also fixes a pre-existing test bug: `SendTestPushCommandTest` was inserting `'created' => now()` but the `users_push_notifications` table column is `'added'` (causing 6 Laravel test errors).

## Alternatives Investigated
- Client-side macro expansion: ruled out — wrong timestamp matches server-side messages_groups.arrival update, which a client without DB access cannot produce.
- NOW()/PHP date() placeholder: ruled out — Michael reports the timestamp is stable post-rejection (a stored column signature, not a live clock).

## Other Call Sites
`getPublicHistory` is only called from `getPublics` (User.php:2732). The only callers previously passing non-APPROVED collections were `Message.php` line 1544 (`[Pending, Approved, Spam]` in modtools mode) — now always filtered to Approved in messagehistory. No other occurrences found.

## Test
iznik-server/test/ut/php/include/RecentWantedMessageHistoryTest.php
Command: docker exec freegle-apiv1 php /var/www/iznik/composer/vendor/phpunit/phpunit/phpunit --configuration /var/www/iznik/test/ut/php/phpunit.xml --filter RecentWantedMessageHistoryTest --test-suffix RecentWantedMessageHistoryTest.php /var/www/iznik/test/ut/php/include
Proves: test FAILS before this commit (see Evidence above), PASSES after (OK (1 test, 11 assertions))
Regression suite: PHP Laravel suite 2347/2347 passed — SUITE_PASSED=yes
UserTest: 92 tests passing, MessageTest: 43 tests passing

Note: Go API test suite shows 2 pre-existing failures unrelated to this PHP change:
- TestMessagePatch_AIImageRemovedWhenUserAddsPhoto (AI image bug, separate issue)
- TestHoldActionMisidentifiedAsSpamReport_BugAssert (AssertFlip STEP1, correctly fails after hold/spam bug was fixed)

## Discourse
https://discourse.ilovefreegle.org/t/9655

### DIFF
--- a/iznik-batch/tests/Unit/Commands/Notification/SendTestPushCommandTest.php
+++ b/iznik-batch/tests/Unit/Commands/Notification/SendTestPushCommandTest.php
@@ -176,7 +176,7 @@ class SendTestPushCommandTest extends TestCase
             'apptype' => $apptype,
             'type' => 'FCMAndroid',
             'subscription' => 'test_token_' . bin2hex(random_bytes(16)),
-            'created' => now(),
+            'added' => now(),
         ]);
     }
 }
--- a/iznik-server/include/user/User.php
+++ b/iznik-server/include/user/User.php
@@ -2199,7 +2199,9 @@ class User extends Entity
             $sql = null;
 
             if (count($idsleft)) {
-                $collq = " AND messages_groups.collection IN ('" . implode("','", $msgcoll) . "') ";
+                // messagehistory is used for $recentwanted and similar macros — only Approved
+                // (publicly visible) posts belong here regardless of what the caller requested.
+                $collq = " AND messages_groups.collection = '" . MessageCollection::APPROVED . "' ";
                 $earliest = $historyfull ? '1970-01-01' : date('Y-m-d', strtotime("midnight 30 days ago"));
                 $delq = $historyfull ? '' : ' AND messages_groups.deleted = 0';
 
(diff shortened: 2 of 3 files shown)


edwh:

PR #465, opened 2026-05-17
fix(modtools): setDeep null crash and Firefox profile upload (9518/293, 9518/294)

edwh:
## Summary

- **Bug 9518/294 — Microvolunteering settings 500 error**: `setDeep` in `ModGroupSetting.vue` only checked `typeof a[b] === 'undefined'` for missing intermediate values, not `null`. When `microvolunteeringoptions` is `null` in the DB, saving a nested setting (e.g. `photorotate`) crashed with "Cannot set properties of null". Fixed by checking `a[b] == null` (covers both `undefined` and `null`) in both the `setrecursively` branch and the recursive descent path.

- **Bug 9518/293 — Firefox group profile upload silent failure**: After selecting a file via the OS file picker in Uppy's DashboardModal, nothing happened in Firefox. Root cause: Firefox doesn't reliably fire `change` events on Uppy's internal file input in certain rendering contexts. Added a `useNativeInput` prop to `OurUploader` that renders a plain `<input type="file">` element wired directly to the existing tus+imageStore upload pipeline, bypassing Uppy's dashboard entirely. `ModSettingsGroup` now uses `use-native-input` for the group profile photo upload.

## Code Quality Review

- No duplication: `useNativeInput` reuses the existing `uploadOneFile` and `imageStore.post` path
- No regressions: Uppy dashboard mode unchanged for all other upload sites
- Tests: 6 new Vitest tests cover `useNativeInput` rendering, file acceptance, and upload emit

## Test Plan

- [x] Vitest: 12,218 tests pass (including 6 new `useNativeInput` tests)
- [ ] Manual: ModTools group settings — save a microvolunteering setting for a group with null `microvolunteeringoptions`
- [ ] Manual: ModTools group profile — upload a photo in Firefox (should now work via native file picker)

### DIFF
--- a/iznik-nuxt3/components/OurUploader.vue
+++ b/iznik-nuxt3/components/OurUploader.vue
@@ -1,45 +1,67 @@
 <template>
   <client-only>
     <div class="wrapper" @dragenter="onDragEnter">
-      <div class="d-flex flex-column justify-content-around">
-        <v-icon
-          v-if="!busy"
-          :size="iconSize"
-          icon="camera"
-          class="camera text-faded"
-        />
+      <template v-if="useNativeInput">
         <Spinner v-if="busy" :size="50" class="fadein" />
-        <div
-          v-else-if="multiple || !modelValue.length"
-          class="d-flex justify-content-around"
-        >
+        <div v-else>
+          <input
+            ref="nativeFileInput"
+            type="file"
+            accept="image/*"
+            class="d-none"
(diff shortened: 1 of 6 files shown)


edwh:
setDeep null fix applied directly to master (commit 12e20fec9). Firefox useNativeInput bypass dropped — the symptom was the PATCH not persisting the profile field (fixed by #464), not a genuine Firefox/Uppy bug.

PR #467, opened 2026-05-17
feat(donations): personalise suggested amount by member tenure

edwh:
## Summary

- Computes a `tenureDefault` from `me.added` in the auth store (no backend change needed)
- Passes it as the `:default` prop to `DonationAskStripe`, replacing the hardcoded `1`
- In `DonationAskStripe`, applies `Math.max(banditAmount, props.default)` so tenure acts as a floor on the A/B bandit's suggestion — the bandit still controls presentation style and can suggest higher amounts, but never lower than the tenure floor

**Suggested donation tiers by tenure:**
| Tenure | Default starting amount |
|--------|------------------------|
| < 6 months | £2 |
| 6 months – 2 years | £3 |
| 2+ years | £5 |

**Backed by production data (last 2 years, PayPal + Stripe donations):**
- Members <1 month: avg £1.92 · Members 5+ years: avg £3.06 (+60%)
- Users with 31+ prior posts: avg £3.03 vs £2.04 for low-activity users
- Users with 21–50 prior actions are nearly 2× as likely to donate >£5 (15% vs 7%)

## Code Quality Review

- No new dependencies; uses existing `authStore.user.added` already in the session
- Tenure calculation is pure and side-effect-free
- Bandit experiment is unaffected; tenure is a floor not an override
- `defineExpose` added only for the values the test needs

## Test Plan

- [x] 6 new unit tests covering all tenure buckets and the floor behaviour
- [x] All 12,220 existing Vitest tests pass
- [ ] Check donation modal at different simulated tenures

### DIFF
--- a/iznik-nuxt3/components/DonationAskModal.vue
+++ b/iznik-nuxt3/components/DonationAskModal.vue
@@ -30,8 +30,7 @@
             :raised="raised"
             :target-met="targetMet"
             :donated="donated"
-            :amounts="[1, 5, 10]"
-            :default="1"
+            :default="tenureDefault"
             @score="score"
             @success="thankyou = true"
             @cancel="hide"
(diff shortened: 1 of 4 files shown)


edwh:
Superseded by engagement-level approach — uses existing users.engagement field (Obsessed/Frequent/Occasional etc.) rather than tenure. New PR incoming.

PR #479, opened 2026-05-18
fix(e2e): wait for domcontentloaded after logout redirect to prevent renderer freeze

edwh:
## Summary

- `logoutIfLoggedIn` in `user.js` now calls `waitForLoadState('domcontentloaded', { timeout: 30000 })` after the logout redirect lands at `'/'`
- `logoutIfLoggedIn.spec.js` updated to use `vi.resetModules()` + async `import()` so `vi.mock()` is applied before user.js loads

**Bug**: When logout redirects to `'/'`, the function skipped `goto('/')` to avoid a double-navigation race, but `page.url()` can show `'/'` before `domcontentloaded` fires. If the caller immediately navigates elsewhere (e.g. via postMessage to `'/give'`), two concurrent navigations cause a 35s+ V8 freeze detected by the heartbeat mechanism.

## Test Plan

- Updated `logoutIfLoggedIn.spec.js` covers the skip-path and verifies `waitForLoadState` is called
- Vitest suite passes

### DIFF
--- a/iznik-nuxt3/tests/e2e/utils/user.js
+++ b/iznik-nuxt3/tests/e2e/utils/user.js
@@ -228,8 +228,31 @@ async function logoutIfLoggedIn(page, navigateToHome = true) {
 
       if (currentPathIsHome) {
         console.log(
-          '[logoutIfLoggedIn] Logout redirect already at home page, skipping explicit navigation'
+          '[logoutIfLoggedIn] Logout redirect already at home page, waiting for load state...'
         )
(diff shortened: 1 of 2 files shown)


edwh:
Closing — this waitForLoadState approach in the skip-path is the wrong fix. The original intent of the skip-path (f89d99a43) was to return immediately with no wait; adding a 30s domcontentloaded wait here risks CI hangs and doesn't solve the caller-navigation race correctly. Need actual CI artifacts before attempting another fix.

PR #483, opened 2026-05-18
fix(avatar): persist new profileid after OurUploader completion in ProfileSection (#9679/3)

edwh:
## Summary

- After the user uploads a new profile picture via `OurUploader`, `ProfileSection.vue` was not persisting the returned `profileid` back to the user store
- The component emitted the completion event but did not update `authStore.user.profile`, so the new avatar was not reflected until a page reload

## Root Cause

`OurUploader` calls an `@uploaded` callback with the new image data including `profileid`. `ProfileSection.vue` was not consuming this to update the store, so the in-memory `user.profile` remained stale.

## Test Plan

- Vitest unit tests via AssertFlip pattern on the component
- Upload a new profile picture in ModTools Settings → Profile — verify it shows immediately without reload

### DIFF
--- a/iznik-nuxt3/components/settings/ProfileSection.vue
+++ b/iznik-nuxt3/components/settings/ProfileSection.vue
@@ -264,13 +264,17 @@ watch(
   async (newVal) => {
     uploading.value = false
     if (newVal?.length) {
-      const atts = {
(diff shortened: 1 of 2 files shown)


edwh:
Bad fix.

PR #484, opened 2026-05-18
fix(eee-browser): reset submitting flag so next item is interactive after classification

edwh:
## Summary

- After submitting a label, `submitting` was set to `true` but only reset in the `catch` block — never in the success path
- This left all classification buttons permanently disabled after the first successful label, so the next item loaded but was unclickable
- Fix: reset `submitting.value = false` before calling `loadNextItem()` in the success timeout

## Test Plan

- [ ] Classify an item — confirm the next item loads and buttons are enabled
- [ ] Classify several items in a row — confirm each advances correctly
- [ ] Simulate a network error — confirm buttons re-enable after the error


edwh:
Closing — superseded by a clean single-commit PR targeting the feature branch

PR #488, opened 2026-05-18
fix(modtools-push): make zero-count Android clear truly silent

edwh:
## Summary
- Zero-count ModTools pushes were leaving empty entries in the Android system tray on the new Capacitor app.
- Commit `af3d96d32` attached `AndroidConfig.notification.tag` to every ModTools Android send. Setting `AndroidConfig.notification` on an otherwise data-only FCM payload can promote it to a notification message on some Android builds, surfacing the empty entry.
- Extracted `buildAndroidConfig()` and only attach the tag (and use high priority) when there's actually a visible title to render. Empty-title pushes now go out as pure data-only on normal priority and stay silent.
- Paired with the Capacitor-side cleanup on `feat/modtools-capacitor-app` (commit 7d0643fcc) which wires `removeAllDeliveredNotifications()` into the zero-count handler so stale "N pending" entries are cleared from the tray when work is handled.

## Test plan
- [ ] CI passes (existing + 3 new unit tests covering `buildAndroidConfig` for visible/silent/forceVisible cases).
- [ ] On a real ModTools Android device with the new Capacitor build, trigger a zero-count push (e.g. open a group with pending work, then have another mod approve it). Confirm no empty tray entry appears and any prior "N pending" entry is cleared.

### DIFF
--- a/iznik-batch/app/Services/PushNotificationService.php
+++ b/iznik-batch/app/Services/PushNotificationService.php
@@ -430,6 +430,34 @@ class PushNotificationService
     }
 
     /**
+     * Build the AndroidConfig (priority, ttl, optional notification tag)
+     * that accompanies the FCM message.
+     *
+     * Zero-work ModTools pushes carry empty title/message — they exist only
+     * to clear the launcher badge and must stay silent. We achieve that by
+     * sending pure data-only at normal priority: no notification block, and
+     * no AndroidConfig.notification (which can promote the message to a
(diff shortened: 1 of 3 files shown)


edwh:

edwh:
Reopening — this PR was incorrectly closed by the monitor FSM. It is unrelated to the message deletion logging bug (topic 9685); it fixes Android zero-count push notifications.

PR #491, opened 2026-05-18
fix(android): handle Sentry 502 API errors without failing the build

edwh:
## Root Cause

CircleCI job [build-android #15093](https://circleci.com/gh/Freegle/Iznik/15093) failed at the **Build Nuxt App** step with:

```
[plugin sentry-release-management-plugin]
Error: Command failed: sentry-cli releases new af407e6eb8a2cb0b62473f97386fc0138f867351
error: API request failed

Caused by:
    sentry reported an error: bad gateway (http status: 502)
```

The production Android build sets `ISAPP=true` and `APP_ENV=production`. In `nuxt.config.ts`, the Sentry Vite plugin for this path had `errorHandler: undefined` when `SENTRY_STRICT=true` (the default, since the env var is not explicitly set in the build-android CI job). The default handler throws on **any** error, so a transient 502 from Sentry's API failed the entire build and blocked the app release.

## Fix

Replace `errorHandler: undefined` with a handler that:
- **Catches** API errors (502, 504, bad gateway, "API request failed") → logs a warning, does not throw
- **Throws** for all other errors (wrong auth token, wrong org/project, etc.) so real misconfigurations still fail loudly

The app binary is the critical artifact; Sentry release creation is supplementary metadata.

## Files Changed

- `iznik-nuxt3/nuxt.config.ts` — updated Sentry plugin error handler for the `ISAPP && production` build path

## Test plan

- [ ] Verify build-android CI passes on the next production push
- [ ] Confirm other Sentry errors (e.g. auth failure) still fail the build

### DIFF
--- a/iznik-nuxt3/nuxt.config.ts
+++ b/iznik-nuxt3/nuxt.config.ts
@@ -485,9 +485,26 @@ export default defineNuxtConfig({
               org: 'freegle',
               project: 'capacitor',
               authToken: config.SENTRY_AUTH_TOKEN,
-              // For non-strict mode (debug builds), log errors but don't fail the build
+              // Never fail the build on Sentry API errors (502/504/bad gateway) —
+              // the app binary is the critical artifact; release creation is supplementary.
+              // In strict mode, still throw on non-API errors (wrong auth token, etc.).
               errorHandler: config.SENTRY_STRICT
-                ? undefined // Use default (throw on error)
+                ? (err) => {
+                    const msg = err.message || ''
+                    if (
+                      msg.includes('502') ||
+                      msg.includes('504') ||
+                      msg.includes('bad gateway') ||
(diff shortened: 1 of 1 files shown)


edwh:
## Production was blocked — fix applied directly to master

Build-android job #15093 on the production branch failed with the exact same Sentry 502 error this PR addresses, blocking the production release.

The same fix has been pushed directly to master (`ecaa29f45efe12c61a3e6520408aff2a430b8ba4`) under the CI-failure exception (critical CI block on master). This PR can still be merged if desired, but master already contains the fix.

**Root cause**: `sentry-cli releases new` returned HTTP 502 (bad gateway) from Sentry's API; the `errorHandler: undefined` (default throw) caused the entire build to abort rather than treating the supplementary release-creation step as non-fatal.

PR #507, opened 2026-05-21
fix(message): remove AI attachment and scrape TN photos when TN user edits post

edwh:
## Summary

- When a TN user edits their post via the TN app, TN sends `PATCH /message/tn/:tnpostid` with an updated textbody
- Previously, AI attachments were only removed when the client sent an explicit `attachments: []` array — which TN never does; the illustrations cron could re-inject the AI photo even after the user deleted it
- **Case 1 — No TN photos**: if the updated textbody contains no `trashnothing.com/pics/` links, treat any AI attachment as implicitly removed: call `recordAIDeletions()` and delete the attachment row
- **Case 2 — TN photos present**: if the updated textbody contains `trashnothing.com/pics/` links, ALSO remove the AI attachment AND scrape the TN /pics/ pages for direct image URLs, download and upload them to TUS, store as `messages_attachments` rows, and strip the "Check out the pictures…" block from the stored textbody
- In both cases `messages_ai_declined` is set to block cron re-injection
- Non-AI attachments are never touched by the TN edit path

## Code Quality Review

- No duplication — reuses `recordAIDeletions()` helper; `aiimage.ImageUploader` shared with the illustration pipeline
- Regexps compiled once at package level
- `TNPageFetcher`, `TNImageFetcher`, `TNPhotoScrapeRunner`, `ScrapeTNPhotosSync` are injectable vars; tests use stubs with no real HTTP or TUS calls
- Scraping runs via `TNPhotoScrapeRunner` (goroutine in production, synchronous in tests) so the PATCH response is not delayed by photo downloads
- Textbody stripping happens before `applyPatchMessageCore` so the DB always stores clean text

## Test Plan

- `TestPatchMessageByTnPostid_RemovesAIPhotoWhenTextbodyHasNoPhotoLinks`: PATCH with no photo links → AI removed + `messages_ai_declined` set
- `TestPatchMessageByTnPostid_RemovesAIAndScrapesTNPhotosWhenLinksPresent`: PATCH with TN pic links → AI removed + textbody stripped + scraped photo stored as attachment
- `TestPatchMessageByTnPostid_NonAIAttachmentPreservedOnTextOnlyEdit`: text-only PATCH with no AI attachment → real attachment untouched, no `messages_ai_declined` set
- Full suite: 2648 tests pass


edwh:
Reopening to fix PR diff — old base SHA was including unrelated monitor-fsm commits from master's history.

PR #518, opened 2026-05-24
fix(test): wait for /browse→/explore redirect via waitForURL before clicking logout

edwh:
## Summary

- **Root cause of production CI job 16194** (test 3.2 "can login and reply from Browse Page"): `/browse` auto-redirects new users (no group membership) to `/explore` via JS **after** DOMContentLoaded fires.
- The existing fix (`2028666e2`) added `waitForLoadState('domcontentloaded')` at the start of `logoutIfLoggedIn`, but that call resolves **immediately** when `/browse`'s DOMContentLoaded has already fired — leaving the JS-initiated `/browse→/explore` redirect still in flight.
- Clicking logout while that redirect is committing causes two simultaneous hard navigations that freeze the V8 renderer for 35+ seconds (heartbeat timeout → page closed → test fails).
- **Fix**: when the entry URL is `/browse`, explicitly call `waitForURL(/\/(explore|myposts)/)` to drain the redirect chain before searching for and clicking the logout button.
- Adds a Vitest unit test verifying the `waitForURL` call is made when entry is `/browse`.

## Test plan

- [x] All 7 Vitest unit tests in `logoutIfLoggedIn.spec.js` pass locally
- [x] New test "waits for /browse→/explore redirect via waitForURL when at /browse" passes
- [ ] CI build passes (Playwright test 3.2 should no longer freeze)

### DIFF
--- a/iznik-nuxt3/tests/e2e/utils/user.js
+++ b/iznik-nuxt3/tests/e2e/utils/user.js
@@ -130,6 +130,34 @@ async function logoutIfLoggedIn(page, navigateToHome = true) {
         .catch(() => {})
     }
 
+    // /browse auto-redirects new users (no group membership) to /explore via JS
+    // AFTER DOMContentLoaded fires. waitForLoadState above resolves immediately
+    // because /browse's DOMContentLoaded has already fired by the time
+    // logoutIfLoggedIn is called, leaving the JS-initiated /browse→/explore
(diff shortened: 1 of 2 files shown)


edwh:
Fix cherry-picked directly to master (093bf4687) to unblock production CI job 16194. This PR is now superseded.

PR #528, opened 2026-05-24
fix(ci): add missing branch filters to build-test-katapult workflow

edwh:
## Summary

- `build-test-katapult` is now the **default CI path** — `check-runner` sets `use_katapult=true` for all normal pushes, so this workflow runs on every branch push
- But `build-test-katapult` was missing the `ignore: production` and `ignore: modtools` branch filters that `build-test-local` and `build-test-cloud` already have
- This caused `freegle/build-and-test` to run on the production branch via a Katapult VM, generating `infrastructure_fail` (heartbeat timeout during the parallel test execution step)
- PR #525 fixed `build-test-local` and `build-test-cloud` but missed `build-test-katapult`

## Root cause

`build-test-katapult` comment said "triggered manually" but it has become the default path since `check-runner` now always routes to Katapult. Without branch filters, it runs on `production` and `modtools` which are deployment-only branches with no need to run the test suite.

## Fix

Added `filters.branches.ignore: [modtools, production]` to `freegle/build-and-test` and `mark-ci-passed` jobs in `build-test-katapult`, matching the pattern already established in `build-test-local` and `build-test-cloud`.

## Files changed
- `.circleci/continue-config.yml`

### DIFF
--- a/.circleci/continue-config.yml
+++ b/.circleci/continue-config.yml
@@ -260,7 +260,9 @@ workflows:
               only:
                 - production
 
-  # Katapult cloud runner test — triggered manually with use_katapult: true
+  # Katapult cloud runner test — the default CI path (use_katapult=true is set by check-runner
+  # for all normal pushes). Branch filters here must match build-test-local/build-test-cloud
+  # so production and modtools are skipped (they use deploy-apps and have no tests to run).
   build-test-katapult:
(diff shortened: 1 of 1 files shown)


edwh:
Superseded by direct push ed1d7ba5c to master — same fix applied to unblock production CI.

PR #545, opened 2026-05-25
fix(modtools): wire up chat-review Delete All to actually reject messages

edwh:
## Summary
- `deleteConfirmed` in `modtools/pages/chats/review.vue` had its `chatStore.reject()` call commented out behind a `// TODO: This needs to use the store properly`. As a result, clicking "Delete All" in the chat review screen, then confirming the modal, was a silent no-op — the modal closed and nothing got rejected.
- Uncommented the call so each non-`widerchatreview` message in `visibleMessages` is rejected via the chat store.
- Added two TDD regression tests in `tests/unit/pages/modtools/chats/review.spec.js`:
  - asserts `chatStore.reject({ id, chatid: null })` is called once per visible message
  - asserts wider-chat-review (mod-only) entries are skipped

## Test plan
- [x] Unit suite for the review page — both new tests pass, 16 tests in file green
- [x] Full unit suite via status API — 277 passed, 0 failed
- [ ] Manual smoke in modtools chat review once deployed: load review queue with >1 message, click Delete All, confirm, verify entries disappear

### DIFF
--- a/iznik-nuxt3/modtools/pages/chats/review.vue
+++ b/iznik-nuxt3/modtools/pages/chats/review.vue
@@ -157,8 +157,7 @@ function deleteAll(callback) {
 function deleteConfirmed() {
   visibleMessages.value.forEach((m) => {
     if (!m.widerchatreview) {
-      // TODO: This needs to use the store properly
-      // chatStore.reject({ id: m.id, chatid: null })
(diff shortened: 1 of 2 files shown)


edwh:
Commit 0e76394eb already on master (cherry-picked directly). This PR's branch is now empty.

PR #547, opened 2026-05-25
components: use Lazy-prefixed auto-import for ProfileModal/PromiseModal

edwh:
## Summary

- 12 components currently resolve `ProfileModal` / `PromiseModal` via `defineAsyncComponent(() => import('~/components/X'))`. The `~` alias points at the project root and ignores Nuxt's layer override system.
- Replaced the explicit imports with Lazy-prefixed bare tags in templates (`<LazyProfileModal/>`, `<LazyPromiseModal/>`). Nuxt auto-imports + the `Lazy` prefix preserves async / code-split loading AND routes through the component resolver, which honours layers.
- No behaviour change for single-layer (Freegle-only) builds — `LazyXxx` compiles to the same `defineAsyncComponent(() => import(...))` Nuxt would have produced.

## Why

When Freegle is consumed as a base layer (e.g. by a sister site that wants to skin Freegle's chat with a different `ProfileModal`), the layer's `components/ProfileModal.vue` is *not* picked up by callers using `import('~/components/ProfileModal')`. The Lazy-prefixed bare tag fixes this without changing Freegle's bundle behaviour.

## Files touched

`components/`: ChatFooter, ChatHeader, ChatMessageInterested, ChatMessagePromised, ChatPane, MessageExpanded, MessageHistoryExpanded, MyMessage, MyMessagePromisedTo, MyMessageReply, NewsReply, NewsUserIntro.

## Test plan

- [ ] Build the existing Freegle nuxt3 site and verify the chat pane, message expanded view, and news reply UI still open the profile / promise modals.
- [ ] `npm run test` — Playwright e2e suite passes.
- [ ] No new unit-test regressions in `tests/unit/components/{Chat*,Message*,My*,News*}.spec.js`.

## Future work

There are other `defineAsyncComponent(() => import('~/components/X'))` calls in this codebase (ChatBlockModal, MessageMap, OurAtTa, ExternalDa, PostMap, PostItem, ExportPost, ExportChat). Same anti-pattern but those components don't currently have layer overrides upstream, so this PR is scoped to just ProfileModal/PromiseModal. The pattern shown here is the recommended refactor for the remainder.

### DIFF
--- a/iznik-nuxt3/components/ChatFooter.vue
+++ b/iznik-nuxt3/components/ChatFooter.vue
@@ -304,7 +304,7 @@
         />
       </div>
     </div>
-    <PromiseModal
+    <LazyPromiseModal
       v-if="showPromise"
       :messages="ouroffers"
       :selected-message="likelymsg ? likelymsg : 0"
@@ -314,7 +314,7 @@
       @hide="fetchMessages"
       @hidden="showPromise = false"
     />
-    <ProfileModal
+    <LazyProfileModal
       v-if="showProfileModal && otheruser"
       :id="otheruser ? otheruser.id : null"
       @hidden="showProfileModal = false"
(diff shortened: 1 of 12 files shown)


edwh:
Replaced by #553 (same-repo branch so CircleCI runs).

PR #570, opened 2026-05-27
feat(mjml): compile-once / substitute-many cache for batch mailables

edwh:
## Summary

Adds a compile-once / substitute-many primitive for batch-send MJML
mailables. Cuts MJML sidecar load when the same content fans out to many
recipients with small per-user variation (immediate digest, newsletter,
mod-team chase, per-group event/volunteering digests, story asks).

Same pattern AWS SES `SendBulkEmail`, Postmark Bulk API, and Mailchimp's
merge tags use internally — verified that the `{{varname}}` placeholder
syntax survives MJML sidecar compile in text content and `href`/`src`/`alt`
attributes (the only places we place placeholders).

## Architecture

- `App\Services\BulkMail\BulkMjmlCompiler`: process-local cache + substitute + validate.
- `App\Mail\Concerns\BulkRenderable` interface: opt-in contract for mailables
  (4 methods: `shapeKey`, `bulkTemplate`, `bulkData`, `mergeVars`).
- `App\Mail\MjmlMailable::setPrerenderedHtml()`: lets a caller hand the
  Mailable a pre-rendered HTML body; `mjmlView()` short-circuits the
  Blade+MJML pipeline when one is set.
- Feature flag `freegle.bulk_mail.enabled` (default `true`) — roll-back hatch
  if a regression slips through.
- Fail-loud validation: any unbound `{{var}}` left in the substituted HTML
  raises `UnboundPlaceholderException` (no garbage emails go out).

## Migrated

| Mailable | Service / call site | Multiplier |
|---|---|---|
| `UnifiedDigest` (immediate) | `UnifiedDigestService::processGroupImmediate` | 1 message → N group members |
| `StoriesNewsletterMail` | `StoriesNewsletterService` | 1 newsletter → N subscribers |
| `AskMail` | `StoriesAskService` | 1 ask → N targets |
| `ChaseAdminMail` | `ChaseAdminCommand` | 1 chase → N mods of a group |
| `EventsDigestMail` | `EventsDigestService` | 1 group's events → N members |
| `VolunteeringDigestMail` | `VolunteeringDigestService` | 1 group's vols → N members (no-jobs path) |

Deliberately not migrated (no fan-out OR per-recipient content):
- `ModNotifMail`, `NewsfeedModNotifMail` — per-mod unique content
- `ChitchatReportMail`, `AlertNoMessagesMail` — single-recipient sends
- `AskForDonation`, daily `UnifiedDigest` — see spec for follow-up notes

Truly per-recipient mailables (welcome, verify-email, forgot-password,
chat-notification, birthday, donation thank-you, etc.) keep using the
existing per-send compile path. No change.

## Tests

- `BulkMjmlCompilerTest` (10): substitution, encoding (matches Blade's
  `e()`), unbound-placeholder validation, cache-hit accounting, `clear()`.
- `UnifiedDigestBulkTest` (7): shape stability, merge-var keys match
  placeholders, byte-identical HTML parity (bulk path output == non-bulk
  path output, normalising per-instance tracking IDs).
- `BulkRenderableMailablesParityTest` (7): byte-identical parity tests for
  each migrated mailable plus shape-bucketing assertions.
- All 65 existing `UnifiedDigest` tests still pass.

## Test plan

- [x] BulkMjmlCompilerTest — all 10 pass
- [x] UnifiedDigestBulkTest — all 7 pass including parity
- [x] BulkRenderableMailablesParityTest — all 7 pass including parity
- [x] Full UnifiedDigest test suite — all 65 pass
- [ ] Full Laravel test suite — running, no regressions expected
- [ ] (Ops, post-merge) Watch immediate-digest cron logs for
      `BulkMjmlCompiler: bulk compile stats` lines. Expected pattern:
      compile_count = 1 per message, hits >> 1 per shape.

## Rollback

Flip `FREEGLE_BULK_MAIL_ENABLED=false` in env and restart batch. All
mailables fall back to the per-recipient compile path immediately.

### DIFF
--- a/iznik-batch/app/Console/Commands/Mail/ChaseAdminCommand.php
+++ b/iznik-batch/app/Console/Commands/Mail/ChaseAdminCommand.php
@@ -202,7 +202,14 @@ class ChaseAdminCommand extends Command
                 );
 
                 if ($bulkCompiler !== null && $mailable instanceof BulkRenderable) {
-                    $mailable->setPrerenderedHtml($bulkCompiler->htmlFor($mailable));
+                    try {
+                        $mailable->setPrerenderedHtml($bulkCompiler->htmlFor($mailable));
+                    } catch (\Throwable $bulkE) {
+                        Log::warning('BulkMjmlCompiler failed; falling back to per-user MJML compile', [
+                            'mod_id' => $mod->id,
+                            'error' => $bulkE->getMessage(),
+                        ]);
+                    }
                 }
 
                 app(\App\Services\EmailSpoolerService::class)->spool($mailable);
--- a/iznik-batch/app/Console/Commands/Mail/ChaseAdminCommand.php
+++ b/iznik-batch/app/Console/Commands/Mail/ChaseAdminCommand.php
@@ -4,13 +4,14 @@ namespace App\Console\Commands\Mail;
 
 use App\Console\Concerns\PreventsOverlapping;
 use App\Mail\Admin\ChaseAdminMail;
+use App\Mail\Concerns\BulkRenderable;
 use App\Models\Group;
 use App\Models\User;
+use App\Services\BulkMail\BulkMjmlCompiler;
 use App\Traits\GracefulShutdown;
 use Illuminate\Console\Command;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Log;
-use Illuminate\Support\Facades\Mail;
 
 class ChaseAdminCommand extends Command
 {
@@ -168,6 +169,12 @@ class ChaseAdminCommand extends Command
 
         $sent = 0;
 
+        // One BulkMjmlCompiler per admin chase: all mods of the group share
+        // the same admin/group/pending data → one MJML compile, N
+        // substitutions for N mods.
+        $bulkEnabled = (bool) config('freegle.bulk_mail.enabled', true);
+        $bulkCompiler = $bulkEnabled ? app(BulkMjmlCompiler::class) : null;
+
         foreach ($moderators as $mod) {
             $email = $mod->email_preferred;
 
@@ -193,6 +200,10 @@ class ChaseAdminCommand extends Command
                     $pendingHours,
                     $admin->id
                 );
+
+                if ($bulkCompiler !== null && $mailable instanceof BulkRenderable) {
+                    $mailable->setPrerenderedHtml($bulkCompiler->htmlFor($mailable));
+                }
 
                 app(\App\Services\EmailSpoolerService::class)->spool($mailable);
 
--- a/iznik-batch/app/Mail/Admin/ChaseAdminMail.php
+++ b/iznik-batch/app/Mail/Admin/ChaseAdminMail.php
@@ -2,6 +2,7 @@
 
 namespace App\Mail\Admin;
 
+use App\Mail\Concerns\BulkRenderable;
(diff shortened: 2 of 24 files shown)


edwh:
## Adversarial review

**Verdict: safe to merge — no cross-recipient data bleed.** Traced a full render loop: `strtr()` returns a new string (never mutates the cached compiled HTML), each recipient gets its own `MjmlMailable` instance with its own `tracking`, AMP/text parts render per-recipient from real data (never cached), and the batch-token namespace makes user-controlled subjects/names unforgeable as merge placeholders. No CRITICAL issues. Two WARNINGs worth handling before/after merge:

### W1 — `StoriesNewsletterMail::bulkData()` passes a real `name` into the shared compile (latent bleed)
`'name' => $this->recipientName` is baked into the compile-once cache instead of being a placeholder. **Harmless today** (the newsletter template never renders `$name`) — but the moment someone adds a `Dear {{ $name }},` greeting, every recipient after the first gets the **first recipient's name**, and no current test would catch it. Cheap fix: `'name' => $this->ph('recipientName')` + add `recipientName` to `mergeVars()`, consistent with every other mailable here.

### W2 — Parity tests don't cover cross-recipient isolation
All HTML parity tests use the **same recipient** for both paths, so a same-shape bleed would pass them all. `BulkMjmlCompilerTest` checks isolation at the primitive level with a hand-crafted mailable, but no integration test wires a *real* mailable (EventsDigest, VolunteeringDigest, …) with **two different recipients** through one compiler. Suggest one parameterised test per mailable asserting `assertStringNotContainsString($emailA, $htmlB)`.

### INFO (no action required, noted for the reviewer)
- `UnifiedDigest::bulkData()` passes `'user' => $this->user` (first recipient's object) into the shared compile; inert only because `$userEmail ?? $user->email_preferred` short-circuits on the placeholder string. Fragile — better not to pass `user` at all.
- `substitute()` uses `ENT_HTML5`; Blade's `e()` does not — byte-identical for `htmlspecialchars`, so no effect.
- `EventsDigestMail::shapeKey()` json-encodes `$events` — ensure entries are plain arrays (no `Carbon` instances) so the shape key is stable across the batch.
- One `BulkMjmlCompiler` per *message*, shared across that message's recipients (UnifiedDigestService ~271) — correct; worth a comment so it isn't "helpfully" moved inside the recipient loop (which would kill the caching).
- No test for the fallback path (compiler throws / unbound placeholder → `prerenderedHtml` null → per-user build). Add one so a regression can't silently drop recipients.

The compile-once/substitute-many design is sound; W1 + the W2/fallback tests are the high-value follow-ups.

edwh:
Closing as **superseded by #717** (in-process mrml compiler).

This PR was complete and green — the compile-once / substitute-many MJML caching layer (BulkRenderable + shapeKey + placeholder substitution) worked, and CI was passing (build-and-test ✓, check-runner ✓, mark-ci-passed ✓) after the final `preparePostsForBulk` byline fix.

We've since decided to attack the root cost instead of caching around it: #717 swaps the MJML engine to **mrml** (Rust) compiled **in-process**, making each compile ~170× cheaper (≈3 ms vs ≈600 ms). That removes the per-compile HTTP round-trip for *every* email path — not just same-shape blasts — so the compile-once caching machinery here is no longer worth its complexity.

Nothing wrong with the work; it's just the wrong layer to optimise once compiles are cheap. Closing in favour of #717.

PR #577, opened 2026-05-30
fix(eee): force MV labels onto the same items the classifier saw

edwh:
## Summary

EEELabel micro-volunteering was serving any recent OFFER with a photo, regardless of whether the classifier had ever classified that item. Result: of the first 161 msgids volunteers labelled, **zero were in `eee_classifications`** — so the Condition/Weight/Size accuracy columns on the eee-browser dashboard stayed permanently empty.

This PR restricts the EEELabel candidate set to attachments the classifier has already processed.

## What changed

- **New MySQL pointer table `eee_classified_attachments`** — lists every `(messageid, attid)` the classifier has touched. The full classification data still lives in iznik-batch's SQLite; this table only mirrors the keys so the Go API can JOIN against it.
- **`EeeClassificationService`** now upserts the pointer from every image-bearing classification path (`classifyMessageWithDriver`, `storeImageResult`). Text-only classifications skipped (no attid).
- **`eee:backfill-classified-table`** artisan command copies the existing ~4,829 SQLite rows into MySQL once after deploy.
- **`getEEELabelChallenge`** Go query rewritten to JOIN the pointer table instead of scanning recent OFFER attachments. Drops the 30-day arrival window and the canonical-primary correlated subquery — both no longer needed because the pointer table already contains the attids the classifier picked. Orders by `classified_at DESC` so freshly classified items surface first.
- **Go test rewritten** — previously asserted multi-photo dedup; now asserts an OFFER whose attachment is NOT in `eee_classified_attachments` is never served, even when it would otherwise be the most recent eligible item.

## Deploy steps

1. Run the migration (creates an empty table — INSTANT, safe under writes).
2. Run `php artisan eee:backfill-classified-table` once (~5k rows in one batch, sub-second).
3. Deploy iznik-server-go. Until step 2 completes the API returns empty challenges — correct behaviour ("no candidate classified items yet").

Once both sides are deployed, every new model classification self-populates the pointer table, so volunteer labels and model output stay on the same item set going forward.

## Test Plan

- [ ] CircleCI passes the new Go test `TestGetMicrovolunteering_EEELabel_RestrictsToClassifiedItems`.
- [ ] After deploy + backfill, `curl /api/microvolunteering?types=EEELabel&jwt=...` returns an item where `msgid` exists in `eee_classified_attachments`.
- [ ] After one volunteer round, `https://freegle-eee-browser.fly.dev/api/review/stats` starts showing model accuracy for Condition/Weight/Size (previously stuck at 0).

### DIFF
--- a/iznik-batch/app/Console/Commands/Eee/EeeBackfillClassifiedTableCommand.php
+++ b/iznik-batch/app/Console/Commands/Eee/EeeBackfillClassifiedTableCommand.php
@@ -1,0 +1,75 @@
+<?php
+
+namespace App\Console\Commands\Eee;
+
+use App\Services\EeeSqliteService;
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\DB;
+
+/**
+ * One-shot backfill: copy every distinct (messageid, attid) the EEE
+ * classifier has ever processed from the iznik-batch SQLite into the new
+ * MySQL `eee_classified_attachments` pointer table.
+ *
+ * Once the new pointer table is populated, the Go micro-volunteering API
+ * joins against it so volunteers label items the model has classified —
+ * which is what the Condition/Weight/Size accuracy scoring relies on.
+ */
+class EeeBackfillClassifiedTableCommand extends Command
+{
+    protected $signature = 'eee:backfill-classified-table
+                            {--batch-size=1000 : Rows per INSERT batch}
+                            {--dry-run         : Show what would be inserted without writing}';
+
+    protected $description = 'Backfill eee_classified_attachments from the SQLite eee_classifications history';
+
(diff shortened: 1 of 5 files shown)


edwh:

edwh:
Reopened — not a duplicate. This PR introduces the eee_classified_attachments pointer table and constrains the EEELabel selector to items the classifier has already processed. It is a separate change from #496 (initial wire-up), the perf fix on master (dcdb5af2b), or the multi-photo dedup fix on master (40cdaa9a7) — those were all about either making EEELabel work at all or stopping the same item being labelled twice; this one fixes the model-vs-human-disjoint-set problem so the Condition / Weight / Size accuracy columns on the eee-browser dashboard can be populated.

PR #578, opened 2026-05-30
fix(modtools-notifications): re-hydrate ModTools groups on login so badges + group list appear together

edwh:
## Root Cause
ModTools client-side initialization race after auth state changes: the notification/newsfeedcount poller starts before the groups store and pending-messages list are hydrated, so badge counters tick over an empty UI; on sign-out the auth-watching stores clear group state but the sign-in handler does not eagerly re-fetch groups, leaving the user with an empty groups list until a manual refresh.

## Evidence
```
× INVERTED — fails on current broken code, passes after fix > after login, modGroupStore.list should be populated with the user's groups 6ms
  → expected [] to have a length of 1 but got +0

AssertionError: expected [] to have a length of 1 but got +0
 ❯ tests/unit/stores/modtools-notifications.spec.js:211:49
     expect(Object.keys(modGroupStore.list)).toHaveLength(1)
```

## Fix
`fetchUser()` in `iznik-nuxt3/stores/auth.js` now calls `modGroupStore.getModGroups()` (guarded by `miscStore.modtools`) after setting `this.groups` and `this.setUser(me)`. This ensures the ModTools group menu and badge counters are populated immediately on sign-in rather than waiting for the first route change or the 30-second `checkWork()` poller tick. The existing `useModGroupStore` import is added to `stores/auth.js`; modgroup.js already imported `useAuthStore` via the mock alias so circular-dependency resolution works cleanly.

## Alternatives Investigated
- Server-side fan-out (counts endpoint warm but message-list endpoint serving stale cache) — ruled out: would not explain the separately reported sign-out/sign-in group-loss symptom, which is purely client-side state.
- Stale local-storage snapshot survives sign-out and is shown on sign-in until reset — ruled out: symptom is empty not stale; a surviving cache would produce stale data not empty data.

## Other Call Sites
All login paths (`login()`, `loginGoogle()`, `loginFacebook()`, `saveAboutMe()`, `saveEmail()`, etc.) call the single shared `fetchUser()` method in `iznik-nuxt3/stores/auth.js`. No other login entry points need the same hydration call.

## Test
iznik-nuxt3/tests/unit/stores/modtools-notifications.spec.js
Command: cd /home/edward/FreegleDockerWSL/iznik-nuxt3 && npm run test:unit:run -- --reporter=verbose modtools-notifications
Regression suite: iznik-nuxt3 unit suite — SUITE_PASSED=yes (590 files, 12787 tests passed)

## Discourse
https://discourse.ilovefreegle.org/t/9737

### DIFF
--- a/iznik-nuxt3/stores/auth.js
+++ b/iznik-nuxt3/stores/auth.js
@@ -8,6 +8,7 @@ import {
 } from '~/api/BaseAPI'
 import { useComposeStore } from '~/stores/compose'
 import { useGroupStore } from '~/stores/group'
+import { useModGroupStore } from '~/modtools/stores/modgroup'
 import api from '~/api'
 import { useMobileStore } from '@/stores/mobile'
 import { useMiscStore } from '~/stores/misc'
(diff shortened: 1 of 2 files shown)


edwh:

PR #579, opened 2026-05-30
fix(modsupport): Unsubscribe button now actually removes member from group

edwh:
## Root Cause
unsubscribeConfirmed() in the ModSupport component called userStore.unsubscribe(props.id), which POSTs an 'Unsubscribe' user action but does NOT delete the membership row or write a log entry. To actually remove a member from a group, memberStore.remove(userid, groupid) (DELETE /memberships) must be called.

## Evidence
Failing test on commit 7bdad9f17 (the AssertFlip post-fix suite, asserting the correct call):
```
FAIL [FIX] unsubscribeConfirmed should call memberStore.remove with userid and groupid
AssertionError: expected "spy" to be called with arguments: [ 123, 789 ]
Number of calls: 0
  expect(mockMemberRemove).toHaveBeenCalledWith(USER_ID, GROUP_ID)
```
The companion AssertFlip Step 1 suite ([BUG] asserts the broken behaviour) PASSES on the same commit, proving the test discriminates fixed vs broken.

## Fix
Replace the userStore.unsubscribe call in unsubscribeConfirmed() with memberStore.remove(userid, groupid) for each of the user's memberships. The membership row is now deleted, the log entry is written by the backend DELETE /memberships handler, and the member no longer appears as active in the group.

## Alternatives Investigated
- Permission denial (silent 403): ruled out — Jos is an established mod and the reproduction shows the wrong endpoint, not a rejected one.
- Stub/no-op button handler: ruled out — the button does fire userStore.unsubscribe, just the wrong action.
- Backend endpoint omits both deletion and logging: ruled out — issue is purely frontend (wrong API called).

## Other Call Sites
grep for userStore.unsubscribe across iznik-nuxt3 production code: **zero results**. The only call site was in ModSupportUser.vue (now fixed). The userStore.unsubscribe method (POST /user action:'Unsubscribe') remains defined in the store but is no longer called from any production component.

## Test
iznik-nuxt3/tests/unit/components/modtools/ModSupportUnsubscribeBug.spec.js
Command: cd /home/edward/FreegleDockerWSL/iznik-nuxt3 && npm run test:unit:run -- --reporter=verbose tests/unit/components/modtools/ModSupportUnsubscribeBug.spec.js
Regression suite: iznik-nuxt3 unit tests — SUITE_PASSED=yes (590 test files, 12787 tests)

## Discourse
https://discourse.ilovefreegle.org/t/9738

### DIFF
--- a/iznik-nuxt3/modtools/components/ModSupportUser.vue
+++ b/iznik-nuxt3/modtools/components/ModSupportUser.vue
@@ -606,6 +606,7 @@
 <script setup>
 import { ref, computed, onMounted } from 'vue'
 import { useUserStore } from '~/stores/user'
+import { useMemberStore } from '~/stores/member'
 import ExternalLink from '~/components/ExternalLink'
 import api from '~/api'
 import { usePreferredEmail } from '~/modtools/composables/usePreferredEmail'
@@ -625,6 +626,7 @@ const props = defineProps({
 })
 
 const userStore = useUserStore()
+const memberStore = useMemberStore()
 
 const user = computed(() => userStore.byId(props.id))
 const expanded = ref(true)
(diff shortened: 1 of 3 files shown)


edwh:

PR #581, opened 2026-05-30
fix(location-map): do not fall back to group lat/lng when user has no location

edwh:
## Root Cause
In iznik-server-go user.GetLatLng (user/user.go:938-941), when a user has no settings.lat/lng and no lastlocation, the function falls back to returning the group's lat/lng. This causes members without a postcode to be plotted on the members map at the group's physical address.

## Evidence
```
location_map_9730_test.go:69: Not equal: expected: 0, actual: 51.5074
    Messages: user with no location must return lat=0, not group lat (51.5074)
location_map_9730_test.go:71: Not equal: expected: 0, actual: -0.1278
    Messages: user with no location must return lng=0, not group lng (-0.1278)
--- FAIL: TestGetLatLngNoLocationDoesNotFallBackToGroupCoords (0.32s)
```

## Fix
Removed the group-coordinate fallback (`else if ulgroups.Lastlat != 0 || ulgroups.Lastlng != 0`) from `GetLatLng`. The function now returns zero coords when a user has no `mylocation` in settings, no `lastlocation`, and no message-posted location. Callers already handle zero coords correctly (several have explicit `if latlng.Lat != 0` guards; the member map will show no pin for these users).

The parallel goroutine structure in the no-mylocation branch was also simplified: the group-query goroutine was removed (dead code after the fallback was dropped), and the message-location query is now done lazily only when the earlier checks yield nothing.

## Alternatives Investigated
- Frontend label substitution (group name shown instead of user location): ruled out by post 2 (Jeni) reporting the group's geographic position on the map, indicating server-side coord substitution.
- Users auto-joined with group lat/lng copied onto user record: ruled out — would persist permanently rather than appear as an on-the-fly fallback.

## Other Call Sites
```
user/user.go:340:               latlng = GetLatLng(id)         (sets user.Lat/Lng in GET /user response)
user/user.go:669:               latlng := GetLatLng(id)        (has guard: if latlng.Lat != 0)
user/user.go:1014:              latlng := GetLatLng(id)        (GetPublicLocation — passes to ClosestPostcode)
user/user.go:1179:              privatePos = GetLatLng(id)     (mod tools private position)
message/bounds.go:37:           latlng = user.GetLatLng(myid)  (post visibility bounds check)
isochrone/message.go:33:        latlng := user.GetLatLng(myid) (isochrone)
isochrone/message.go:137:       latlng := user.GetLatLng(myid) (isochrone)
newsfeed/newsfeed.go:137:       latlng := user.GetLatLng(uid)  (has guard: if latlng.Lat > 0)
newsfeed/newsfeed.go:316:       latlng := user.GetLatLng(myid) (newsfeed)
newsfeed/newsfeed.go:1158:      latlng := user.GetLatLng(myid) (newsfeed)
newsfeed/newsfeed.go:1322:      latlng := user.GetLatLng(myid) (newsfeed)
```
No other files contain a group-lat/lng fallback pattern (`ulgroups.Lastlat`/`ulgroups.Lastlng` was only in `GetLatLng`). All callers either have `if latlng.Lat != 0` guards or treat 0/0 as "no location" — the same as a user who genuinely has no location set.

## Test
iznik-server-go/test/location_map_9730_test.go
Command: `cd iznik-server-go && MYSQL_HOST=... MYSQL_PORT=3306 MYSQL_USER=root MYSQL_PASSWORD=iznik MYSQL_DBNAME=iznik go test ./test/ -run TestGetLatLngNoLocationDoesNotFallBackToGroupCoords -v -count=1`
Regression suite: go-api full suite — SUITE_PASSED=yes

## Discourse
https://discourse.ilovefreegle.org/t/9730

### DIFF
--- a/iznik-server-go/test/location_map_9730_test.go
+++ b/iznik-server-go/test/location_map_9730_test.go
@@ -1,0 +1,73 @@
+package test
+
+// Regression test for bug reported in topic 9730:
+// Members with no postcode/location are shown on the map at the group's physical
+// location instead of being shown as having no known location.
+//
+// Root cause: GetLatLng falls back to the most-recently-joined group's lat/lng
+// when a user has no personal location set (no mylocation in settings, no
+// lastlocation, no message location).
+//
+// Fix: GetLatLng should return zero/empty location for such users.
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/freegle/iznik-server-go/database"
+	user2 "github.com/freegle/iznik-server-go/user"
+	"github.com/freegle/iznik-server-go/utils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetLatLngNoLocationDoesNotFallBackToGroupCoords(t *testing.T) {
+	db := database.DBConn
+
+	prefix := uniquePrefix("latlng9730")
+
+	// Create a group with a distinct known lat/lng (clearly non-zero London coords).
+	groupLat := 51.5074
+	groupLng := -0.1278
+	groupName := fmt.Sprintf("TestGroup_%s", prefix)
+	db.Exec(fmt.Sprintf(
+		"INSERT INTO `groups` (nameshort, namefull, type, onhere, polyindex, lat, lng) "+
+			"VALUES (?, ?, 'Freegle', 1, ST_GeomFromText('POINT(-0.1278 51.5074)', %d), ?, ?)",
+		utils.SRID),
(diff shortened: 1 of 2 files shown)


edwh:
Closing — premise is wrong. The fallback to group lat/lng when a user has no postcode IS correct behaviour (so the user still appears somewhere meaningful on the members map; better than showing nothing). The actual reported issue (per #9730) is that some users are displaying the **group's name** as their location text — not the group's coordinates being used as a fallback. This is likely a map-area display issue. Needs proper debugging against the specific reported example via V2 API live connection before another fix is attempted.

PR #582, opened 2026-05-31
fix(chat-reviews): include messages from deleted senders in V2 review queue

edwh:
## Root Cause
The V2 (Go) chat-review list endpoint joins users with INNER JOIN ... WHERE deleted IS NULL, which silently drops chat_messages whose sender has been soft-deleted. V1 (PHP) does not apply this filter, so Old ModTools still shows those messages while new live ModTools does not — causing the moderator's review queue to appear stuck/empty even though messages exist.

## Evidence
```
=== RUN   TestReviewQueueIncludesMessagesFromDeletedSenders
    chat_test.go:4464: 
        	Error Trace:	.../chat_test.go:4464
        	Error:      	Should be true
        	Test:       	TestReviewQueueIncludesMessagesFromDeletedSenders
        	Messages:   	review queue must include message 6006 (chat 5805) even when sender 27258 is deleted; V2 drops it via INNER JOIN users AND deleted IS NULL but V1 does not
--- FAIL: TestReviewQueueIncludesMessagesFromDeletedSenders (0.66s)
FAIL
```

## Fix
Removed the `INNER JOIN users ON users.id = cm.userid AND users.deleted IS NULL` from two places:

1. **`chat/chatmessage.go` `getReviewQueue` — base query** (line 878): The join was filtering out messages whose sender was soft-deleted. No columns from `users` were selected, so the join served only as a filter. Removed entirely; group membership checks on the *recipient* still enforce mod scope.

2. **`chat/chatmessage.go` `getReviewQueue` — wider-review UNION** (line 917): Same pattern in the wider-review query. Removed.

3. **`session/session.go` `chatReviewSQL` — notification badge count** (line 1108): The sidebar badge count uses the same filter. The code comment explicitly says it "must match the logic in getReviewQueue()". Removed for consistency so the badge count matches what the queue shows.

4. **`session/session.go` — wider-review badge count** (line 1152): Same fix for the wider-review badge count subquery.

Access control is still enforced through group membership checks on the *recipient* (the person who received the suspicious message), not the sender. Removing the sender-deleted filter does not expose any data outside mod scope.

## Alternatives Investigated
- Frontend store dropping records: ruled out — reproduction test shows the backend itself drops them.
- Background job failure: ruled out — would affect both V1 and V2; reporter confirmed V1 still shows them.

## Other Call Sites
Grepped `iznik-server-go` for `INNER JOIN users ON users.id = cm.userid AND users.deleted IS NULL`. Found four occurrences — all in the two files above, all fixed. Other `INNER JOIN users` patterns in the codebase are for different contexts (active user display, session auth, story listing, users_related lookup) and intentionally filter deleted users. No other chat-review-adjacent files have the same bug pattern.

## Test
iznik-server-go/test/chat_test.go
Command: cd /home/edward/FreegleDockerWSL/iznik-server-go && MYSQL_HOST=172.19.0.3 MYSQL_PORT=3306 MYSQL_USER=root MYSQL_PASSWORD=iznik MYSQL_DBNAME=iznik_go_test JWT_SECRET=secret go test ./test/... -run TestReviewQueueIncludesMessagesFromDeletedSenders -v
Regression suite: go test ./... — SUITE_PASSED=yes (pre-existing unrelated failures: AMP_SECRET not set, BST timezone test, TestChangesWithSince — all confirmed to fail before this commit)

## Discourse
https://discourse.ilovefreegle.org/t/9656

### DIFF
--- a/iznik-server-go/changes/changes.go
+++ b/iznik-server-go/changes/changes.go
@@ -92,7 +92,7 @@ func GetChanges(c *fiber.Ctx) error {
 		since = time.Now().Add(-1 * time.Hour)
 	}
 
-	mysqlTime := since.Format("2006-01-02 15:04:05")
+	mysqlTime := since.UTC().Format("2006-01-02 15:04:05")
 
 	// Fetch message changes, user changes, and ratings in parallel.
 	var messages []MessageChange
--- a/iznik-server-go/chat/chatmessage.go
+++ b/iznik-server-go/chat/chatmessage.go
@@ -875,7 +875,6 @@ func getReviewQueue(c *fiber.Ctx, myid uint64) error {
 		"COALESCE((SELECT m2.groupid FROM memberships m2 WHERE m2.userid = cm.userid AND m2.groupid IN (" + groupIDList + ") LIMIT 1), 0) AS groupidfrom " +
 		"FROM chat_messages cm " +
 		"INNER JOIN chat_rooms cr ON cr.id = cm.chatid " +
-		"INNER JOIN users ON users.id = cm.userid AND users.deleted IS NULL " +
 		"LEFT JOIN chat_images ci ON ci.chatmsgid = cm.id " +
 		"LEFT JOIN chat_messages_held cmh ON cmh.msgid = cm.id " +
 		"LEFT JOIN chat_messages_byemail cme ON cme.chatmsgid = cm.id " +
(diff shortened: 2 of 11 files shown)


edwh:
Closing — premise is wrong. Excluding deleted users from the chat-review queue is the intended V2 behaviour (deleted senders typically = spammers/banned; surfacing their messages adds noise with no actionable outcome). The diagnosis was based on a grep diff with V1 rather than a verified specific example from #9656 post 36. The underlying bug (chat review notifications stuck/visible in Old ModTools but missing in Live ModTools) needs a real reproduction against an actual stuck message before another fix is attempted.

PR #585, opened 2026-05-31
fix(public-location): suppress display when user has no real location source

edwh:
## Root Cause

Discourse 9730 — mods reported that some members were showing the **group's name** as their location ("Freegle Hertford" appearing in the member's address slot on the modtools members map/list).

Verified against live data for the specific example (member #44932706, Medway Freegle):

- `users.lastlocation = NULL`, `users.settings.mylocation` absent → no real location source
- `GetLatLng` (correctly, for map-pin placement) falls back to the user's most-recent group centroid → returns `(51.405, 0.565)` which matches Medway Freegle's `(51.407659, 0.560547)`
- **`GetPublicLocationForUser`** (modtools view, `user/userInfo.go:295-314`) had its own explicit fallback that returned the group's `namefull` as the public `Display`
- **`GetPublicLocation`** (public endpoint, `user/user.go:1025`) ran `ClosestPostcode` + `ClosestSingleGroup` on the fallback coords and concatenated them: `"[area], [groupname]"`

Both paths produced misleading "[area], [groupname]" or just "[groupname]" strings that read like the member's address but were actually just the group's own location.

## Fix

The group-coord fallback in `GetLatLng` itself is **correct** — needed for map-pin placement and distance/visibility calculations. The bug was only in display-text generation.

1. New `hasRealLocation(id)` helper checks for `settings.mylocation`, `lastlocation`, or any message-attached location.
2. `GetPublicLocation` early-returns an empty `Publiclocation` when the user has no real source.
3. `GetPublicLocationForUser` (modtools) drops its "fall back to most recent group membership" branch entirely — returns `nil`. Frontend (`ModMember.vue`) already gates on `user.info.publiclocation` truthy, so the line is skipped cleanly.

## Alternatives Investigated

- **Remove the group-coord fallback in `GetLatLng`** — this is what closed PR #581 attempted. Per @edwh, that's the wrong layer: the fallback is correct for map placement, just not for display text.
- **Add an `IsFallback` flag to `GetLatLng`'s return** — would touch all 11 call sites of `GetLatLng`. Rejected as bigger change than needed; the display paths can detect "no real source" directly.

## Test

- `TestPublicLocation_UserWithNoRealLocation` — reproduces the public-endpoint bug; asserts empty `Display` for a user with no `mylocation` / `lastlocation` / message location.
- `TestPublicLocation_NoGroupNameFallback` (renamed from `TestPublicLocation_MostRecentMembership`) — flips the previous assertion. The original test codified the buggy "V1 parity" behaviour; that's exactly the behaviour mods complained about, so the test was wrong.

## Discourse
https://discourse.ilovefreegle.org/t/strange-location/9730

### DIFF
--- a/iznik-server-go/test/user_test.go
+++ b/iznik-server-go/test/user_test.go
@@ -3215,44 +3215,87 @@ func TestPublicLocation_ResponseStructure(t *testing.T) {
 	assert.True(t, hasGroupname, "Response should have 'groupname' field")
 }
 
-func TestPublicLocation_MostRecentMembership(t *testing.T) {
-	// V1 parity: public location uses the most recently added membership,
-	// regardless of role. Moderators commonly moderate their own local group.
-	prefix := uniquePrefix("publocrecent")
+// TestPublicLocation_UserWithNoRealLocation reproduces Discourse 9730:
+// a member who has joined a group but never set a postcode was being
+// displayed (in modtools and the public location endpoint) with their
+// group's centroid coords resolved to "[area], [groupname]" — looking like
+// the member's home address. The fix returns empty Display when the user has
+// no real location source (settings.mylocation / lastlocation / message
+// location), regardless of group membership. Group-coord fallback in
+// GetLatLng is still correct for map-pin placement.
+func TestPublicLocation_UserWithNoRealLocation(t *testing.T) {
+	prefix := uniquePrefix("publocnolocation")
 	db := database.DBConn
 
-	// Create a moderator user who will view the target user's profile.
+	// Create a user with NO mylocation in settings and NO lastlocation —
+	// the CreateTestUser helper always sets mylocation, so build directly.
+	fullname := "Locationless " + prefix
+	db.Exec("INSERT INTO users (firstname, lastname, fullname, systemrole, lastlocation, settings) "+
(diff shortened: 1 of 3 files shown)


edwh:
Closing — diverges from V1 (`iznik-server/include/user/User.php:4906`) which intentionally falls back to the most-recent group's name when a user has no real location. That branch covers one symptom path but is the *V1-parity* behaviour, not a V2 regression.

The actual root cause of #9730 'group name as location' is in the data, confirmed via live DB:

- 21 rows in `locations` have a Freegle group name in the `name` column (e.g. id 9506662 'Eastleigh Freegle', id 9352865 'Eastleigh-Freegle', several '...freegle CGA'). These are imported group polygons that landed in the place-name table.
- Real users in those areas (e.g. user 2937991, postcode SO17 1TG) get `settings.mylocation.area.id = 9506662` and `area.name = 'Eastleigh Freegle'` copied verbatim into their settings JSON at signup.
- When the modtools view renders `publiclocation.area.name`, it shows 'Eastleigh Freegle' as if it were the member's town.

Next step is to (a) rename or relink the polluted `locations` rows, (b) repair affected `settings.mylocation.area.name` values for existing users, (c) locate the import/write path that allows group names into `locations.name` and prevent recurrence.

PR #591, opened 2026-05-31
fix(message): log message deletion once for cross-posted messages

edwh:
## Root Cause
`api/message.php` (~line 424) calls `$m->delete($reason, NULL, ...)` ignoring the request groupid; `Message::delete()` then loops over every group and inserts one `SUBTYPE_DELETED` log row per group, so a 2-group cross-posted message is logged twice. Introduced by PR #590.

## Evidence
Failing test before fix: `assertEquals(1, 2)` — count==2 for a 2-group cross-posted message.

CONFIRMED_FAILING=yes — test output: `Failed asserting that 2 matches expected 1` at `messageAPITest.php:4063`

## Fix
Changed `api/message.php` line 424 from `$m->delete($reason, NULL, $subject, $body, $stdmsgid)` to `$m->delete($reason, $groupid, $subject, $body, $stdmsgid)`. Passing the request groupid scopes `Message::delete()`'s SQL query to a single `messages_groups` row, so the per-group loop runs once and inserts exactly one `SUBTYPE_DELETED` log entry. The single mod-mail `chat_messages` entry is preserved (the loop body still calls `maybeMail()` for the target group).

## Alternatives Investigated
- Dual-write V2 Go / V1 PHP — ruled out: both rows originate from the same PHP delete path.
- DB trigger / duplicate listener — ruled out: no trigger/listener on logs/messages tables; would not correlate with post-#590 regression.

## Other Call Sites
Three other `->delete()` calls pass NULL groupid intentionally:
- `api/message.php:248` — PUT/draft-delete path, deletes from all groups with no group context
- `api/message.php:923` — "Withdrawn pending", deletes from all groups
- `include/spam/Spam.php:787` — spam deletion, removes from all groups

No other occurrences of the ModTools-style group-scoped delete with NULL groupid.

## Test
`iznik-server/test/ut/php/api/messageAPITest.php::testDeleteCrossPostedMessageDoesNotDoubleLog`
Command: `php .../phpunit --filter messageAPITest::testDeleteCrossPostedMessageDoesNotDoubleLog ...`
Proves: FAILS before (count==2), PASSES after (count==1, mod-mail preserved).
Regression suite: `messageAPITest.php` — SUITE_PASSED=yes (70 tests, 2724 assertions)

## Discourse
https://discourse.ilovefreegle.org/t/9685

### DIFF
--- a/iznik-server/http/api/message.php
+++ b/iznik-server/http/api/message.php
@@ -421,7 +421,7 @@ function message() {
                     switch ($action) {
                         case 'Delete':
                             # The delete call will handle any rejection on Yahoo if required.
-                            $m->delete($reason, NULL, $subject, $body, $stdmsgid);
+                            $m->delete($reason, $groupid, $subject, $body, $stdmsgid);
                             break;
                         case 'Reject':
                             # Ignore requests for messages which aren't pending.  Legitimate timing window when there
--- a/iznik-server/test/ut/php/api/messageAPITest.php
+++ b/iznik-server/test/ut/php/api/messageAPITest.php
@@ -3986,4 +3986,81 @@ class messageAPITest extends IznikAPITestCase
 
         $this->assertEquals(0, count($ret['message']['outcomes']));
     }
+
+    /**
+     * AssertFlip test: deleting a cross-posted message via ModTools PHP API must write
+     * exactly ONE deletion log entry, not one per group.
+     *
+     * Root cause: message.php line 424 calls $m->delete($reason, NULL, ...) — the groupid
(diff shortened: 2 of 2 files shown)


edwh:
Closing as a bad fix. This patches `api/message.php` (`Message::delete()` log loop) — but that V1 PHP code is now obsolete and is not what runs in production, so it cannot be the cause of the double deletion-logging reported in 9685.7. Re-diagnosis should target the V2 (Go API) message-deletion path, not V1.

Bug kept open for re-diagnosis.

PR #599, opened 2026-06-01
fix(member-management): Unsubscribe button in Support tools not removing member or logging

edwh:
## Root Cause

`PostUser` (Go API `user/user.go`) handled actions via a switch statement but had no `\"Unsubscribe\"` case. It fell through to `default: return fiber.NewError(StatusBadRequest, \"Unknown action\")`. Every click of the Support tools Unsubscribe button sent `POST /user {action: \"Unsubscribe\", id: <userid>}`, received a 400 error, and silently did nothing — no membership removal, no log entry, no limbo state set.

## Evidence

- `UserAPI.js:109` calls `POST /user` with `action: 'Unsubscribe'`
- `PostUser` switch (user/user.go) has `Rate`, `RatingReviewed`, `AddEmail`, `RemoveEmail`, `Unbounce`, `Merge` — no `\"Unsubscribe\"` case
- V1 parity: PHP `user.php:327-329` handles action=Unsubscribe by calling `\$u->limbo()` (sets `users.deleted = NOW()`)
- Test confirms: before fix, POST returns 400, membership survives, user.deleted NULL, zero log entries

## Fix

Added `\"Unsubscribe\"` case to `PostUser` dispatching to `handleUserUnsubscribe`:
1. Validates caller is admin/support or the user themselves
2. `DELETE FROM memberships WHERE userid = ? AND collection = 'Approved'` — removes group memberships
3. `UPDATE users SET deleted = NOW()` — marks account as limbo (14-day soft-delete)
4. `INSERT INTO logs (type=User, subtype=Deleted)` — creates audit log entry

## Other Call Sites

- `SessionAPI.js:75` calls `POST /session` with `action: Unsubscribe` (email-based flow) — different endpoint, unaffected
- `UserAPI.js:109` — the only caller of the now-fixed `POST /user action=Unsubscribe`

## Test

`TestPostUserUnsubscribeBySupportRemovesMembership`: Support user calls Unsubscribe on a member; asserts membership removed, account in limbo, log entry written. Failed before fix (400 response), passes after.

`TestPostUserUnsubscribeNonSupportForbidden`: Regular user cannot unsubscribe another. Returns 403.

## Discourse
https://discourse.ilovefreegle.org/t/9738/1

### DIFF
--- a/iznik-server-go/test/user_test.go
+++ b/iznik-server-go/test/user_test.go
@@ -1814,6 +1814,79 @@ func TestPostUserUnbounceNotAdmin(t *testing.T) {
 	assert.Equal(t, fiber.StatusForbidden, resp.StatusCode)
 }
 
+// TestPostUserUnsubscribeBySupportRemovesMembership verifies that a Support user clicking
+// "Unsubscribe" in the Support tools removes the target user's memberships and marks their
+// account as deleted (limbo). Regression: POST /user action=Unsubscribe was missing from
+// PostUser's switch, returning "Unknown action" so nothing happened (Discourse #9738 post 1).
+func TestPostUserUnsubscribeBySupportRemovesMembership(t *testing.T) {
+	prefix := uniquePrefix("unsub_support")
+	db := database.DBConn
+
+	supportID := CreateTestUser(t, prefix+"_support", "Support")
+	targetID := CreateTestUser(t, prefix+"_target", "User")
+	groupID := CreateTestGroup(t, prefix)
+	CreateTestMembership(t, targetID, groupID, "Member")
+	_, supportToken := CreateTestSession(t, supportID)
+
+	// Verify member exists before unsubscribe.
+	var memberBefore int64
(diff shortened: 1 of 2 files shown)


edwh:
Refactored per review (limbo + shared code):

- **Shared code**: extracted the soft-limbo core (remove approved memberships → mark `users.deleted` → log `User/Deleted`) into `softLimboUser()`. Both `LimboUser`'s self-delete branch and the Support-tools `Unsubscribe` action now call it, so there's a single definition instead of duplicated raw SQL.
- **Semantics**: `POST /user action=Unsubscribe` is a **recoverable limbo** for any target (admin/support may unsubscribe others) — matching V1's `User::limbo()`. This is deliberately distinct from `DELETE /user`, which hard-purges (GDPR forget) when support acts on another user. A user who unsubscribed by mistake (e.g. one-click email unsubscribe) can recover by logging back in within the grace period.
- Corrected the earlier commit's inaccurate "calls `User::limbo()`" note.

Note: V1 `limbo()` also emails the user ("account removed"); the V2 path (both `LimboUser` and this) does not send that mail. Left out of scope here to match existing V2 behaviour — flag if you want it added.

PR #602, opened 2026-06-01
fix(chat): do not hold chat messages for review for containing a postcode

edwh:
## Root Cause

`ContentCheckService::checkChatMessage()` passes the raw message text directly to `checkConcernKeywords()`. When the `concern_keywords` table contains a regex entry whose pattern coincidentally matches UK postcode shapes (e.g. a scam-detection rule covering `\b[A-Z]{1,2}\d\s+\d[A-Z]{2}\b`), any chat message sharing a collection address is incorrectly held for moderator review.

## Evidence

Discourse report: https://discourse.ilovefreegle.org/t/9656/38 (post 38, reporter Neville_Reid).

A message like `"Please collect at AB1 2CD"` was flagged as `reviewrequired=1` because the postcode matched a globally-scoped regex concern keyword. The phone-number exemption (commit 2804d6e, "stop phone-number checking on chat messages") addressed the same class of problem for phone numbers but left postcodes unprotected.

## Fix

Strip UK postcodes from the chat message text **before** running any text checks in `checkChatMessage()`. Uses the standard Royal Mail postcode regex (matching V1 `Utils::POSTCODE_PATTERN`) as a `private const POSTCODE_PATTERN`. All other spam content in the same message is still caught — the sanitised string is passed through the full check chain.

V1 parity: `Spam::checkReview()` never checked postcodes either.

## Other Call Sites

`checkChatMessage()` is only called in one place:
- `ChatProcessService::processMessage()` (line 136) — the only consumer of this check

`POSTCODE_PATTERN` is `private const`, used only within `ContentCheckService::checkChatMessage()`.

## Test

`ChatProcessServiceTest::test_moderated_user_message_with_postcode_is_not_held()`:
- Seeds a global regex concern keyword matching simplified UK postcodes (`\b[A-Z]{1,2}\d\s+\d[A-Z]{2}\b`)
- Sends a message `"Please collect at AB1 2CD"` from a Moderated user
- Asserts `reviewrequired = 0` (not held) and `reportreason = null`
- Without the fix, `reviewrequired` was `1` (AssertFlip confirmed)

## Discourse

https://discourse.ilovefreegle.org/t/9656/38

### DIFF
--- a/iznik-batch/app/Services/ContentCheckService.php
+++ b/iznik-batch/app/Services/ContentCheckService.php
@@ -160,6 +160,12 @@ class ContentCheckService
      * checked phone numbers either.) The checkPhoneNumbers() check still runs
      * for posts via checkMessage().
      *
+     * UK postcodes are stripped from the message before concern-keyword and
+     * other text checks run: sharing a collection address is as legitimate as
+     * sharing a phone number, and a postcode-shaped string must not trigger a
+     * regex concern keyword added for other purposes. (V1 parity: Spam::checkReview()
+     * never checked postcodes either.)
+     *
      * @return array|null Reason ['check','category','action','detail'] or null.
      */
     public function checkChatMessage(string $message): ?array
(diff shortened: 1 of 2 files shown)


edwh:
## Closing — root cause is incorrect

I checked this against the **live production database** (via the V2 DB tunnel) and the postcode-concern-keyword theory does not hold up:

**Production `concern_keywords` contains no pattern that matches a UK postcode.**
- Exactly **1** `regex` keyword exists globally: `(?<!\bwater\W)\bbutt\b(?!\s+rd)` (the "water butt" review rule). It matches no postcode.
- 1,033 `literal` and 221 `fuzzy` keywords — none match a postcode either (literal matching is `\bword\b` exact-word; tested all three modes against many postcode samples e.g. `AB1 2CD`, `SW1A 1AA`, `M1 1AE`).

**The report was speculation, not a confirmed cause.** Discourse [9656/38](https://discourse.ilovefreegle.org/t/9656/38) (Neville_Reid) says only: *"this one **seemed** to be held for showing a postcode."*

**What was actually held:** every review-held message containing a postcode in that window had `reportreason = 'Spam'` — marketing blasts ("Half-price Italian…", "Your Weekend Deals Are Inside") that happen to carry a postcode in their footer. They were held correctly **as spam**; the postcode was just the visible thing the moderator noticed.

So stripping postcodes before the checks fixes a cause that doesn't exist (no concern keyword can fire on a postcode), and removes a (harmless) defensive behaviour for a misdiagnosed report. Closing.

The genuine hold reason is being investigated separately via the DB + Loki tunnels; if a real bug surfaces there, it'll get its own PR with the correct root cause.

PR #606, opened 2026-06-01
fix(modtools): reject cross-posted messages on all authorized pending groups

edwh:
## Root Cause

`handleReject()` called `resolveAuthorizedGroups(myid, reqGid, ctx.Groupids)` with the specific `groupid` supplied by the frontend. This returns exactly one group when a `groupid` is specified, so the `UPDATE … WHERE groupid IN (?) AND collection = 'Pending'` only touched that one row.

For a message cross-posted to multiple groups (both Pending), rejecting via `groupid=A` left group B in `Pending`. On the next frontend fetch, `message.groups[0]` could resolve to the already-Rejected group A (no `ORDER BY` in the DB query), and the retry UPDATE would silently match 0 rows — leaving the message permanently stuck in the mod's pending queue.

Additionally, `handleReject()` never cleared `messages_groups.heldby` / `messages.heldby`, unlike `handleApprove()`. Held-and-then-rejected messages left dangling `heldby` references that inflated the "held pending" dashboard count.

## Evidence

**Failing test (before fix)**:
```
--- FAIL: TestRejectCrossPostedClearsAllPendingGroups
    message_test.go:1175: group B must also be Rejected
    Expected "Rejected", got "Pending"
```

**After fix**: both tests pass in the full suite (2991✓ 0✗).

## Fix

`iznik-server-go/message/message.go` — `handleReject()`:

1. **Validate `reqGid` explicitly** — same 403/404 error semantics as before when an invalid groupid is specified.
2. **Call `resolveAuthorizedGroups(myid, 0, ctx.Groupids)`** — returns all groups the caller moderates, not just the one passed in. The UPDATE still has `AND collection = 'Pending'` so Approved/Spam rows are unaffected.
3. **Clear `heldby` after rejection** — mirrors the cleanup `handleApprove` already does; prevents stale held-pending dashboard counts.

## Other Call Sites

All other handlers (`handleApprove`, `handleDeleteMessage`, `handleSpam`, `handleHold`, `handleRelease`, `handleBackToPending`) use `resolveAuthorizedGroups(myid, reqGid, ...)` unchanged — per-group semantics are correct for those actions.

## Test

| File | Command | Proves |
|---|---|---|
| `iznik-server-go/test/message_test.go` | `TestRejectCrossPostedClearsAllPendingGroups` | cross-post rejection clears both groups — FAIL before fix, PASS after |
| `iznik-server-go/test/message_test.go` | `TestRejectHeldMessageClearsHeldBy` | held rejection clears heldby — FAIL before fix, PASS after |

## Discourse
https://discourse.ilovefreegle.org/t/9729/1

### DIFF
--- a/iznik-server-go/message/message.go
+++ b/iznik-server-go/message/message.go
@@ -1832,7 +1832,25 @@ func handleReject(c *fiber.Ctx, myid uint64, req PostMessageRequest) error {
 	if req.Groupid != nil {
 		reqGid = *req.Groupid
 	}
-	authorizedGroups, err := resolveAuthorizedGroups(myid, reqGid, ctx.Groupids)
+	// Validate the specified groupid (if any), but reject from ALL groups the caller
+	// is authorised to moderate.  A cross-posted message may be pending on several
+	// groups; limiting the UPDATE to one groupid leaves the others stuck in Pending.
+	if reqGid > 0 {
+		if !auth.IsModOfGroup(myid, reqGid) {
+			return fiber.NewError(fiber.StatusForbidden, "Not a moderator for this group")
+		}
+		onGroup := false
+		for _, gid := range ctx.Groupids {
+			if gid == reqGid {
+				onGroup = true
+				break
+			}
+		}
+		if !onGroup {
+			return fiber.NewError(fiber.StatusNotFound, "Message not on that group")
+		}
+	}
+	authorizedGroups, err := resolveAuthorizedGroups(myid, 0, ctx.Groupids)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


edwh:
## Closing — misdiagnosed (wrong model of the bug)

Verified against the **live production database**: this PR's premise does not hold.

- **0** single-msgid messages cross-posted to multiple groups with a mixed Rejected+Pending state in the last month — i.e. the scenario this PR fixes (reject left one *group* of the same message Pending) **does not occur** in production.
- The real "rejected but stuck in pending" pattern is **duplicate messages**: a member posts the same item several times, creating **separate `messages.id` rows**. Two live examples:
  - `WANTED: pipe benders` — `120486658:Rejected`, but `120486640 / 120501385 / 120504832 : Pending`
  - `WANTED: Toys` — `120490522:Rejected`, but `120490543 / 120490546 : Pending`

This matches the report ([9729](https://discourse.ilovefreegle.org/t/9729/1)): Michael rejected the **duplicate**, the member acknowledged it, yet it's still in his pending queue — because the *other duplicate msgid(s)* are still Pending.

`handleReject()` operates on the groups of **one** `msgid`. It cannot clear the **sibling duplicate msgids**, so it would not fix 9729. The correct fix is at the duplicate level — e.g. when rejecting a message detected as a duplicate, also reject/collapse the sibling duplicate posts (same poster + subject still Pending), and/or dedup the pending queue. The `heldby` cleanup here is a reasonable separate tidy-up but isn't the fix for this report.

Closing; the real duplicate-post handling will be scoped separately.

PR #616, opened 2026-06-02
fix(chat): strip UK postcodes before concern-keyword check to prevent false review holds

edwh:
## Root Cause

`checkChatMessage()` passes the raw message (including postcodes) to `checkConcernKeywords()`, which applies all global concern keywords — including regex-mode ones. The `concern_keywords` table can contain global regex patterns designed to catch scam patterns (e.g. `\b[A-Z]{1,2}\d\s+\d[A-Z]{2}\b`) whose syntax accidentally matches UK postcode inward/outward codes like "AB1 2CD". When a Moderated member shares a collection address, the postcode portion fires the regex concern keyword and the message is held for moderator review.

**Why the prior attempt (origin/fix/chat-postcode-review-9656-38) was rejected:** That PR had the correct postcode-stripping approach but also reverted the `reportReasonForCheck` mapping introduced by commit `478171c80` (feat: surface specific hold reason in chat review). This caused a regression — the modtools review UI would no longer show the specific reason (e.g. "It looks like it refers to money") and instead fall back to the generic unhelpful "failed spam checks". This fix keeps `reportReasonForCheck` and `CHECK_TO_REPORTREASON` intact.

**How this differs:** Only strips postcodes — all other content checks run on the unmodified message. The `reportReasonForCheck` and `CHECK_TO_REPORTREASON` mapping from `478171c80` are preserved.

## Evidence

The test `test_moderated_user_message_with_postcode_is_not_held` inserts a global regex concern keyword `\b[A-Z]{1,2}\d\s+\d[A-Z]{2}\b` that matches UK postcodes, then sends "Please collect at AB1 2CD". Without the fix, `checkConcernKeywords` returns a match and `reviewrequired=1`. With the fix, the postcode is stripped before the check, `checkConcernKeywords` sees only whitespace, returns null, and the message passes through.

Parallel to the phone-number exclusion (commit `2804d6e`): sharing a collection address (postcode) in chat is as normal as sharing a phone number. V1's `Spam::checkReview()` never had postcode-matching entries in `spam_keywords`.

## Fix

Added `POSTCODE_PATTERN` private const to `ContentCheckService` (mirrors V1 `Utils::POSTCODE_PATTERN`). In `checkChatMessage()`, strips postcodes from the message with `preg_replace` before running all content checks. The sanitised string replaces postcodes with a space, leaving all other content (including genuine spam signals) intact.

## Other Call Sites

`checkChatMessage()` is only called from `ChatProcessService::processMessage()` (one call site).

## Test

**File:** `iznik-batch/tests/Unit/Services/ChatProcessServiceTest.php`  
**Class/Filter:** `ChatProcessServiceTest::test_moderated_user_message_with_postcode_is_not_held`  
**Proves:** fail-before (postcode IS held with regex concern keyword in DB) → pass-after (postcode stripped, message passes through).

Existing tests for concern-keyword holds, money-symbol holds, and the specific `reportReasonForCheck` mapping are all preserved and pass.

## Discourse

https://discourse.ilovefreegle.org/t/9656/38

### DIFF
--- a/iznik-batch/app/Services/ContentCheckService.php
+++ b/iznik-batch/app/Services/ContentCheckService.php
@@ -143,6 +143,11 @@ class ContentCheckService
         return $reasons;
     }
 
+    // UK postcode pattern — mirrors Utils::POSTCODE_PATTERN from V1.
+    // Covers GIR 0AA plus all standard AN/ANN/AAN/AANN/ANA/AANA inward forms.
+    // Used to sanitise chat messages before concern-keyword checks run.
+    private const POSTCODE_PATTERN = '/([Gg][Ii][Rr] 0[Aa]{2})|((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?))))\s?[0-9][A-Za-z]{2})/m';
+
     /**
      * Run the text-based content checks relevant to a chat message and return
      * the first failure reason (or null when clean).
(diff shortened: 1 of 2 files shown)


edwh:
Closing — this re-fixes bug 9656/38, which is working-as-designed and was already closed once (PR #602). Verified previously via the live DB: chat postcodes do not match any concern/spam keyword, so the hold the reporter saw was not caused by postcode text. There is no code defect to fix here. The FSM re-opened this after #602 was closed; marking the bug terminal so it isn't re-attempted.

PR #622, opened 2026-06-03
fix(chat-review): stop false "not English" flags via restricted language set + 80-char gate (#9481)

edwh:
## Problem

The chat/post content check false-flags **plain English** as "It might not be in English" (Discourse #9481). Confirmed in production: the exact reply *"Yes please can collect.Paul / Possible collection times: Asap"* (chat_messages 108721900) was shown in ModTools chat review with that reason, and ~5 clearly-English messages/day are flagged `reportreason='Language'` (e.g. "Still available?", "Thank you for letting me know kind regards Danielle", "Yes please lovely vases!").

## Root cause (verified against the real detector)

`ContentCheckService::checkLanguage` runs `patrickschur/language-detection` over the **full ~110-language set** and accepts English only if it's top-ranked or `P(en) >= 0.9 × P(top)`. On short, app-templated replies the library ranks **constructed/Latinate languages** top — Interlingua (`ia`), Occitan (`oc`), Esperanto (`eo`) — with English at ~0.86–0.89 of the top, so English misses the 0.9 bar and gets flagged.

## Why not just lower the threshold to 0.8

The 0.9 was **deliberate** (commit `38ac487`: *"0.8 let French through; 0.9 flags it correctly"*). Verified the cost of reverting: at 0.8 the detector **misses real Spanish** (en/top 0.80) and re-opens the French hole. The en-ratio bands overlap (terse English 0.86–0.90 vs simple French ~0.84), so no single threshold separates them. A straight 0.8 revert trades English false-positives for foreign false-negatives.

## Fix (verified on real production strings)

1. **Restrict the detector to a realistic UK + common-spam language set** (drop the conlangs; Scandinavian kept) → English ranks top on the false positives, while French/Dutch/Spanish are still detected.
2. **Raise the length gate 50 → 80** → sub-80-char replies are unreliable to detect and low-risk, so the remaining short-text false positives are skipped entirely.
3. **Keep the 0.9 threshold** → no regression of foreign detection.

Empirical check (real detector):
- Long English → `en` top, **not flagged**; the production false positives → skipped (<80) or English-top.
- French / Dutch / Spanish (>80 chars) → **still flagged** at 0.9.

## Tests

`ContentCheckServiceTest`:
- `test_check_language_text_below_80_chars_skipped` — sub-80-char text skipped before detection even with an always-flag detector.
- `test_check_language_real_world_terse_english_reply_not_flagged` — the actual production false-positive string (real detector) is not flagged.
- `test_check_language_long_english_not_flagged_with_restricted_set` — long English not flagged (restricted set, real detector).
- `test_clearly_non_english_still_flagged_with_v1_threshold` — genuinely non-English still flagged at 0.9.

## Discourse

https://discourse.ilovefreegle.org/t/9481

### DIFF
--- a/.circleci/continue-config.yml
+++ b/.circleci/continue-config.yml
@@ -23,7 +23,7 @@ parameters:
     description: "Forwarded from trigger — true when build_modtools_production was set"
 
 orbs:
-  freegle: freegle/tests@1.1.337
+  freegle: freegle/tests@1.1.339
 
 jobs:
   mark-ci-passed:
--- a/.circleci/continue-config.yml
+++ b/.circleci/continue-config.yml
@@ -23,7 +23,7 @@ parameters:
     description: "Forwarded from trigger — true when build_modtools_production was set"
 
 orbs:
-  freegle: freegle/tests@1.1.334
+  freegle: freegle/tests@1.1.336
 
 jobs:
   mark-ci-passed:
--- a/.circleci/continue-config.yml
+++ b/.circleci/continue-config.yml
@@ -23,7 +23,7 @@ parameters:
     description: "Forwarded from trigger — true when build_modtools_production was set"
 
 orbs:
-  freegle: freegle/tests@1.1.336
+  freegle: freegle/tests@1.1.337
 
 jobs:
   mark-ci-passed:
--- a/iznik-batch/app/Services/ContentCheckService.php
+++ b/iznik-batch/app/Services/ContentCheckService.php
@@ -977,10 +977,12 @@ class ContentCheckService
     // Language detection — flag non-English/Welsh messages over 50 chars
     // (V1 Spam.php parity using patrickschur/language-detection).
     // English is accepted if it is the top language, or if P(en|cy) >= 0.8 *
-    // P(top language) — the same lax threshold V1 uses.
+    // P(top language) — the same lax threshold V1 uses (Spam.php line 413).
+    // The optional $detector callable accepts the lowercased text and returns
+    // an array of language => probability; used in tests for deterministic results.
     // -------------------------------------------------------------------------
 
-    public function checkLanguage(string $subject, string $textbody): ?array
+    public function checkLanguage(string $subject, string $textbody, ?callable $detector = null): ?array
     {
         $text = trim(str_ireplace('xxx', '', strtolower($textbody)));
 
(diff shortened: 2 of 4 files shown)


edwh:
Closing — this fix is not grounded in the reported problem.

**Why:** Topic 9656 is about chat-review **false positives**, but the actual reports are phone numbers (post 8), the word "grass" (post 13) and postcodes (post 38) — all `WorryWord`/content holds. **None mention language detection.** This PR was derived from code inspection (the 0.8→0.9 threshold change in 38ac487), not a user report — hence the bare `t/9656` link with no specific post.

**Production data (last 14–30 days) confirms it's a non-issue:**
- `chat_messages` held for review by reason (14d): Spam 9, WorryWord 5, Last 2, Money 1 — **Language: 0 held**.
- `reportreason='Language'` total over 30 days: 16, none held for review.

So the language threshold is not flagging anything in practice, and changing 0.9→0.8 does not address any of the 9656 reports.

The 0.8-vs-0.9 discrepancy vs V1/`SpamCheckService` may still be worth aligning, but that's a separate, separately-justified change — not a fix for this thread. Closing to avoid shipping a fix attributed to a problem it doesn't solve.

edwh:
Reopening — I was wrong to close this. It is **not** hallucinated; the language false-positive is real and reaching the mod review queue.

**Evidence (live production):**
- A clearly-English chat reply — *"Yes please can collect.Paul … Possible collection times: Asap"* (chat_messages id 108721900) — was flagged `reportreason='Language'` and shown in ModTools chat review as *"It might not be in English, so needs checking via Google Translate."*
- Over the last 3 days, **16** clearly-English messages were flagged `Language` (e.g. "Still available?", "Thank you for letting me know kind regards Danielle", "Yes please lovely vases!"), ~5/day reaching review.

**Why I closed it wrongly:** I measured `reviewrequired=1` at a single instant (0, because they'd already been mod-approved back to 0) instead of the `reportreason='Language'` flag rate. The flag fires regularly on obvious English.

**Why the bug is real:** `ContentCheckService::checkLanguage` accepts English only if it's top-ranked or `P(en) >= threshold × P(topLang)`. The code uses **0.9** (line 1006) but its own comment says **0.8** ("the same lax threshold V1 uses"), and V1 `Spam.php` + `SpamCheckService` both use 0.8. On terse English the detector ranks English just below the 0.9 bar → false "Language" flag. This PR reverts 0.9→0.8 (V1 parity), which fixes these cases.

PR #631, opened 2026-06-03
fix(ai-image): exclude suppressed/non-active images from /illustration endpoint

edwh:
## Root Cause

`GetIllustration` (`iznik-server-go/misc/illustration.go:66`) queried `ai_images` with no `status` filter:

```sql
SELECT externaluid FROM ai_images WHERE name = ?
```

This meant suppressed (and rejected/regenerating) AI images were still returned during compose. The user saw the illustration, attached it, submitted their post — but then `GET /api/message/:id`'s LEFT JOIN blanked the `externaluid` for non-active `ai_images` rows. The result was a ghost photo slot that looked like spam suppression (topic 9753).

The `feat(ai-image): suppress AI images for unsuitable items (topic 9630)` commit updated the message display query and the PHP batch generator to respect `status = 'suppressed'`, but missed updating the compose-time `/illustration` endpoint.

## Evidence

```
TestIllustrationSuppressedNotReturned (illustration_test.go)
  FAILS on buggy code: GetIllustration returns ret=0 with the suppressed image externaluid
  PASSES after fix:    GetIllustration returns ret=3 (not cached / don't show)
```

## Fix

Add `AND status = 'active'` to the `GetIllustration` query so only approved illustrations are offered via the compose preview. Suppressed/rejected/regenerating images return `ret=3`; the frontend's `getIllustration()` returns `null`, so no illustration is added to the compose attachments.

## Other Call Sites

- `microvolunteering.go:542` — already filters `status = 'active'` for challenges ✓
- `aiimage.go:502,618` — admin review/regen endpoints deliberately query `status IN ('rejected','regenerating')` ✓
- `message.go:4054` — `recordAIDeletions` looks up AI image by externaluid for deletion tracking; no status filter needed ✓

## Test

**File**: `iznik-server-go/test/illustration_test.go`
**Added**: `TestIllustrationSuppressedNotReturned`, `TestIllustrationActiveStillReturned`
**Command**: `go test ./... -run TestIllustration -v`
**Proves**: suppressed image returns ret=3 after fix; active image still returns ret=0

## Discourse

https://discourse.ilovefreegle.org/t/9753/1

### DIFF
--- a/iznik-server-go/misc/illustration.go
+++ b/iznik-server-go/misc/illustration.go
@@ -63,7 +63,10 @@ func GetIllustration(c *fiber.Ctx) error {
 	db := database.DBConn
 	var externalUID sql.NullString
 
-	err := db.Raw("SELECT externaluid FROM ai_images WHERE name = ?", itemName).Scan(&externalUID).Error
+	// Only serve active illustrations — suppressed/rejected/regenerating images are
+	// hidden by the message API, so returning them here causes a blank photo slot
+	// after the user attaches the image and submits their post (topic 9753).
+	err := db.Raw("SELECT externaluid FROM ai_images WHERE name = ? AND status = 'active'", itemName).Scan(&externalUID).Error
 
 	if err != nil || !externalUID.Valid || externalUID.String == "" {
 		// Not cached - frontend should fall back to image generation.
--- a/iznik-server-go/test/illustration_test.go
+++ b/iznik-server-go/test/illustration_test.go
@@ -127,3 +127,60 @@ func TestIllustrationLocationSuffixStripping(t *testing.T) {
 	// Clean up.
 	db.Exec("DELETE FROM ai_images WHERE name = ?", testItem)
 }
+
+// TestIllustrationSuppressedNotReturned verifies that a suppressed AI image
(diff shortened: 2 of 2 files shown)


edwh:
Closing as a bad fix (re-reviewed against the actual Discourse #9753 report).

#9753 ("AI drawings assumed to be spam", Carole) is a single terse post + screenshot — a **perception complaint** that AI-generated drawings look like / are being treated as spam. It does not describe a ghost photo slot, a compose-time problem, or attachment blanking.

This PR's premise — that `GetIllustration` returning suppressed images at compose causes a "ghost photo slot" on display — is **speculative and unrelated to the report**:
- It only applies to the edge case of *suppressed* AI images (unsuitable items, topic #9630); a normal item's illustration is `status='active'` and already works.
- Nothing in #9753 mentions a blank/ghost image slot.

The one concrete technical point (`GetIllustration` not filtering `status='active'`) is at most a minor *consistency* gap with the #9630 suppress feature — if we want the compose preview to respect suppression that should be a deliberate, separately-scoped change, not shipped as a fix for #9753. The reported issue is about how AI drawings are perceived, which is a product/UX matter, not a code bug.

No actionable code bug for #9753 → closing.

PR #632, opened 2026-06-03
fix(modtools-notifications): badge count overshoots visible actions by +1

edwh:
## Root Cause

`work.total` (the ModTools badge count) included `spam` (COLLECTION_SPAM messages) and `pendingmembers` (pending membership applications) even though V2 modtools has no page to view or action either of these items. When a moderator had even one spam message or pending member application, the badge showed 1 (or 2) more than the number of actionable items visible in the left sidebar.

In V1 PHP both fields were always 0 in practice:
- `pendingmembers` was initialised in the work struct but never populated by any query  
- `spam` used a query filtered to `COLLECTION_PENDING` messages only, so the spam branch of its loop never fired

V2 Go correctly queries both fields (spam messages from COLLECTION_SPAM, pending memberships) but then incorrectly adds them to `total`, inflating the badge with items the moderator cannot see or action.

## Evidence

The test `TestSpamAndPendingMembersNotInBadgeTotal` creates a spam message and a pending member application in a test group, takes a baseline `total` before, then asserts that `total` is unchanged after inserting the phantom items. On **buggy** code this assertion fails because `total` increases by `spamDelta + pendingMembersDelta = 2`. On **fixed** code the assertion passes.

`TestWorkCountTotalExcludesInformational` is updated to reflect the new exclusions.

## Fix

Removed `spam` and `pendingmembers` from the `total` calculation in `session/session.go`. Both fields remain in the work map so the frontend can use them if/when a spam messages page or a pending members approval page is added to V2 modtools.

## Other Call Sites

No other Go call sites compute a badge `total` that sums these fields. `groupWork.go` returns raw per-group counts without a `total` field. The Nuxt `menuCount` consumes `work.total` directly without re-summing individual fields.

## Test

`iznik-server-go/test/session_test.go`  
Test: `TestSpamAndPendingMembersNotInBadgeTotal`  
Command: `go test ./test/... -run TestSpamAndPendingMembersNotInBadgeTotal` (or via status API)  
Proves: fail-before (total inflated by spam+pendingmembers), pass-after (total unchanged)

## Discourse

https://discourse.ilovefreegle.org/t/9654/9

### DIFF
--- a/iznik-server-go/session/session.go
+++ b/iznik-server-go/session/session.go
@@ -1292,7 +1292,10 @@ func GetSession(c *fiber.Ctx) error {
 
 		// Total only includes actionable work items (primary/red badge counts),
 		// not informational ones (other/blue badge counts like chatreviewother, happiness, giftaid, pendingother).
-		total := pending + spam + pendingmembers + spammembers + pendingevents +
+		// spam (COLLECTION_SPAM messages) and pendingmembers (pending membership applications) are excluded:
+		// V2 modtools has no page to view or action either of these, so including them inflates the badge
+		// beyond what the moderator can see or act on (Discourse #9654 post 9). In V1 PHP both were always 0.
+		total := pending + spammembers + pendingevents +
 			pendingadmins + editreview + pendingvolunteering + stories +
 			spammerpendingadd + spammerpendingremove +
 			chatreview + newsletterstories + relatedmembers + housekeeping + cronjobs +
--- a/iznik-server-go/test/session_test.go
+++ b/iznik-server-go/test/session_test.go
@@ -1932,16 +1932,19 @@ func TestWorkCountTotalExcludesInformational(t *testing.T) {
 	total := work["total"].(float64)
 
 	// Total should include actionable items but NOT informational ones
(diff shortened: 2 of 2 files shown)


edwh:
Closing as a bad fix (rediagnosed for Discourse #9654).

This PR removes `spam` and `pendingmembers` from `work.total`:
- **`pendingmembers`**: a no-op — production has only 15 `Pending`-collection memberships total (all pre-2020, 0 in the last 90 days), so it contributes ~0 to any badge.
- **`spam`**: removing it is wrong. Spam-flagged messages are reviewable by mods (they surface on the Pending review page), so the count represents *actionable* work — e.g. Hackney currently has 1 spam message, which is exactly the '+1 over visible' in post #9. Hiding it from `work.total` would stop mods from ever seeing/releasing false-positive spam, rather than fixing anything.

The actual #9654 bug (post #1) is that *notification email* counts include hidden pending messages from *deleted* users (the work-badge query already excludes deleted users via `u.deleted IS NULL`, but the email-count path does not). Post #10's 466/793 is Discourse/general notifications, not the work badge. The real fix will target the email-count path; this PR is superseded.

PR #634, opened 2026-06-03
fix(modtools-notifications): exclude pending member applications from work.total

edwh:
## Root Cause

`pendingmembers` (membership applications in the `Pending` collection) was included in `work.total`, the value that drives the red badge on the ModTools hamburger menu button.

**V1 parity failure**: V1 PHP (`Group::getWorkCounts`) always initialised `pendingmembers` to 0 and never populated it from the database, so it never contributed to the work badge. V2 correctly queries the count but introduced new behaviour without a matching UI.

**No actionable UI**: ModTools has no UI to action pending member applications — there is no "Pending" filter on the Members › Approved page and no dedicated Members › Pending menu item. So a large historical backlog causes an inflated red badge with nothing visible in the menu to explain it ("no red tasks in the menu").

## Evidence

Failing test on buggy code:
\`\`\`
=== RUN   TestNewModPendingMembersNotInWorkTotal
    session_test.go:3228:
        Error:      Not equal:
                    expected: 0
                    actual  : 1
        Messages:   work.total must not include pendingmembers — no ModTools UI exists to action them (Discourse #9654)
--- FAIL: TestNewModPendingMembersNotInWorkTotal (0.10s)
\`\`\`

## Fix

Removed `pendingmembers` from the `total` arithmetic in `session.go`. The value is still computed and returned in the work object as informational data (it was informational in V1 too — always 0 but present in the response schema). The per-group `/api/group/work` endpoint is unchanged.

Updated `TestWorkCountTotalExcludesInformational` to match (removed `pendingmembers` from the expected-actionable sum).

Added `TestNewModPendingMembersNotInWorkTotal` which:
- 3a: asserts `pendingmembers > 0` after inserting a pending member (verifies the count IS returned)
- 3b: asserts `work.total` does not increase (fails on buggy code, passes after fix)

## Other Call Sites

`pendingmembers` in `group/groupWork.go` is a struct field returned by `/api/group/work`. It is not used in any total calculation there — informational only. No change needed.

## Test

- File: `iznik-server-go/test/session_test.go`
- Run: `go test ./test/... -run TestNewModPendingMembersNotInWorkTotal -v`
- Proves: FAIL before fix (total increases by 1 when pending member added), PASS after fix (total unchanged)

## Discourse
https://discourse.ilovefreegle.org/t/9654/10

### DIFF
--- a/iznik-server-go/session/session.go
+++ b/iznik-server-go/session/session.go
@@ -1291,8 +1291,16 @@ func GetSession(c *fiber.Ctx) error {
 		wg2.Wait()
 
 		// Total only includes actionable work items (primary/red badge counts),
-		// not informational ones (other/blue badge counts like chatreviewother, happiness, giftaid, pendingother).
-		total := pending + spam + pendingmembers + spammembers + pendingevents +
+		// not informational ones (other/blue badge counts like chatreviewother, happiness, giftaid,
+		// pendingother, pendingmembers).
+		//
+		// pendingmembers (membership applications in the Pending collection) is excluded because:
+		//   1. V1 PHP never counted it — Group::getWorkCounts() initialised it to 0 and never populated it.
+		//   2. No ModTools UI exists to action pending member applications (no Pending filter on the
+		//      Members > Approved page and no dedicated Members > Pending menu item).
+		//   3. Large historical backlogs (e.g. 466 on Hackney, 793 on Newham) caused a spurious red
+		//      badge flood for newly-promoted moderators with "no red tasks in the menu" (Discourse #9654).
+		total := pending + spam + spammembers + pendingevents +
 			pendingadmins + editreview + pendingvolunteering + stories +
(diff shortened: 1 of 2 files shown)


edwh:
Closing as a bad fix (rediagnosed against live production data for Discourse #9654).

This PR removes `pendingmembers` from `work.total`, on the premise that a backlog of pending member applications inflates the ModTools badge. Production data shows that's not the case: there are only **15 `Pending`-collection memberships in the entire database, all added pre-2020, zero in the last 90 days**. So `pendingmembers` contributes ~0 to any mod's badge — removing it is a no-op and does not address the reported overcount.

The real #9654 issues: (1) notification *email* counts include hidden pending messages from *deleted* users (post #1); (2) the '+1 over visible' (post #9) is an unreleased spam message (e.g. Hackney has exactly 1). Post #10's 466/793 is Discourse/general notifications, not the work badge. Those will be addressed separately.

PR #654, opened 2026-06-04
fix(message-deletion-logging): restore synchronous audit-log write on message delete

edwh:
## Root Cause

Commit `ffbfc7b1a` ("remove duplicate sync log write on message delete") removed the direct `logModAction` call from `handleDeleteMessage`, intending the batch processor to be the sole writer. But the batch task (`email_message_rejected`) runs asynchronously — so mods saw **no audit-log entry at all** immediately after deleting a message. The log eventually appeared after the batch processor ran, but that delay made it effectively invisible in the modtools audit view.

V1 parity: `Message::delete()` in PHP always called the log insertion synchronously, per group, before returning.

## Evidence

`TestHandleDeleteMessageCreatesAuditLog` (step 3b AssertFlip) **fails on the buggy code** because `handleDeleteMessage` only queues a background task — no `logs` row is written before the response returns.

```
FAIL: handleDeleteMessage must synchronously write a Message/Deleted audit-log row
      logCount = 0, want >= 1
```

`TestPostMessageDeleteWritesExactlyOneLog` confirms exactly one log row exists (not zero as the old test asserted, and not two after the batch processor runs).

## Fix

- **`iznik-server-go/message/message.go`**: Added `logModAction(...)` call in `handleDeleteMessage` for each authorised group *before* the background task loop — mirrors the pattern already used in `DeleteMessageEndpoint` and `handleOutcome`.
- **`iznik-batch/app/Console/Commands/Queue/ProcessBackgroundTasksCommand.php`**: Skip the `logs` insert in `handleModStdMessage` when `action === 'Delete Approved Message'`, preventing a duplicate entry when the batch processor later processes the queued task. All other actions (Reject, Approve, Reply) still create their log entry in the batch processor as before.

## Other Call Sites

- `DeleteMessageEndpoint` (line ~3104): already had direct `logModAction` — no change needed ✓
- `handleOutcome` Withdrawn pending (line ~3719): already had direct `logModAction` — no change needed ✓
- `handleReject` (line ~1873): action="Reject" — batch processor still creates log (no change needed) ✓

## Test

**File**: `iznik-server-go/test/message_test.go`

- `TestHandleDeleteMessageCreatesAuditLog` — new test; **fails on buggy code**, passes on fix
- `TestPostMessageDeleteWritesExactlyOneLog` — renamed from `TestPostMessageDeleteNoDuplicateLog`; assertion updated from `count==0` to `count==1` to reflect correct new behaviour

**Run via status container API:**
```
curl -s -X POST http://localhost:8081/api/tests/go
```

Tests prove the audit-log row is written synchronously by the Go handler, and that only one row appears (no duplicate from the batch processor).

## Discourse

https://discourse.ilovefreegle.org/t/9685/8

### DIFF
--- a/iznik-batch/app/Console/Commands/Queue/ProcessBackgroundTasksCommand.php
+++ b/iznik-batch/app/Console/Commands/Queue/ProcessBackgroundTasksCommand.php
@@ -489,18 +489,22 @@ class ProcessBackgroundTasksCommand extends Command
             default => 'Approved',
         };
 
-        // Always create the mod log entry (even if no stdmsg content).
-        DB::table('logs')->insert([
-            'timestamp' => now(),
-            'type' => 'Message',
-            'subtype' => $subtype,
-            'msgid' => $msgId,
-            'user' => $posterId ?: null,
-            'byuser' => $byUser,
-            'groupid' => $groupId ?: null,
-            'stdmsgid' => $stdmsgId ?: null,
-            'text' => $subject,
-        ]);
+        // Create the mod log entry for approve, reject, and reply actions.
+        // For "Delete Approved Message" the Go handler already wrote the log synchronously,
+        // so we skip it here to avoid a duplicate entry.
+        if (($data['action'] ?? '') !== 'Delete Approved Message') {
+            DB::table('logs')->insert([
+                'timestamp' => now(),
+                'type' => 'Message',
+                'subtype' => $subtype,
+                'msgid' => $msgId,
+                'user' => $posterId ?: null,
+                'byuser' => $byUser,
+                'groupid' => $groupId ?: null,
+                'stdmsgid' => $stdmsgId ?: null,
+                'text' => $subject,
+            ]);
+        }
 
         // Queue push notifications to group moderators.
         if ($groupId > 0) {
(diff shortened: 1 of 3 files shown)


edwh:
Closing — this fix is unrelated to the actual report and reverses the fix that already resolved it.

**The actual report (topic 9685):**
- Post #1 (Neville_Reid): message deletions were not being logged at all.
- Post #6 (Edward): "I did some fixes."
- Post #7 (Neville_Reid), the live issue: *"Here's a message deletion which was logged, **but it was logged twice**."*

So the outstanding problem was a **duplicate** log entry, not a missing/delayed one.

**It was already fixed** by `ffbfc7b1a` (2026-05-21, "remove duplicate sync log write on message delete — batch processor is sole writer"), which landed *after* post #7 (2026-05-20). On master today, `handleDeleteMessage` only queues the `email_message_rejected` background task and the batch processor (`handleModStdMessage`) is the sole writer, so a deletion is logged **exactly once**.

This PR re-adds the synchronous Go `logModAction` write that `ffbfc7b1a` deliberately removed, justified by a "log appears with a delay / effectively invisible" premise that was never reported. It also relies on an untested batch-skip path to avoid re-introducing the duplicate. Net effect is no improvement over the current correct behaviour, while reversing a deliberate recent fix.

(The bug row was mis-mapped to post #8, which is the automated retest reply, not a user report.)

PR #658, opened 2026-06-05
Compose rework: single-call submit + resume after login

edwh:
## Compose rework: single-call submit + resume after login

Replaces the client's draft → image → JoinAndPost dance for new posts with a
single `PUT /message/submit` call, removing the fake client-side
message/attachment ids and the draft state that went with them.

- **Server**: `message_submit.go` creates + attaches (inline by
  externaluid/Cloudflare uid) + auto-joins + posts a complete message in one
  request. Honours V1 collection logic (moderated group/member → Pending,
  otherwise Approved immediately, added to `messages_spatial` when approved).
  Old draft/image endpoints are kept for edit/repost.
- **Client**: `compose.submitSingle()` + `MessageAPI.submit()` route new
  posts through the single call; reposts are unchanged. Uploaded photos are
  carried inline by uid (no `POST /image` round-trip, no orphan
  `messages_attachments` rows).
- **Resume after forced login**: if the poster's email is already
  registered, the deferred submit is persisted (survives a refresh) and
  fires automatically once logged in, navigating to My Posts.
- **Adversarial-review hardening**: fixed an account-takeover path
  (unauthenticated submit against an existing email could post on that
  account and leak a freshly-minted password), a race on reading the new
  message id (now via `LastInsertId`), a read/write-split race that could
  silently drop the item link, and an unchecked group-membership insert
  error that could leave a post linked to no group. Added a submit-in-flight
  guard so a second click during resume can't double-post, a TTL on stale
  deferred submits, and fixed a photo-wipe on remount.

Local: Go suite and full vitest green (see individual commits for exact
counts at each stage).

## Also included: moderation oversight + auto-approve-clean

This branch carries an earlier, still-unmerged thread of moderation work
that the compose rework was built on top of:

- **Auto-approve-clean** (`AutoApproveCleanService` +
  `messages:auto-approve-clean`, scheduled every minute): posts from
  members with a NULL (never explicitly set) group posting status —
  previously stuck in Pending until a mod acted or the 48h fallback fired —
  now auto-approve after a configurable per-group delay (default 20 min) if
  content-check-clean and no danger signal (mod notes, recent
  rejections/deletions/spam, known-spammer records, pending membership
  review). A configurable sample is always held back for manual
  quality-check review.
- **Moderation views restructured** into Pending / Checked / Trusted /
  Approved: Checked and Trusted are oversight queues of live posts that
  went out via the auto-approve or trusted-member paths respectively, each
  clearable with a "mark all as checked" action once a mod has reviewed
  them. `messages_groups.checkedat`/`checkedby` record that review.
- **SysAdmin "Moderation" tab**: `GET /modtools/moderationstats` +
  `ModSysAdminModerationStats.vue` report where posts went (arrived,
  manual approve/reject, auto-approved, trusted), the auto-publish error
  rate (auto-approved posts a mod later had to intervene on), and the
  quality-check sample's reject rate, to judge whether auto-approving is
  working over a selectable period.

See `plans/active/autoapprove-delay.md` for the fuller design notes.

## Note on this rebase

This PR has now been rebased onto current `master` twice. The first pass
(2026-08-07) resolved conflicts with moderation/SysAdmin work that had
landed independently on master under different names (e.g.
`AutoApproveService`/`AutoApproveCommand` vs. this branch's
`AutoApproveCleanService`) — that overlap is still unreconciled; both
services exist side by side on this branch, and a reviewer should decide
whether this branch's auto-approve-clean / moderation-oversight scope
should still land as-is or be dropped/merged with master's version first.

The second pass (2026-08-12) rebased onto master again and resolved:
- **SysAdmin tabs**: master retired the "First reply" tab entirely
  (`chore(firstreply): drop the sysadmin dashboard`) rather than keeping it
  alongside this branch's new "Moderation" tab as before, so First reply is
  now gone from this branch too — Moderation is the only addition.
- **PhotoUploader.vue**: master had separately hardened the old
  `POST /image` registration call with a timeout guard (#1271, a hang there
  left photos stuck "uploading" forever). This branch's Phase 5 removes
  that network call altogether — the uploaded photo's Cloudflare/tusd uid
  is carried inline with no round-trip — so the specific hang the guard
  defended against can no longer happen; the now-dead timeout/import was
  dropped rather than kept alongside the inline-uid path.
- **compose.js store**: master's `storyDraft` field and this branch's
  `pendingSubmit` field are independent and both kept.
- **Deadline tests**: two compose store tests still asserted the old
  ISO-datetime deadline format; master's separate #9481 fix (send the
  deadline as a plain date, since `messages.deadline` is a DATE column)
  applies to this branch's new single-call submit path too, so the tests
  were updated to match rather than the behaviour reverted.

Verified after this rebase: `go build ./...` clean, and the full vitest
suite (733 files / 15441 tests) green.


edwh:
Pushed two fixes: (1) the failing SchedulerResilienceTest was caused by this branch's new every-minute `messages:auto-approve-clean` job using the bare `withoutOverlapping()` (24h default expiry) - now bounded to 15min matching the file's other every-minute jobs; (2) merged master in (was 27 commits behind), which also refreshes the Playwright baseline the previous run failed against. CI re-running.

PR #660, opened 2026-06-06
fix(ai-image): skip non-active illustrations in compose and batch (9753)

edwh:
## Root Cause

Two code paths returned AI illustration URLs without checking whether the illustration was active, and a third path failed to clean up ghost blank-slot attachments on existing messages:

1. **`GetIllustration` endpoint** (`iznik-server-go/misc/illustration.go`) — used during message composition, queried `ai_images` with no `status` filter. A suppressed or rejected illustration UID was returned, then masked as `''` by the message-serve query at `message.go:459`, leaving a blank image slot.

2. **`MessageIllustrationsService::processBatches`** (`iznik-batch/app/Services/MessageIllustrationsService.php`) — the batch cron had the same missing filter, attaching a suppressed/rejected UID that the serve query then blanked.

3. **Missing cleanup for existing messages** — when an illustration is suppressed or rejected AFTER being attached to messages, those messages retain the stale `messages_attachments` row. The Go message API returns `externaluid=''` for such rows, rendering a ghost blank image slot. No cleanup was removing these stale rows.

## Evidence

New AssertFlip tests fail on buggy code and pass after the fix:
- Go: `TestIllustrationSuppressedNotReturned`, `TestIllustrationRejectedNotReturned`
- PHP: `test_does_not_attach_suppressed_illustration`, `test_does_not_attach_rejected_illustration`
- PHP: `test_removes_existing_suppressed_illustration_attachment`, `test_removes_existing_rejected_illustration_attachment` (re-attempt addition)

## Fix

- `illustration.go`: add `AND status = 'active'` to GetIllustration SQL
- `MessageIllustrationsService::processBatches`: add `->where('status', 'active')` to cached-lookup
- `MessageIllustrationsService::cleanupNonActiveAttachments()` (new, re-attempt): each run deletes `messages_attachments` rows where `externalmods.ai = true` and the linked `ai_images.status` is `'suppressed'` or `'rejected'`, eliminating ghost blank slots on existing messages

## Other Call Sites

- `microvolunteering.go:542` — already filters `status = 'active'` ✓
- `aiimage.go` admin endpoints deliberately query non-active statuses ✓
- `JobIllustrationsService` — does not attach to `messages_attachments` ✓

## Test

`iznik-batch/tests/Unit/Services/MessageIllustrationsServiceTest.php`
Command: `vendor/bin/phpunit --filter=MessageIllustrationsServiceTest`
Proves: fail-before / pass-after for all new tests

## Discourse

https://discourse.ilovefreegle.org/t/9753/1

### DIFF
--- a/iznik-batch/app/Console/Commands/Message/GenerateIllustrationsCommand.php
+++ b/iznik-batch/app/Console/Commands/Message/GenerateIllustrationsCommand.php
@@ -37,9 +37,9 @@ class GenerateIllustrationsCommand extends Command
             $result = $service->processIllustrations($dryRun);
 
             if ($dryRun) {
-                $this->info("Illustrations: would clean {$result['cleaned']}, cache-hits {$result['cached_hits']}, would-fetch {$result['would_fetch']} (skipped — costs \$).");
+                $this->info("Illustrations: would clean {$result['cleaned']}, would remove non-active {$result['cleaned_non_active']}, cache-hits {$result['cached_hits']}, would-fetch {$result['would_fetch']} (skipped — costs \$).");
             } else {
-                $this->info("Illustrations: {$result['processed']} processed, {$result['cleaned']} cleaned up.");
+                $this->info("Illustrations: {$result['processed']} processed, {$result['cleaned']} cleaned up, {$result['cleaned_non_active']} non-active removed.");
                 Log::info('Message illustrations complete', $result);
             }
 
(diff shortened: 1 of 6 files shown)


edwh:
Closing as a bad fix — no fix required, as decided on #631.

#9753 is a perception/UX complaint ("AI drawings assumed to be spam"), not an actionable code bug. The FSM re-attempted it after #631 was closed; that was wrong. I've parked the bug so the monitor won't try again, and I'm fixing the monitor so a human "bad fix / no fix required" close is honoured instead of retried.

See the full analysis in #631.

PR #663, opened 2026-06-06
fix(notifications): route ModTools push to dashboard when no pending messages exist

edwh:
## Root Cause

`getBadgeCount()` counts pending + spam + volunteering items together, but `buildModToolsPayload()` always set:
- `route = '/modtools/messages/pending'`
- `title = "X messages pending"`

…when the total was > 0. If the only work items were spam or pending volunteering (no pending messages), the notification said "messages pending" and tapped through to an empty `/messages/pending` page.

## Evidence

Failing test output (step 3b, before fix):

```
1) Tests\Unit\Services\PushNotificationServiceTest::test_buildModToolsPayload_spam_only_routes_to_dashboard
When only spam exists (no pending messages), route must be /modtools — Discourse #9692
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'/modtools'
+'/modtools/messages/pending'

/var/www/html/tests/Unit/Services/PushNotificationServiceTest.php:520

FAILURES!
Tests: 1, Assertions: 2, Failures: 1
```

## Fix

Added `getPendingMessageCount()` — a targeted query counting only unheld pending messages in the user's active groups. `buildModToolsPayload()` uses it to choose route and label:

- `pendingCount > 0` → `route = /modtools/messages/pending`, title = "X messages pending" (existing behaviour, unchanged)
- `pendingCount == 0` → `route = /modtools` (dashboard), title = "X items to review"

Badge count remains `total` (pending + spam + volunteering) in both cases.

## Other Call Sites

`notifyTest()` always routes to `/modtools` regardless — already correct, no change needed.

No other hardcoded `/modtools/messages/pending` route strings in iznik-batch, iznik-server-go, or iznik-nuxt3.

## Test

**File:** `iznik-batch/tests/Unit/Services/PushNotificationServiceTest::test_buildModToolsPayload_spam_only_routes_to_dashboard`

**Command:** filter=`PushNotificationServiceTest::test_buildModToolsPayload_spam_only_routes_to_dashboard` via status API

**Proves:** spam-only setup → badge=1, route=/modtools, title doesn't say "pending". After fix: Tests: 1, Assertions: 3, OK.
Full `PushNotificationServiceTest` (35 tests) passes with no regressions.

## Discourse
https://discourse.ilovefreegle.org/t/9692/10

### DIFF
--- a/iznik-batch/app/Services/PushNotificationService.php
+++ b/iznik-batch/app/Services/PushNotificationService.php
@@ -479,8 +479,11 @@ class PushNotificationService
     /**
      * Build the ModTools notification payload.
      *
-     * For ModTools, we send a simple "pending messages" notification.
-     * Matches legacy User::getNotificationPayload(modtools=true).
+     * Routes to /modtools/messages/pending only when there are actual pending
+     * messages — spam-only or volunteering-only work routes to the dashboard
+     * (/modtools) so tapping never lands on an empty pending queue.
+     * Fixes Discourse #9692: notification wrongly labelled "pending" and routing
+     * to /messages/pending when the badge count came from spam/volunteering only.
      */
     private function buildModToolsPayload(int $userId): ?array
     {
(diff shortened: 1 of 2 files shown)


edwh:
Closing — the approach is flawed (confirmed by the monitor's own adversarial review).

`getPendingMessageCount()` re-implements `getBadgeCount()` with a *separate* pending-count query and a parallel memberships/active-mod filter. The two must stay perfectly in sync or push routing breaks: a held pending message (m.heldby IS NULL etc.) can be counted in the badge total but excluded here, so the notification routes to /modtools (or away from it) incorrectly.

The correct fix is to have `getBadgeCount` return a per-collection breakdown and read `breakdown.pending`, rather than a copy-pasted query that can silently diverge. Closing rather than auto-retrying; the underlying routing issue is parked for a properly-designed fix.

PR #665, opened 2026-06-06
fix(notifications): exclude orphaned Pending memberships from work-count (Discourse #9654/12)

edwh:
## Root Cause

The `pendingmembers` count queries in `session.go` (work total badge) and `groupWork.go` (per-group breakdown) tallied memberships **without JOINing to the users table**. Orphaned membership rows — where the user was hard-deleted (before FK CASCADE was in place) but the `memberships` row with `collection='Pending'` was left behind — were counted by the query but returned zero by the listing query (which INNER-JOINs users). This produced a persistent phantom `+1` badge that moderators could never clear.

This is the identical bug pattern fixed for `Spammembers` in PR #590, now fixed for `Pendingmembers`.

## Evidence

```
// BEFORE (buggy):
db.Raw("SELECT COUNT(*) FROM memberships WHERE groupid IN ? AND collection = ?",
    modGroupIDs, utils.COLLECTION_PENDING).Scan(&pendingmembers)
// → returns 1 even when the only pending member's user row is deleted

// AFTER (fixed):
db.Raw("SELECT COUNT(*) FROM memberships m "+
    "INNER JOIN users u ON u.id = m.userid AND u.deleted IS NULL "+
    "WHERE m.groupid IN ? AND m.collection = ?",
    modGroupIDs, utils.COLLECTION_PENDING).Scan(&pendingmembers)
// → returns 0 when user row is deleted (matches listing behaviour)
```

The test `TestPendingMembersOrphanCountVsListing` reproduces the scenario:
- Creates a Pending membership, then hard-deletes the user row with FK checks off
- Asserts `group/work Pendingmembers == 0` (fails on buggy code: returns 1)
- Asserts `session work.pendingmembers == 0` (fails on buggy code: returns 1)
- Asserts `GET /memberships?collection=Pending` returns 0 rows

## Fix

Added `INNER JOIN users u ON u.id = m.userid AND u.deleted IS NULL` to both count queries:
- `session.go` — the work-total badge count
- `groupWork.go` — the per-group breakdown count

Both fixes are identical in structure to PR #590's `Spammembers` fix.

## Other Call Sites

`grep -rn "pendingmembers"` — only the two fixed sites compute this count. The listing queries (`GetMemberships` with `collection=Pending`) already JOIN users and are unaffected.

## Test

**File**: `iznik-server-go/test/member_pending_orphan_test.go`  
**Command**: `go test ./test/... -run TestPendingMembersOrphanCountVsListing -v`  
**Proves**: fails-before (returns 1 when orphan present) / passes-after (returns 0)

## Discourse

https://discourse.ilovefreegle.org/t/9654/12

### DIFF
--- a/iznik-server-go/group/groupWork.go
+++ b/iznik-server-go/group/groupWork.go
@@ -180,13 +180,16 @@ func GetGroupWork(c *fiber.Ctx) error {
 	}()
 
 	// --- Pending members: all groups, no active/inactive split ---
+	// INNER JOIN users excludes orphaned membership rows (user deleted, membership remains)
+	// so the count matches the listing query (Discourse #9654/12, mirrors PR #590 for Spammembers).
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
 		var rows []countRow
-		db.Raw("SELECT groupid, COUNT(*) as count FROM memberships "+
-			"WHERE groupid IN ? AND collection = ? "+
-			"GROUP BY groupid", allGroupIDs, utils.COLLECTION_PENDING).Scan(&rows)
+		db.Raw("SELECT m.groupid, COUNT(*) as count FROM memberships m "+
+			"INNER JOIN users u ON u.id = m.userid AND u.deleted IS NULL "+
+			"WHERE m.groupid IN ? AND m.collection = ? "+
+			"GROUP BY m.groupid", allGroupIDs, utils.COLLECTION_PENDING).Scan(&rows)
 		mapMutex.Lock()
 		for _, r := range rows {
 			if w := workMap[r.Groupid]; w != nil {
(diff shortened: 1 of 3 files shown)


edwh:
Closing — bad fix, wrong place (verified against live production data).

**Why it's a no-op:** #665 adds `INNER JOIN users ... u.deleted IS NULL` to the pending-**members** count. Live data: there are **0 orphaned pending memberships** (all 15 pending memberships have live users), so the join changes the count by nothing.

**Wrong place:** the iOS/Android badge isn't computed by session.go at all — it's `PushNotificationService::getBadgeCount()` in iznik-batch, which doesn't even count pending members.

**Real root cause of #9654/12 (stuck at +1):** `getBadgeCount()` counts pending **messages** from **deleted users** (its pending query only checks `m.fromuser IS NOT NULL`, missing `u.deleted IS NULL`), whereas the app menu (session.go) filters them out. Live data: **32 such phantom pending messages across 24 groups** — counted in the badge, hidden in the menu, so the badge never clears. Same bug pattern as #9547, on a query that was missed.

**Correct fix:** add `INNER JOIN users u ON u.id = m.fromuser AND u.deleted IS NULL` to the pending (and spam) queries in `getBadgeCount()`, mirroring session.go.

PR #666, opened 2026-06-06
fix(modtools): show manually-reposted Pending messages in posting history

edwh:
## Root Cause

`GetUserMessageHistory` (used by ModPostingHistoryModal via `GET /api/user/:id?modtools=true`) filtered messages to `collection = 'Approved'` only.

When a member manually reposts a message (BackToDraft → edit → JoinAndPost), the message re-enters the content-check pipeline as `Pending`. During that window — until the `messages:contentcheck` batch job promotes it to Approved — the post was invisible in the member's Posts history.

Auto-reposts (`AutoRepostService`) are unaffected because they keep the message in Approved by updating `arrival` in-place rather than moving it through JoinAndPost.

## Fix

Include `Pending` alongside `Approved` in the `GetUserMessageHistory` query:

```go
// Before:
"WHERE m.fromuser = ? AND mg.deleted = 0 AND m.deleted IS NULL AND mg.collection = ?"
// userid, COLLECTION_APPROVED

// After:
"WHERE m.fromuser = ? AND mg.deleted = 0 AND m.deleted IS NULL AND mg.collection IN (?, ?)"
// userid, COLLECTION_APPROVED, COLLECTION_PENDING
```

`Rejected` is intentionally excluded (rejected messages should not count as active posts).

The `UserMessageHistory` struct already carries a `Collection` field, and `ModPostingHistoryModal.vue` already renders `collection === 'Pending'` in red — so no frontend changes are needed.

## Test

Added `TestManualRepostPendingVisibleInHistory` (AssertFlip pattern):
- Inserts a message in `Pending` collection for a user
- Fetches that user with `modtools=true`
- **FAILS** on buggy code (Pending filtered out of messagehistory)
- **PASSES** after fix (Pending included)

The pre-existing `TestRecentWanted_GoAPI_ExcludesNonApproved` continues to pass (Rejected still excluded).

## Discourse

https://discourse.ilovefreegle.org/t/9481/562

### DIFF
--- a/iznik-server-go/test/message_test.go
+++ b/iznik-server-go/test/message_test.go
@@ -1986,6 +1986,72 @@ func TestJoinAndPostRejectsEmptyDraft(t *testing.T) {
 	assert.Equal(t, int64(0), msgGroupCount, "Empty draft must not produce a messages_groups row")
 }
 
+// TestManualRepostPendingVisibleInHistory verifies that a user's message appears in
+// their posting history while it is in Pending collection (e.g. after a manual repost).
+// Bug: GetUserMessageHistory filtered to collection='Approved' only, so a manually
+// reposted message was invisible in the Posts history until the content-check batch
+// approved it. Discourse #9481 post 562.
+//
+// FAILS on buggy code (Pending filtered out), PASSES after fix (Pending included).
+func TestManualRepostPendingVisibleInHistory(t *testing.T) {
+	prefix := uniquePrefix("repost_pending_history")
+	db := database.DBConn
+
+	groupID := CreateTestGroup(t, prefix)
+	posterID := CreateTestUser(t, prefix+"_poster", "User")
+	modID := CreateTestUser(t, prefix+"_mod", "User")
+	CreateTestMembership(t, posterID, groupID, "Member")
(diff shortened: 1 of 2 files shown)


edwh:
Closing — bad fix, misses the point.

This makes a manually-reposted **Pending** message visible in the Posts history (`GetUserMessageHistory` now includes `collection='Pending'`). That's a band-aid on the symptom and doesn't address what Neville actually reported (#9481/562):

> auto reposts don't go through Pending, they just repost as Approved. However, manual reposts get held in Pending. This wastes time for volunteers...

The real issue is the **manual-vs-auto repost moderation routing**: a manual repost of an already-approved message shouldn't re-enter Pending / require re-moderation at all — it should go straight to Approved like an auto-repost. The correct fix is in the repost/moderation path (make manual reposts of already-vetted content skip Pending), not in the Posts-history display.

Surfacing Pending (unmoderated-state) items in the Posts history is also questionable on its own — it can misrepresent moderation state. Closing rather than auto-retrying; the routing issue is parked for a properly-scoped fix.

(Note: this PR's agent timed out mid-task and was never adversarially reviewed.)

PR #669, opened 2026-06-06
fix(authentication): scope GetSession to the requesting session to fix cross-context logout (#9748/4)

edwh:
## Summary

- **Root cause**: `GetSession` ran `SELECT … FROM sessions WHERE userid = ? LIMIT 1`, returning an arbitrary (typically oldest) session row regardless of which session authenticated the request. When a user is logged into both Freegle and ModTools, the second app's GET /session overwrote its auth store with the first app's series. Subsequent logout then called DELETE with the wrong series, clearing the other app's sessions.
- **Fix**: Extract the session ID from the request's JWT claim (or Authorization2 persistent-token ID) and query `WHERE id = ? AND userid = ?` to return only the authenticating session's row. Falls back to `LIMIT 1` only when neither auth method identifies a specific session (backward compatibility).
- **Existing fix preserved**: DeleteSession's series-based deletion from #611 is unchanged and correct; this fix ensures the series data it relies on is always sourced from the right session.

## Test plan

- [ ] `TestGetSessionReturnsSendersSeries` (new) — creates two sessions with distinct series, calls GET /session authenticated as the newer session, asserts `persistent.id` and `persistent.series` match the authenticating session, not the LIMIT 1 result
- [ ] `TestDeleteSessionScopedToCurrentSeries` (existing) — verifies DELETE /session only removes the caller's series
- [ ] Full Go test suite passes

Discourse: https://discourse.ilovefreegle.org/t/9748/4

### DIFF
--- a/iznik-server-go/session/session.go
+++ b/iznik-server-go/session/session.go
@@ -835,7 +835,31 @@ func GetSession(c *fiber.Ctx) error {
 	}()
 	go func() {
 		defer wg.Done()
-		db.Raw("SELECT id, series, token FROM sessions WHERE userid = ? LIMIT 1", myid).Scan(&sessionRow)
+		// Use the session that authenticated this specific request so the
+		// returned JWT/persistent carries the correct series for this context.
+		// LIMIT 1 returned an arbitrary (typically oldest) session, so a user
+		// logged into both Freegle and ModTools would receive the wrong series
+		// after GET /session, causing subsequent logout to delete the wrong
+		// app's sessions. Discourse #9748.
+		if _, reqSessionId, _ := user.GetJWTFromRequest(c); reqSessionId > 0 {
+			db.Raw("SELECT id, series, token FROM sessions WHERE id = ? AND userid = ?",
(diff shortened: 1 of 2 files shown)


edwh:
Closing — V1 parity (working-as-designed), not a regression.

V1 (iznik-server `Session::destroy($userid, $series)`) deletes ALL of a user's sessions on an explicit logout — the code path is gated exactly on this, with the comment:

> If we're doing an explicit logout we're called with a null $series and want to zap all sessions for this user.

So "logging out logs you out everywhere" is **intentional V1 behaviour**, and V1's session lookup is also by `userid` (session.php:677) — the same pattern this PR treats as the bug. #9748's premise that ModTools sessions were independent doesn't hold against the V1 code; scoping logout to one device would be a deliberate product change (a feature), not a regression fix, and would need its own design (session/device identity, not just GetSession series selection).

This is the STEP 1a V1-parity check that should have deferred it up front. Closing; the request is parked as a possible feature for human decision rather than auto-retried.

PR #670, opened 2026-06-06
fix(image-upload): mark aborted TUS upload as error to prevent stuck uploading state

edwh:
## Root cause

`PhotoUploader.vue` keeps a single `uploadInstance` variable for the in-progress TUS upload. When a new upload starts while one is already in-flight, it calls `uploadInstance.abort()`. The TUS library's `abort()` never fires `onError` or `onSuccess`, so the `Promise` wrapping the first upload hangs forever, leaving `photo.uploading` stuck at `true` with no error shown. On slow mobile connections (Samsung Android) users add several photos in quick succession — each new photo aborts the previous one — and every aborted photo ends up permanently stuck in a spinner state. After 6–7 uploads the component appears broken.

## Fix

Track the in-progress photo alongside `uploadInstance` in a new `uploadingPhoto` variable. On abort, immediately mark that photo `error = true` / `uploading = false` so the user gets feedback and can retry, rather than being silently blocked.

```diff
+let uploadingPhoto = null // Photo associated with the current in-progress TUS upload

 if (uploadInstance) {
   uploadInstance.abort()
+  if (uploadingPhoto && uploadingPhoto.uploading) {
+    uploadingPhoto.error = true
+    uploadingPhoto.uploading = false
+  }
 }

+uploadingPhoto = photo
 uploadInstance = new tus.Upload(file, { ... })
```

## Evidence

- Discourse topic 9764 post 1: "Member on Samsung Android cannot upload images after 6-7 attempts."
- The abort path is the only code path that can leave `photo.uploading = true` permanently — `onError` and `onSuccess` always clear it; `abort()` does not.

## Test

New unit test in `PhotoUploader.spec.js` (suite `concurrent upload abort (9764-1)`):
- Calls `processPhoto` twice without `await` to simulate rapid mobile photo selection
- Verifies the aborted first photo ends up with `uploading = false` and `error = true`
- **Fails on unfixed code** ("expected true to be false"); **passes after fix** (89/89 tests passing)

## Donation error (second symptom from 9764)

The report also mentioned "donation via PayPal card payment shows an error message." After reviewing `StripeDonate.vue`, `DonationButton.vue`, and `donate.vue`: no Freegle-owned code defect was identified. The error originates inside the Stripe/PayPal payment processor iframe. Would need a reproduction with browser devtools to investigate further.

## Discourse link

https://discourse.ilovefreegle.org/t/9764

### DIFF
--- a/iznik-nuxt3/components/PhotoUploader.vue
+++ b/iznik-nuxt3/components/PhotoUploader.vue
@@ -254,6 +254,7 @@ const pendingPhoto = ref(null)
 // Always show first photo as the featured/primary one
 const selectedIndex = ref(0)
 let uploadInstance = null
+let uploadingPhoto = null // Photo associated with the current in-progress TUS upload
 let tempIdCounter = 0
 
 // Computed property for the selected photo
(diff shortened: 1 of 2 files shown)


edwh:
Closing — fails the monitor's own adversarial review: "Partial implementation (symptom, not root cause)". The report names the root cause (`tus.abort()` handling on aborted uploads); this patches the symptom instead. The image-upload issue (9764.1) is parked for a root-cause fix rather than auto-retried.

PR #672, opened 2026-06-08
fix(community-settings): TN settings link missing on first load (#9767)

edwh:
## Root Cause

The batch fetch (`getModGroups` → `fetchGroupsMTBatch`) populates the group store without `tnkey` data, because the Go batch endpoint (`getMultipleGroups`) never fetches it — `tnkey` requires an external HTTP call to TrashNothing per group, prohibitively expensive for a batch of 25+ groups.

When `ModSettingsGroup.fetchGroup()` called `fetchIfNeedBeMT(id)`, that function returned early (group already in store), so `group.tnkey` stayed undefined and `v-if="group.tnkey && group.tnkey.url"` never rendered the link.

**Design decision**: The batch endpoint correctly omits tnkey. The fix belongs on the frontend: Community Settings must call the individual group endpoint (`fetchGroupMT`) which fetches tnkey, not rely on the cached batch result.

## Evidence

**Before fix — 3 failures** (on old code with new tests):
```
FAIL bug #9767 > calls fetchGroupMT on first mount (regression guard)
FAIL fetchGroup function > fetchGroup calls fetchGroupMT  
FAIL watch effects > fetches group when groupid changes
```
**After fix — all 50 pass.**

## Fix

Changed `fetchGroup()` in `ModSettingsGroup.vue` to call `fetchGroupMT` instead of `fetchIfNeedBeMT`. `fetchGroupMT` always calls the individual group endpoint with `tnkey=true`, ensuring the TN settings link data is present on first load.

## How this differs from rejected #672

The code change is the same. The test was rewritten:

| | Old #672 | This re-attempt |
|---|---|---|
| TN link renders when tnkey present | missing | new test |
| TN link hidden when tnkey absent (bug state) | missing | new test |
| `fetchGroupMT` called on mount | only test | regression guard, now supported by the two user-visible tests |

Tests 1+2 prove *why* calling `fetchGroupMT` matters (link needs tnkey data); test 3 is the regression guard.

## Other Call Sites

Other `fetchIfNeedBeMT` callers (ModMessage, ModStdMessageModal, ModSupportFindGroup, ModSettingsModConfig) display group info without needing `tnkey` — correct as-is.

## Test

**File**: `iznik-nuxt3/tests/unit/components/modtools/ModSettingsGroup.spec.js`

**Command**: `npm run test:unit:run -- tests/unit/components/modtools/ModSettingsGroup.spec.js`

3 new tests in `describe('bug #9767: TN settings link missing on first load')` — fail on old code, pass on fix.

## Discourse

https://discourse.ilovefreegle.org/t/9767/1

### DIFF
--- a/iznik-nuxt3/modtools/components/ModSettingsGroup.vue
+++ b/iznik-nuxt3/modtools/components/ModSettingsGroup.vue
@@ -1417,7 +1417,7 @@ async function fetchGroup() {
   if (!groupid.value) return
   editingDescription.value = false
 
-  await modGroupStore.fetchIfNeedBeMT(groupid.value)
+  await modGroupStore.fetchGroupMT(groupid.value)
   const groupData = modGroupStore.get(groupid.value)
   let groupRules = groupData?.rules || {}
   // console.log('fetchGroup rules',groupRules)
--- a/iznik-nuxt3/tests/unit/components/modtools/ModSettingsGroup.spec.js
+++ b/iznik-nuxt3/tests/unit/components/modtools/ModSettingsGroup.spec.js
@@ -773,11 +773,31 @@ describe('ModSettingsGroup', () => {
   })
 
   describe('bug #9767: TN settings link missing on first load', () => {
-    // The batch fetch (getModGroups → fetchGroupsMTBatch) ignores the tnkey
-    // query parameter in the Go batch endpoint, so cached groups have no tnkey.
-    // fetchIfNeedBeMT returns early for cached groups — skipping the tnkey fetch.
-    // Fix: fetchGroup must call fetchGroupMT directly so tnkey is always fetched.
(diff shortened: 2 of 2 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #672) — this fix did not pass and is not ready to merge.

- **Partial implementation: symptom not root cause**: The author's own comment in ModSettingsGroup.spec.js (the new describe block, lines ~772-775) states the root cause explicitly: 'The batch fetch (getModGroups → fetchGroupsMTBatch) ignores the tnkey query parameter in the Go batch endpoint, so cached groups have no tnkey.' The diff does NOT touch th
- **Test proves nothing / not a reproduction test**: The new test (spec lines 776-780) asserts only `expect(mockModGroupStore.fetchGroupMT).toHaveBeenCalledWith(123)`. It mirrors the production change one-for-one and verifies an implementation detail (which store method was invoked), not the user-visible bug condition. It never checks that tnkey is ac

PR #673, opened 2026-06-08
fix(modtools-editing): show and preserve neighbourhood name when editing posts

edwh:
## Root Cause

`ModMessage.vue` only displayed and transmitted the raw postcode string (e.g. `E1 1AA`) in its edit form — the neighbourhood/area name (e.g. `Stepney`) stored in `message.location.areaname` was never shown or included in the `messageStore.patch()` call.

On every save the Go API's `constructLocationString` re-derived the area name from the DB postcode mapping, permanently overwriting any custom area name the original moderator had set.  The PostCode field therefore acted as a one-way reset button.

## Evidence

```
// ModMessage.vue save() before fix — areaname never sent
await messageStore.patch({
  id: ..., msgtype: ..., item: ..., location: editmessage.value.location.name,
  //   ^^^ only raw postcode, no areaname
})
```

## Fix

**Frontend (`ModMessage.vue`)**
- Added editable `<b-form-input>` for `editmessage.location.areaname` alongside the PostCode field so the neighbourhood name is visible and editable.
- Added `areaname: editmessage.value.location.areaname` to `messageStore.patch()`.

**Go API (`message/message.go`)**
- Added `Areaname *string` to `patchMessageRequest`.
- Extracted new pure helper `buildLocStrFromInfo(locName, locType, dbAreaName, customArea string)` — testable without a DB.
- PATCH subject-rebuild path: when `req.Areaname` is non-empty, uses the custom area instead of the DB-derived default; falls back to `constructLocationString` when no area is supplied (existing behaviour unchanged).

## Other Call Sites

`constructLocationString` has three call sites:
- `message.go:2414` — POST create path; no moderator override needed; unchanged.
- `message.go:2767` — PATCH edit path; patched (uses custom area when supplied).
- `message.go:3433` — Draft subject rebuild on submission; no override needed; unchanged.

## Test

**Go pure tests** (`message_pure_test.go`): 8 new cases for `buildLocStrFromInfo` covering postcode + DB area, postcode + custom area override, postcode no area, non-postcode types.

**Vue unit tests** (`ModMessage.spec.js`): 2 regression tests confirming `messageStore.patch` receives the `areaname` field with both default and custom values.

Full suite: 596/597 files pass — the single failing file (`settings_ProfileSection.spec.js`) is pre-existing and unrelated (reproduces on master HEAD).

## Discourse

https://discourse.ilovefreegle.org/t/9769/1

### DIFF
--- a/iznik-nuxt3/modtools/components/ModMessage.vue
+++ b/iznik-nuxt3/modtools/components/ModMessage.vue
@@ -49,6 +49,12 @@
               </div>
               <div v-if="editmessage.item && editmessage.location">
                 <b-input-group>
+                  <b-form-input
+                    v-model="editmessage.location.areaname"
+                    size="lg"
+                    class="me-1 areaname-input"
+                    placeholder="Neighbourhood"
+                  />
                   <PostCode
                     :value="editmessage.location.name"
                     :find="false"
@@ -1236,6 +1242,7 @@ async function save() {
         msgtype: editmessage.value.type,
         item: editmessage.value.item.name,
         location: editmessage.value.location.name,
+        areaname: editmessage.value.location.areaname,
         attachments: attids,
         textbody: editmessage.value.textbody,
       })
@@ -1444,6 +1451,10 @@ function spamReport() {
   /* Cap growth on wide desktop so the field doesn't stretch the whole row.
      flex-grow-1 keeps it filling available space up to this cap. */
   max-width: 500px;
+}
+
+.areaname-input {
+  max-width: 160px;
 }
 
 .fullsubject {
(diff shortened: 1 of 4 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #673) — this fix did not pass and is not ready to merge.

- **Duplicate implementation**: buildLocStrFromInfo reimplements logic that constructLocationString already owns. Its own doc-comment calls it 'the pure, database-free core of the location string builder', yet constructLocationString (still used in the else branch of applyPatchMessageCore and documented as building 'the area name 
- **Test proves nothing / coverage of exact trigger path**: The actual fix lives in the new branch in applyPatchMessageCore (the `if req.Areaname != nil && *req.Areaname != ""` decision, the SELECT l.name,l.type,areaid query, and the fallback to constructLocationString). That integration logic has ZERO test coverage. The eight Go tests in message_pure_test.g

PR #674, opened 2026-06-08
fix(ios-badge): include notification count in app badge (topic 9654 post 13)

edwh:
## Root Cause

`chatCount` computed in `useNavbar.js` called `mobileStore.setBadgeCount(count)` using only `chatStore.unreadCount`, ignoring `notificationStore.count`. The server formula in `PushNotificationService::buildUserNotificationPayload` is `badge = chatcount + notifcount`. When a user deleted chat-review items, `chatStore.unreadCount` dropped to 0, the computed fired `setBadgeCount(0)`, and the OS home-screen badge cleared — even though a notification still existed and was visible in the in-app bell.

## Evidence

- **Reporter**: Neville_Reid, [Discourse topic 9654 post 13](https://discourse.ilovefreegle.org/t/9654/13)
- **Symptom**: Badge showed 3 on app open, then 1 after deleting chat reviews, then 0 after browsing menus — while in-app count stayed at 1.
- **Server code** (`iznik-batch/app/Services/PushNotificationService.php`): `$total = $chatcount + $notifcount`, `'badge' => (string)$total`. Client code did not match.

## Fix

PR #674 added `notificationStore.count` to the badge but did NOT clamp the combined value. With 99 unread chats + 5 notifications it called `setBadgeCount(104)`. This re-attempt wraps the sum: `Math.min(99, count + (notificationStore.count || 0))`.

## Other Call Sites

- `mobile.js` badge updates via push payload — server-supplied value already correct
- `modtools/` badge calls — separate ModTools badge semantics, unaffected

## Test

`iznik-nuxt3/tests/unit/composables/useNavbar.spec.js`
Command: `npx vitest run tests/unit/composables/useNavbar.spec.js`

| Test | Scenario | Before fix | After fix |
|------|----------|-----------|-----------|
| sets badge to chatCount + notificationCount | chat=0, notif=1 → 1 | ❌ setBadgeCount(0) | ✅ setBadgeCount(1) |
| sets badge to 0 when both are 0 | chat=0, notif=0 → 0 | ✅ | ✅ |
| clamps combined badge to 99 | chat=99, notif=5 → 99 | ❌ setBadgeCount(104) | ✅ setBadgeCount(99) |

All 37 vitest tests pass.

## Discourse
https://discourse.ilovefreegle.org/t/9654/13

### DIFF
--- a/iznik-nuxt3/composables/useNavbar.js
+++ b/iznik-nuxt3/composables/useNavbar.js
@@ -103,7 +103,7 @@ export function useNavbar() {
   const chatCount = computed(() => {
     const count = Math.min(99, chatStore.unreadCount)
     if (mobileStore.isApp) {
-      mobileStore.setBadgeCount(count + (notificationStore.count || 0))
+      mobileStore.setBadgeCount(Math.min(99, count + (notificationStore.count || 0)))
     }
     return count
   })
--- a/iznik-nuxt3/composables/useNavbar.js
+++ b/iznik-nuxt3/composables/useNavbar.js
@@ -103,7 +103,7 @@ export function useNavbar() {
   const chatCount = computed(() => {
     const count = Math.min(99, chatStore.unreadCount)
     if (mobileStore.isApp) {
-      mobileStore.setBadgeCount(count)
+      mobileStore.setBadgeCount(count + (notificationStore.count || 0))
     }
     return count
   })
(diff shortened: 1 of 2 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #674) — this fix did not pass and is not ready to merge.

- **Partial/incorrect implementation**: The fix defeats the existing 99 clamp. On the changed line, `count = Math.min(99, chatStore.unreadCount)` is clamped, but the new badge value `count + (notificationStore.count || 0)` is NOT clamped. So with 99 unread chats and 5 notifications the code now calls `setBadgeCount(104)`, whereas before i

PR #675, opened 2026-06-08
fix(modtools-mapping): midpoint-drag vertex lost on save (ST_Simplify removes nearby vertices)

edwh:
## Root Cause

`UpdateLocation` applied `ST_Simplify(polygon, 0.001)` to every polygon before saving. This Douglas-Peucker pass removes any vertex within **0.001 degrees (~111 m)** of the straight line between its neighbours. Dragging a geoman midpoint marker creates exactly such a vertex — it starts on the original edge and is moved only slightly before the user clicks Save.

## Evidence

- **Post 1 (point not saved at all)**: midpoint dragged < ~111 m → dropped by `ST_Simplify` → polygon unchanged on reload.
- **Post 2 (adjacent vertices deleted)**: midpoint dragged far enough to survive → the *original* adjacent vertices are now within 0.001° of the new edge lines → they are dropped.
- Confirmed client-side behaviour is correct: geoman's `_onMarkerDragEnd` fires `pm:edit` with the final position; `wicket.fromObject(layer).write()` produces the correct 5-vertex WKT; the UI calls `PATCH /api/locations` with it. The bug is entirely server-side.

## Fix

Remove the `ST_Simplify` block (7 lines) from the save path in `UpdateLocation`. The `ST_Simplify` calls in the **SELECT** queries (lines 478–480) are for display only and are left unchanged.

The simplification was added in `9cd0dcb28` to "match V1 behaviour" but V1 only simplified on read/display, never on write.

## Other Call Sites

`ST_Simplify` also appears at lines 478–480 in the location **fetch** query (wrapped in `CASE WHEN ... IS NULL`) — these are safe display-only simplifications and are NOT changed.

## Test

`TestUpdateLocationPreservesNearbyVertex` — saves a 5-vertex polygon (Edinburgh area) where the 5th vertex sits 0.0005° above the base edge (well inside the old tolerance), then asserts `ST_NumPoints(ST_ExteriorRing(ourgeometry)) = 6` (5 distinct vertices + the closing copy). Confirmed:
- FAILS on buggy code: `expected 6, actual 5`
- PASSES after fix: `actual 6`

All existing `TestUpdate*` and `TestLocation*` tests continue to pass.

## Discourse

https://discourse.ilovefreegle.org/t/9770/1 — Reporter: Neville_Reid, ref Limehouse group polygon editing.

### DIFF
--- a/iznik-server-go/location/location.go
+++ b/iznik-server-go/location/location.go
@@ -795,14 +795,6 @@ func UpdateLocation(c *fiber.Ctx) error {
 			return fiber.NewError(fiber.StatusBadRequest, "Invalid geometry")
 		}
 
-		// Simplify the polygon to reduce complexity, matching V1 behaviour.
-		var simplified string
-		db.Raw(fmt.Sprintf("SELECT ST_AsText(ST_Simplify(ST_GeomFromText(?, %d), 0.001)) AS simplified", utils.SRID), *req.Polygon).Scan(&simplified)
-
-		if simplified != "" {
-			req.Polygon = &simplified
-		}
-
 		// Capture old geometry and compute union with new for remap scope (matching V1).
 		// If old and new intersect, remap the union (covers both). If separate, remap both.
 		type OldGeom struct {
--- a/iznik-server-go/location/location.go
+++ b/iznik-server-go/location/location.go
@@ -795,6 +795,14 @@ func UpdateLocation(c *fiber.Ctx) error {
 			return fiber.NewError(fiber.StatusBadRequest, "Invalid geometry")
 		}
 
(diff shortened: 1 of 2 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #675) — this fix did not pass and is not ready to merge.

- **Removes documented intent / V1-parity regression**: The deleted block was explicitly annotated 'Simplify the polygon to reduce complexity, matching V1 behaviour.' It is removed wholesale with no replacement comment justifying the deliberate divergence from V1. Two consequences: (1) if the V1 PHP UpdateLocation path still calls ST_Simplify (which the 

PR #681, opened 2026-06-09
fix(modtools-editing): show and preserve neighbourhood name when editing posts

edwh:
## Root Cause

The ModTools edit form showed only the raw postcode (e.g. `E1 1AA`) via the PostCode component. The area/neighbourhood name (`areaname`, e.g. `Stepney`) was never displayed or sent to the API. On every save, `constructLocationString` re-derived the area name from the DB postcode mapping, silently overwriting any custom spelling the moderator had set.

## Evidence

```
// ModMessage.vue save() before fix — areaname never sent
await messageStore.patch({
  id: ..., location: editmessage.value.location.name,
  //   ^^^ only raw postcode, areaname omitted
})

// constructLocationString always queries DB area, ignores any mod input
locStr := constructLocationString(db, req.ID)  // returns DB-derived "Stepney E1"
```

## Fix vs Rejected #673

**PR #673 was rejected for two reasons — both fixed here:**

1. **No duplicate helper** — #673 added `buildLocStrFromInfo` which reimplemented `constructLocationString`'s logic. This re-attempt folds the area-override into `constructLocationString` itself via an optional variadic parameter (`overrideAreaName ...string`), so there is exactly one function handling all callers. Existing call sites pass no override and behaviour is unchanged.

2. **Integration test coverage** — #673 only tested the pure helper, leaving `applyPatchMessageCore`'s PATCH path with zero coverage. This re-attempt adds two integration tests in `message_test.go` that exercise the actual `PATCH /message` endpoint:
   - `TestPatchMessageAreanameOverride` — confirms mod-supplied areaname appears in the rebuilt subject instead of the DB default
   - `TestPatchMessageAreanameOmittedKeepsDBDefault` — confirms existing DB-default behaviour is unchanged when no areaname is supplied

## Other Call Sites

`constructLocationString` has three call sites:
- `message.go:2396` — POST create path; no moderator override needed; unchanged (no variadic arg passed).
- `message.go:2739/2741` — PATCH edit path; patched to pass `*req.Areaname` when provided.
- `message.go:3407` — Draft subject rebuild on submission; no override needed; unchanged.

## Test

| File | Command | Proves |
|---|---|---|
| `iznik-server-go/test/message_test.go` | `TestPatchMessageAreanameOverride` | PATCH with areaname uses custom spelling in subject, not DB default |
| `iznik-server-go/test/message_test.go` | `TestPatchMessageAreanameOmittedKeepsDBDefault` | PATCH without areaname preserves DB-derived area (non-regression) |

## Discourse
https://discourse.ilovefreegle.org/t/9769/1

### DIFF
--- a/iznik-nuxt3/modtools/components/ModMessage.vue
+++ b/iznik-nuxt3/modtools/components/ModMessage.vue
@@ -49,6 +49,12 @@
               </div>
               <div v-if="editmessage.item && editmessage.location">
                 <b-input-group>
+                  <b-form-input
+                    v-model="editmessage.location.areaname"
+                    size="lg"
+                    class="me-1 areaname-input"
+                    placeholder="Neighbourhood"
+                  />
                   <PostCode
                     :value="editmessage.location.name"
                     :find="false"
@@ -1236,6 +1242,7 @@ async function save() {
         msgtype: editmessage.value.type,
         item: editmessage.value.item.name,
         location: editmessage.value.location.name,
+        areaname: editmessage.value.location.areaname,
         attachments: attids,
         textbody: editmessage.value.textbody,
       })
@@ -1444,6 +1451,10 @@ function spamReport() {
   /* Cap growth on wide desktop so the field doesn't stretch the whole row.
      flex-grow-1 keeps it filling available space up to this cap. */
   max-width: 500px;
+}
+
+.areaname-input {
+  max-width: 160px;
 }
 
 .fullsubject {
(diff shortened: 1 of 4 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #681) — this fix did not pass and is not ready to merge.

- **Partial implementation (symptom, not root cause)**: The bug is 'area name not preserved on save'. The diff adds an editable 'Neighbourhood' input bound to editmessage.location.areaname and sends it on save, but contains NO code to PRE-POPULATE editmessage.location.areaname from the existing message. The fact that the backend constructLocationString m
- **Test proves nothing about the real trigger path**: Neither the frontend nor backend tests exercise the bug's actual scenario (open an existing message that already has a custom area name, save, expect it preserved). The frontend tests inject location.areaname directly into mount props (location: { name:'E1 1AA', areaname:'Stepney' }), bypassing any 

PR #698, opened 2026-06-10
fix(chat): show 'Replied via email' label on digest/email reply messages

edwh:
## Summary

- Chat messages sent via email reply (including digest replies) were not labeled as such.  
- Root cause: Go API `ChatMessage` struct lacked the `platform` field (0=email, 1=website), so the frontend could never identify email-origin messages.
- The `ChatMessageDateRead.vue` component referenced `bymailid` in its v-if but no label text existed for it anywhere.

## Changes

**`iznik-server-go/chat/chatmessage.go`**
- Add `Platform int` field to `ChatMessage` struct — already selected by `chat_messages.*` query, now returned in API response
- Set `Platform = 1` in `CreateChatMessage` and `CreateChatMessageLoveJunk` (website/app origin) to prevent GORM from inserting zero-value for these paths
- `PostDigestReply` (AMP email reply) intentionally omits Platform, so Go's zero value correctly inserts `platform=0`

**`iznik-nuxt3/components/ChatMessageDateRead.vue`**
- Add `|| chatmessage?.platform === 0` to outer v-if so email-reply messages keep the date row visible even when `sameasnext=true`
- Add `<b-badge variant="info">Replied via email</b-badge>` shown when `platform === 0` and message is from the other user

**Tests**
- `ChatMessageDateRead.spec.js`: 3 new tests — label shows for platform=0, absent for platform=1, date section visible when sameasnext=true+platform=0
- `chat_test.go`: new `TestChatMessagePlatformField` verifying JSON serialization of platform=0 and platform=1

## Test plan

- [x] All 31 `ChatMessageDateRead.spec.js` vitest tests pass  
- [x] All 10,674 component unit tests pass  
- [x] Go code compiles and `go vet ./chat/... ./amp/...` clean  
- [ ] Go integration tests (run via status API — in progress on main checkout)

Fixes [topic 9775, post 1](https://discourse.ilovefreegle.org/t/9775/1).

### DIFF
--- a/iznik-nuxt3/components/ChatMessageDateRead.vue
+++ b/iznik-nuxt3/components/ChatMessageDateRead.vue
@@ -4,7 +4,8 @@
       !chatmessage?.sameasnext ||
       last ||
       chatmessage?.bymailid ||
-      chatmessage?.gap
+      chatmessage?.gap ||
+      chatmessage?.platform === 0
     "
     class="text-muted fontsize mb-1"
   >
@@ -15,6 +16,13 @@
       >
         {{ timeadaptChat(chatmessage?.date) }}
       </span>
+      <b-badge
+        v-if="chatmessage?.platform === 0"
+        variant="info"
+        class="ms-1"
+      >
+        Replied via email
+      </b-badge>
       <b-badge
         v-if="chatmessage?.replyexpected && !chatmessage?.replyreceived"
         variant="danger"
--- a/iznik-nuxt3/modtools/components/ModMessageWorry.vue
+++ b/iznik-nuxt3/modtools/components/ModMessageWorry.vue
@@ -48,12 +48,12 @@
         to describe the item more specifically.
       </span>
       <span v-else-if="reason.check === 'PhoneNumber'">
-        <strong>Phone number:</strong> This group restricts personal info in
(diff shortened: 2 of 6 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #698) — this fix did not pass and is not ready to merge.

- **Incomplete diff / partial implementation**: The diff adds `Platform int` to the `ChatMessage` struct in chatmessage.go but the comment immediately below the edited struct ('We need a separate struct for the query so that we can return image info in a single query') reveals there is a SEPARATE query struct used on the GET/read path. That query
- **No code path sets platform=0 for email replies**: The frontend treats `chatmessage?.platform === 0` as 'email reply', but the diff only ever explicitly sets Platform=1 (website/LoveJunk creation). Nothing in the diff sets Platform=0 for an actual email-originated reply. The feature relies entirely on the int zero-value default coinciding with 'emai
- **Test proves nothing**: TestChatMessagePlatformField (chat_test.go) only json.Marshal/Unmarshal's a struct literal with Platform set to 1 and 0 and asserts the round-trip preserves the int. This is a tautology — it would pass for ANY int field on ANY struct and tests Go's encoding/json, not this feature. It does NOT assert

PR #699, opened 2026-06-10
test(modtools-mapping): add regression test for adjacent-vertex deletion on midpoint save (#9770 post 2)

edwh:
## Root Cause

`UpdateLocation` applied `ST_Simplify(polygon, 0.001)` to every polygon before saving. This Douglas-Peucker pass (threshold ≈ 111 m) can drop **original** vertices adjacent to a new midpoint vertex, not just the midpoint itself.

**Post 2 scenario:** User drags a midpoint far enough that M survives ST_Simplify (> 0.001° from the original edge). However, the **original** adjacent vertex B is now only 0.00098° from line M→C — inside the tolerance — so ST_Simplify silently drops B. The user sees their edited boundary shrink at the corners.

The write-time ST_Simplify was already removed in b9b32a2b4 (PR #675, merged 2026-06-09), which fixed both post 1 and post 2. That fix has a test for post 1 (`TestUpdateLocationPreservesNearbyVertex`); this PR adds the explicit post 2 regression guard.

## Evidence

Geometry proof (coordinate distances in degrees):
- M = (-3.1975, 55.9415) is **0.0015°** north of original edge A-B (> 0.001 → M survives ST_Simplify)
- B = (-3.195, 55.940) distance to line M→C = **0.00098°** (< 0.001 → B dropped on buggy code)
- On buggy code: `ST_NumPoints = 6` (B silently discarded)
- On fixed code: `ST_NumPoints = 7` (all 6 distinct vertices preserved)

## Fix

No production code change — the fix is already on master (b9b32a2b4). This PR adds `TestUpdateLocationPreservesAdjacentVertexOnMidpointDrag` to guard the adjacent-vertex scenario explicitly.

## Other Call Sites

`ST_Simplify` in the `GET /locations` SELECT queries (lines 478–480) is display-only and intentionally left unchanged. No other write-path uses.

## Test

**File:** `iznik-server-go/test/location_test.go`
**Command:** `go test ./test/... -run TestUpdateLocationPreservesAdjacentVertexOnMidpointDrag -v`
**Proves:** fail-before (buggy code returns 6 points, assertion for 7 fails) / pass-after (fixed code returns 7 points)

## Discourse

https://discourse.ilovefreegle.org/t/9770/2

### DIFF
--- a/iznik-server-go/test/location_test.go
+++ b/iznik-server-go/test/location_test.go
@@ -221,6 +221,48 @@ func TestUpdateLocationPreservesNearbyVertex(t *testing.T) {
 	db.Exec("DELETE FROM locations WHERE id = ?", locID)
 }
 
+// TestUpdateLocationPreservesAdjacentVertexOnMidpointDrag tests Discourse #9770 post 2:
+// when a new midpoint vertex M is far enough from the original edge to survive
+// ST_Simplify(0.001), the ORIGINAL adjacent vertex B must also survive.
+// After inserting M between A and B, B is only 0.00098° from line M→C — inside the old
+// 0.001° tolerance — so ST_Simplify would silently drop B (the "adjacent marker deleted"
+// report).  The fix removes write-time ST_Simplify entirely, so B is preserved.
+func TestUpdateLocationPreservesAdjacentVertexOnMidpointDrag(t *testing.T) {
+	prefix := uniquePrefix("locwr_adjvert")
+	adminID := CreateTestUser(t, prefix+"_admin", "Admin")
+	_, adminToken := CreateTestSession(t, adminID)
+
+	db := database.DBConn
+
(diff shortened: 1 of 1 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #699) — this fix did not pass and is not ready to merge.

- **Incomplete diff**: The PR's own description (the test docstring, lines beginning 'The fix removes write-time ST_Simplify entirely, so B is preserved') claims a production behaviour change, but this diff modifies ONLY iznik-server-go/test/location_test.go. No production code path that called ST_Simplify on write is tou
- **Test proves nothing (reproduction ordering)**: Because the production change is not present in this diff, there is zero evidence here that the test was RED before the fix and GREEN after. The entire causal premise — B at 0.00098° from line M→C, inside the old 0.001 tolerance, so ST_Simplify drops B and returns 6 — is asserted only in comments an

PR #701, opened 2026-06-10
fix(chat): append digest-reply label when email reply quotes a digest

edwh:
## Problem

Chat messages sent by users who replied to the digest email no longer display the 'replied to digest' label. This was present in V1 but regressed in the V2 migration.

Reported by @Jos in [topic 9775, post 1](https://discourse.ilovefreegle.org/t/9775/1).

## Root cause

V1 `MailRouter.php` (lines 704–716) detected when a user's email reply included quoted digest content — either `-----Original Message-----` or `On ... -auto@GROUP_DOMAIN> wrote:` — and appended `(Probably replied to digest - check View original email)` to the chat message body before storing it.

V2 `IncomingMailService::createChatMessageFromEmail()` calls `StripQuotedService::strip()` to remove quoted content, but never added this label. The stripped body was stored without any indication it was a digest reply.

## Fix

In `createChatMessageFromEmail()`, detect the two digest-reply patterns **before** stripping (so we can still see them in the raw body), then **after** stripping append the label if either pattern was found.

The `$skipStripQuoted` guard (used for TN "Reporting…" messages which contain a conversation transcript as useful context) correctly bypasses both detection and appending.

## Tests

Two new regression tests in `IncomingMailServiceTest`:
- `test_digest_reply_via_original_message_separator_appends_label` — covers the `-----Original Message-----` pattern
- `test_digest_reply_via_auto_address_header_appends_label` — covers the `On ... -auto@groups.ilovefreegle.org> wrote:` pattern

All 153 `IncomingMailServiceTest` tests pass.

## Files changed

- `iznik-batch/app/Services/Mail/Incoming/IncomingMailService.php` — detect digest patterns before strip, append label after strip
- `iznik-batch/tests/Unit/Services/Mail/Incoming/IncomingMailServiceTest.php` — two new regression tests

### DIFF
--- a/iznik-batch/app/Services/Mail/Incoming/IncomingMailService.php
+++ b/iznik-batch/app/Services/Mail/Incoming/IncomingMailService.php
@@ -1881,12 +1881,30 @@ class IncomingMailService
             $body = $prependSubject . "\r\n\r\n" . $body;
         }
 
+        // Detect digest-reply patterns before stripping so we can append the label after.
+        // V1 parity: MailRouter.php detected "On ... -auto@GROUP_DOMAIN> wrote:" and
+        // "-----Original Message-----" to identify replies that include the full digest,
+        // then stripped the quoted content and appended the label text.
+        $isDigestReply = false;
+        if (! $skipStripQuoted) {
+            $groupDomain = preg_quote(config('freegle.mail.group_domain', 'groups.ilovefreegle.org'), '/');
+            if (preg_match('/^\s*On.*?-auto@' . $groupDomain . '>\s*wrote\s*:/ms', $body) ||
+                preg_match('/-----Original Message-----/', $body)) {
+                $isDigestReply = true;
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


edwh:
Adversarial review (Opus): **passed** — genuine fail-before/pass-after tests, correct detect-before-strip ordering, clean security. Two warnings for the human reviewer:

1. **Over-broad matcher**: the `-----Original Message-----` branch isn't digest-scoped — any Outlook-quoted reply (e.g. to a direct chat notification) would get the digest label. Consider requiring the `-auto@<group_domain>` marker for that branch too, or verify V1 was genuinely this broad.
2. **No negative test**: nothing asserts a non-digest reply does NOT get the label, so a label-everything regression would pass the current tests. An `assertStringNotContainsString` case (plain reply + volunteer $skipStripQuoted path) would close that.

Minor: verify the label text matches V1 verbatim; the `On…wrote:` regex spans newlines (s+m with lazy dot).

PR #702, opened 2026-06-10
fix(modtools): show notice for Pending posts with no content issues (group moderation)

edwh:
## Summary

- Posts on fully-moderated groups land in Pending with `contentcheck_checked_at` set but `contentcheck_reasons` null — the post has no content issues, it's just held by the group's moderation settings.
- Previously `ModMessageWorry.vue` showed nothing for these posts, leaving mods asking "should I leave or approve this?" (Discourse #9481 post 567, reporter: @Matty).
- Adds an informational notice: "This post has no content issues. It is in the pending queue because this group's settings require all posts to be reviewed manually."

## Root cause

`ContentCheckService::processUnprocessed` keeps a post in Pending when `$isModerated = true` (set when either `isGroupModerated()` or `isUserModerated()` is true). For fully-moderated groups, `isGroupModerated()` = true even for DEFAULT members. The batch job sets `contentcheck_checked_at` but leaves `contentcheck_reasons` null — no content flag, just group policy. The UI had no code to explain this case.

## Fix

`ModMessageWorry.vue` gets a new `pendingNoIssues` computed that fires when a message group has `collection === 'Pending'` AND `contentcheck_checked_at` is set AND `contentcheck_reasons` is null/empty. The template renders an `info`-variant `NoticeMessage` in that case.

## Test plan
- [x] New Vitest unit tests in `ModMessageWorry.spec.js` covering: shows notice when Pending + checked + no issues; no notice when not yet checked; no notice when issues exist; no notice when not Pending
- [x] All 45 tests pass

Fixes [topic 9481, post 567](https://discourse.ilovefreegle.org/t/9481/567) — original report on Discourse.

### DIFF
--- a/iznik-nuxt3/modtools/components/ModMessageWorry.vue
+++ b/iznik-nuxt3/modtools/components/ModMessageWorry.vue
@@ -1,5 +1,15 @@
 <template>
   <div v-if="message">
+    <!-- Group moderation: post is checked and clean but held by group settings -->
+    <NoticeMessage
+      v-if="pendingNoIssues"
+      variant="info"
+      class="mb-1"
+    >
+      This post has no content issues. It is in the pending queue because this
+      group's settings require all posts to be reviewed manually.
+    </NoticeMessage>
+
     <!-- Worry words flagged by the Go API (real-time, per-request check) -->
     <NoticeMessage
       v-for="(match, i) in worryMatches"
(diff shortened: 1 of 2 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #702) — this fix did not pass and is not ready to merge.

- **Contradictory UI / partial implementation**: The `pendingNoIssues` computed only inspects `contentcheck_reasons` (the batch job) and ignores `worryMatches` (the Go real-time worry-word check that is rendered immediately below it in the same template). A Pending post can have worry words flagged in real time while the batch content-check record
- **Incorrect root-cause attribution**: The notice asserts the post is pending 'because this group's settings require all posts to be reviewed manually,' but `pendingNoIssues` never checks any group-moderation setting (or the member's posting status). Per the moderation pipeline (all posts insert Pending; contentcheck keeps NULL/MODERATED

PR #703, opened 2026-06-10
fix(modtools): show info notice for Pending posts with no content issues

edwh:
## Root Cause

`ModMessage.vue` only rendered `ModMessageWorry` when `message.worry?.length` or `contentcheck_reasons` were non-empty. For posts on fully-moderated groups, the batch content-check sets `contentcheck_checked_at` but leaves `contentcheck_reasons` null (no flags) — the post is Pending purely because the group requires manual approval. The component was never rendered for this case, so mods saw an empty worries area and no explanation, causing the "should I leave or approve this?" confusion.

Previous attempt PR #702 added a `pendingNoIssues` notice but its computed only checked `contentcheck_reasons`, ignoring `worryMatches` — a Pending post with worry-word flags but no batch reasons would have incorrectly shown "no content issues."

## Evidence

```
 FAIL  tests/unit/components/modtools/ModMessage.spec.js > ModMessage > ModMessageWorry > shows worry component for Pending post with contentcheck_checked_at set (no worry, no reasons)
 FAIL  tests/unit/components/modtools/ModMessageWorry.spec.js > ModMessageWorry > pendingNoIssues — info notice for Pending posts with no flags > shows info notice for Pending + checked + no worry + no reasons
 FAIL  tests/unit/components/modtools/ModMessageWorry.spec.js > ModMessageWorry > pendingNoIssues — info notice for Pending posts with no flags > info notice text mentions pending queue
 Tests  4 failed | 43 passed (47)
```
After fix: all 47 (ModMessageWorry) + 77 (ModMessage) tests pass.

## Fix

1. **`ModMessage.vue`**: expanded the `v-if` guard on `<ModMessageWorry>` to include a third condition — render also when any group has `collection === 'Pending'` AND `contentcheck_checked_at` set.
2. **`ModMessageWorry.vue`**: added `pendingNoIssues` computed that fires only when `contentcheck_checked_at` is set, collection is Pending, **and** both `worryMatches` and `contentcheckReasons` are empty — ensuring no false "no issues" notice appears when real flags exist.

## Other Call Sites

`ModMessageWorry` is rendered only from `ModMessage.vue` (single call site). No other files use this pattern.

## Test

| File | Command | Proves |
|---|---|---|
| `tests/unit/components/modtools/ModMessageWorry.spec.js` | `vitest run ModMessageWorry` | pendingNoIssues notice shown/suppressed correctly (47 tests pass) |
| `tests/unit/components/modtools/ModMessage.spec.js` | `vitest run ModMessage` | ModMessageWorry rendered for Pending+checked posts (77 tests pass) |

Fail-before: 4 tests fail on unmodified code. Pass-after: all pass with fix.

## Discourse

https://discourse.ilovefreegle.org/t/9481/567

### DIFF
--- a/iznik-nuxt3/modtools/components/ModMessage.vue
+++ b/iznik-nuxt3/modtools/components/ModMessage.vue
@@ -356,6 +356,9 @@
                   (g) =>
                     g.contentcheck_reasons &&
                     g.contentcheck_reasons.length
+                ) ||
+                message.groups?.some(
+                  (g) => g.collection === 'Pending' && g.contentcheck_checked_at
                 )
               "
               :messageid="message.id"
--- a/iznik-nuxt3/modtools/components/ModMessageWorry.vue
+++ b/iznik-nuxt3/modtools/components/ModMessageWorry.vue
@@ -36,6 +36,16 @@
       </p>
     </NoticeMessage>
 
+    <!-- Pending post with no content flags: explain it's held by manual-review group setting -->
+    <NoticeMessage
+      v-if="pendingNoIssues"
+      variant="info"
+      class="mb-1"
+    >
+      This post has no content issues. It is in the pending queue because this
+      community requires all posts to be approved manually.
+    </NoticeMessage>
+
     <!-- Content check failure reasons (stored, from batch processing) -->
     <NoticeMessage
       v-for="(reason, i) in contentcheckReasons"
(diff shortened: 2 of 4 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #703) — this fix did not pass and is not ready to merge.

- **Partial implementation / unverified root-cause claim**: The new NoticeMessage asserts a definitive cause: "It is in the pending queue because this community requires all posts to be approved manually." But `pendingNoIssues` never inspects the group's moderation setting or the poster's moderation status — it only checks `collection === 'Pending' && conten

PR #704, opened 2026-06-10
fix(discourse-badge): count only topics active in last 30 days (newly-promoted mod flood)

edwh:
## Root cause

A newly-promoted moderator added to the Discourse group categories (Hackney, Newham, …) sees **every existing forum topic they have never read** counted as "new"/"unread". The green Discourse badge in ModTools (`discourseCount` in `modtools/layouts/default.vue`) sums `notifications + newtopics + unreadtopics`, so the badge floods into the hundreds (466 + 793 in the reported case — Discourse [topic 9654 post 10](https://discourse.ilovefreegle.org/t/9654/10)) while the red work badge stays at 0.

V1 PHP used the same three-field formula but was masked by short HTTP timeouts that often returned 0 for `newtopics` plus a 5-minute session cache; V2 Go reads the lists reliably, so the latent flood became visible.

## Approach

The badge's **meaning is unchanged** — it still counts new/unread forum topics, not just notifications — but only topics with **activity in the last 30 days**. A one-time backlog of years-old unread topics no longer counts; genuinely active threads still do.

The filter lives in the **Go API** (`iznik-server-go/session/session.go`), which already fetches the full `/new.json` and `/unread.json` topic lists. It's the **same two API calls** plus a per-topic timestamp check — no extra requests, no measurable cost.

## Changes

- **`iznik-server-go/session/session.go`**
  - `TopicActiveWithin(createdAt, bumpedAt, lastPosted, since)` — pure, unit-tested predicate. Uses the most-recent-activity timestamp (`bumped_at` → `last_posted_at` → `created_at`); a missing/malformed timestamp counts as inactive.
  - `discourseTopicWindowDays = 30` — named constant documenting the newly-promoted-mod flood.
  - The `/new.json` and `/unread.json` goroutines now count only topics passing the predicate.
- **`iznik-nuxt3/modtools/layouts/default.vue`** — `discourseCount` reverted to the original `notifications + newtopics + unreadtopics` sum (the meaning-changing notification-only version is withdrawn).
- **Tests** — `TestTopicActiveWithin` (recent `bumped_at` → counts; everything old → not; recent `created_at` only → counts; missing/malformed → not). Frontend spec reverted to master.

## Test plan

- [ ] Go suite green (incl. `TestTopicActiveWithin`) via status API / CI.
- [ ] Frontend unit suite green (reverted to master behaviour).

## Discourse
https://discourse.ilovefreegle.org/t/9654/10

### DIFF
--- a/iznik-nuxt3/modtools/layouts/default.vue
+++ b/iznik-nuxt3/modtools/layouts/default.vue
@@ -374,9 +374,7 @@ const inMTapp = ref(false)
 
 const discourseCount = computed(() => {
   const discourse = authStore.discourse
-  return discourse
-    ? discourse.notifications + discourse.newtopics + discourse.unreadtopics
-    : 0
+  return discourse ? discourse.notifications : 0
 })
 
 const menuCount = computed(() => {
--- a/iznik-nuxt3/modtools/layouts/default.vue
+++ b/iznik-nuxt3/modtools/layouts/default.vue
@@ -374,7 +374,9 @@ const inMTapp = ref(false)
 
 const discourseCount = computed(() => {
   const discourse = authStore.discourse
-  return discourse ? discourse.notifications : 0
+  return discourse
+    ? discourse.notifications + discourse.newtopics + discourse.unreadtopics
+    : 0
 })
 
 const menuCount = computed(() => {
--- a/iznik-nuxt3/tests/unit/components/modtools/ModtoolsDefaultLayout.spec.js
+++ b/iznik-nuxt3/tests/unit/components/modtools/ModtoolsDefaultLayout.spec.js
@@ -279,43 +279,6 @@ describe('modtools default layout — pending badge includes spam', () => {
   })
 })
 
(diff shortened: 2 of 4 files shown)


edwh:
Adversarial review (Opus): **passed** — genuine before/after reproduction (old formula renders 1262 for the reported 3+466+793 case; test asserts the exact rendered count), well-scoped diff, no security/test regressions.

Two warnings for the human reviewer:
1. **Broader than the bug**: dropping `newtopics`+`unreadtopics` permanently changes what the green badge *means* (notifications-only vs forum-activity), and the underlying anomaly (a newly-promoted mod seeing every existing topic as 'new') is left unaddressed server-side. Confirm notifications-only is the intended product behaviour — an alternative is marking all topics read on first Discourse sync.
2. **Sibling aggregations**: verify nothing else (e.g. `menuCount`) independently sums the same discourse fields and reproduces the inflation.

Pipeline note: this PR was opened by a delegate that hit its hard cap before reporting, so the link/review gates ran retroactively — bug 9654/10 is now bound and the Discourse link added.

PR #711, opened 2026-06-11
fix(sso): authenticate Discourse SSO cookie by id+token (series now numeric → redirect loop)

edwh:
## Root cause

PR #679 (merged 2026-06-09) changed `session.go` to emit the persistent token's `series` field as a `uint64` JSON number in the `/session` response. The Discourse SSO cookie (`Iznik-Discourse-SSO`) is built from that persistent token by ModTools.

`discourse.go`'s `validateDiscourseSession` parsed the cookie into:

```go
var cookie struct {
    ID     uint64 `json:"id"`
    Series string `json:"series"`   // ← string field
    Token  string `json:"token"`
}
```

When `json.Unmarshal` encountered `"series": 12345` (a JSON number) against a `string` field, it returned a type error. `validateDiscourseSession` returned `"invalid cookie JSON"` → `DiscourseSSO` redirected back to `https://modtools.org/discourse` → ModTools re-set the cookie and retried → **infinite redirect loop**, all moderators locked out of Discourse login.

CI missed it because `discourse_test.go` always mocked `series` as a string literal (`"s"`, `"abc"`), not the numeric format that the real session emitter now produces.

## Fix

Drop `Series` from the cookie parse struct entirely. Authenticate using **id+token only**, matching the pattern already established in `auth/auth.go WhoAmI` after the same PR #679. Token is a cryptographically random secret, so id+token is a sufficient credential without series.

Also remove `AND sessions.series = ?` from the DB session lookup query (series is no longer read from the cookie).

```go
// Series is deliberately absent: it changed from a JSON string to a
// JSON number in PR #679. Parsing id+token is sufficient.
var cookie struct {
    ID    uint64 `json:"id"`
    Token string `json:"token"`
}
```

```sql
-- Before (broken):
AND sessions.id = ? AND sessions.series = ? AND sessions.token = ?
-- After (fixed):
AND sessions.id = ? AND sessions.token = ?
```

## Regression test

`TestValidateDiscourseSession_NumericSeries_PassesPreDBCheck` in `discourse_handler_test.go`:

- Feeds a cookie with `"series": 12345` (JSON number, as PR #679 now produces)
- On **old code**: `json.Unmarshal` errors → `"invalid cookie JSON"` returned → test would assert this is NOT the result and fail
- On **fixed code**: `Unmarshal` succeeds (unknown fields are ignored) → passes the pre-DB check → only fails at the DB lookup (no DB in unit tests), which is correct behaviour

Two companion tests verify the pre-DB completeness check still works for id=0 and empty token even when series is a number.

## Impact

Unblocks all Freegle moderators from logging in to Discourse via ModTools SSO. This is a regression introduced by #679.

### DIFF
--- a/iznik-server-go/sso/discourse.go
+++ b/iznik-server-go/sso/discourse.go
@@ -99,20 +99,29 @@ func DiscourseSSO(c *fiber.Ctx) error {
 
 // validateDiscourseSession validates the cookie against the sessions table.
 // The user must be a Freegle moderator.
+//
+// Authentication uses id+token only. The series field is intentionally ignored:
+// PR #679 (2026-06-09) changed the session emitter to output series as a JSON
+// number (uint64) rather than a string, which caused json.Unmarshal to error
+// when trying to decode a number into a string field — producing an infinite
+// ModTools ⇄ Discourse redirect loop for all moderators. Authenticating by
+// id+token alone (matching the approach in auth/auth.go WhoAmI) fixes the loop
+// and is sufficient because token is a cryptographically random secret.
 func validateDiscourseSession(cookieValue string) (*ssoSession, error) {
 	db := database.DBConn
 
+	// Series is deliberately absent from this struct: the field changed from a
+	// JSON string to a JSON number in PR #679. Parsing id+token is sufficient.
 	var cookie struct {
-		ID     uint64 `json:"id"`
-		Series string `json:"series"`
-		Token  string `json:"token"`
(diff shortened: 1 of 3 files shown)


edwh:
Pushed directly to master as defaaef96 (production-breaking ModTools⇄Discourse redirect loop — fix applied on master per maintainer request). Same change: SSO cookie authenticated by id+token (series-agnostic), with the numeric-series regression refactored to a pure parseSSOCookie test. Full Go suite green (3061).

PR #716, opened 2026-06-11
fix(modtools): show action buttons for Spam-collection messages in Pending view (#9518)

edwh:
## Root Cause

PR #638 (2026-06-04) made the server-side Pending listing return both `Pending` AND `Spam`-collection messages. PR #709 (2026-06-11 03:44) added `'Spam'` to the client-side allowed-collection filter so those messages are now visible. However `ModMessageButtons.vue` only had `v-if`/`v-else-if` branches for `pending` (`hasCollection('Pending')`) and `approved` (`hasCollection('Approved')`). A `Spam`-collection message shown in the Pending view hit none of those conditions, so **no action buttons were rendered**.

## Evidence

Discourse #9518 post 330 — Derek: *"A couple of 'historic' posts have arrived in Pending overnight but I have no way to process them as there are no buttons. First one is 20 days old, second one is 5 years old."*

Filed 2026-06-11 06:13 — approximately 2.5 hours after PR #709 merged at 03:44, confirming the Spam-in-Pending visibility was the trigger.

## Fix

**`modtools/components/ModMessageButtons.vue`**
- Added `const spam = computed(() => hasCollection('Spam'))`
- Added `v-else-if="spam"` section with **Approve** (false positive), **Reject**, and **Delete as Spam** buttons — the standard triage set for a message auto-classified as spam
- Extended `validActions` to include the spam case so configured standard messages (e.g. rejection templates) are also available

## Other Call Sites

`ModMessage.vue` passes the message through to `ModMessageButtons`; no additional changes required there.

## Test

New Vitest tests in `ModMessageButtons.spec.js`:
- `spam message buttons > shows approve button for spam messages appearing in pending view` ✓
- `spam message buttons > shows reject button for spam messages appearing in pending view` ✓
- `spam message buttons > shows Delete as Spam button for spam messages appearing in pending view` ✓
- `spam message buttons > hides approve button for spam messages when cantpost is true` ✓
- `spam message buttons > spam computed is true when message has Spam collection` ✓
- `spam message buttons > spam computed is false when message has Pending collection` ✓

Full suite: 49 tests passed (49).

## Discourse

Fixes Discourse #9518 / post 330

### DIFF
--- a/iznik-nuxt3/modtools/components/ModMessageButtons.vue
+++ b/iznik-nuxt3/modtools/components/ModMessageButtons.vue
@@ -135,6 +135,33 @@
         @handle="outcome($event, 'Withdrawn')"
       />
     </div>
+    <div v-else-if="spam" class="d-inline">
+      <ModMessageButton
+        v-if="!cantpost"
+        :messageid="message.id"
+        :groupid="groupid"
+        variant="primary"
+        icon="check"
+        approve
+        label="Approve"
+      />
+      <ModMessageButton
+        :messageid="message.id"
+        :groupid="groupid"
+        variant="warning"
+        icon="times"
+        reject
+        label="Reject"
+      />
+      <ModMessageButton
+        :messageid="message.id"
+        :groupid="groupid"
+        variant="danger"
+        icon="ban"
+        spam
+        label="Delete as Spam"
+      />
+    </div>
     <div v-if="!editreview" class="d-lg-inline">
       <ModMessageButton
         v-for="stdmsg in filtered"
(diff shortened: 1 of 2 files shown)


edwh:
Superseded — the fix has landed directly on master (commit 3cbea9eab) as a more complete change.

That commit includes everything here (action buttons for Spam-collection messages in the Pending queue) **plus** the two related fixes this PR didn't cover:
- **Go API:** the Pending review list now excludes Spam-collection messages older than 30 days, so stale spam ages out of the queue instead of accumulating.
- **Auto-approve:** `AutoApproveService` will never auto-approve a message that is in the Spam collection on any group (spam must be actioned by a human, not auto-sent by the 48h fallback).

Frontend behaviour matches this PR (Approve/Reject/Delete-as-Spam) and additionally exposes Delete/Hold by routing Spam through the existing Pending action block. Tests carried over too. Thanks!

PR #720, opened 2026-06-11
fix(modtools): reword misleading personal-info pending warning

edwh:
## Summary

- The `PhoneNumber` and `EmailAddress` contentcheck warnings in `ModMessageWorry.vue` said *"This group restricts personal info in posts"*, implying an automated enforcement that does not happen — the check only **flags** the post for manual review (`action: flag`).
- The `PhoneNumber` check also runs **regardless** of the group's `restrictpersonalinfo` setting, making the group attribution doubly wrong.
- Reworded both messages to describe **what the post contains** rather than implying an automated restriction. Per reporter suggestion (Discourse #9766, post 8).

## Code Quality Review

- Wording-only change to two moderator-facing warning strings; no logic/behaviour change.
- Spec updated to assert the new wording.

## Test Plan

- `ModMessageWorry.spec.js` updated and asserts the reworded PhoneNumber/EmailAddress messages.
- Merges cleanly onto current master (the touched files are unchanged on master since the commit's base).

### DIFF
--- a/iznik-nuxt3/modtools/components/ModMessageWorry.vue
+++ b/iznik-nuxt3/modtools/components/ModMessageWorry.vue
@@ -48,12 +48,12 @@
         to describe the item more specifically.
       </span>
       <span v-else-if="reason.check === 'PhoneNumber'">
-        <strong>Phone number:</strong> This group restricts personal info in
-        posts. Please ask the member to remove their phone number.
+        <strong>Phone number:</strong> This post contains a phone number.
+        Please ask the member to remove it.
       </span>
       <span v-else-if="reason.check === 'EmailAddress'">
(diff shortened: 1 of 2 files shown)


edwh:
Superseded by #677 (`be86c9532`, *fix(contentcheck): gate phone-number check on group restrictpersonalinfo (#9766)*), which is already merged to master.

#677 makes `checkPhoneNumbers` respect the group's `restrictpersonalinfo` rule (the real bug — the check previously ran regardless) **and** adds test coverage in `ContentCheckServiceTest` + `ContentCheckTest` asserting it flags only when the rule is set. With that gating in place, the existing "This group restricts personal info in posts" wording is accurate, so the wording change here is no longer needed.

PR #733, opened 2026-06-13
feat(api): server-side kill switch for out-of-date builds (webversion-age) [BLOCKED]

edwh:
## ⛔ BLOCKED — do not merge / do not arm yet

This is the **enforcement half** of forcing app upgrades (Sentry CAPACITOR-44Z context). It ships **inert** and must not be armed until the prerequisites below are met.

## What it does
`webversion` (the `BUILD_DATE` every client sends on `GET /session`) is compared against a configurable `app_min_webversion` threshold. When the client's build is **older**, the session is failed with `{ "ret": 123, "status": "App is out of date" }`, making an out-of-date client non-functional until it updates. No client cooperation needed.

**Gate = webversion-age only (Option A).** There is no server-visible app/web signal on `GET /session` (modern apps send only `webversion`, same as the website), so a date-only gate is used. Website bundles are always fresh and self-heal on reload, so they aren't meaningfully affected; the companion web auto-reload (#729) handles stale web tabs.

## How
- **Go (V2)** — `session.GetSession` reads `app_min_webversion` and rejects older builds. The ISO-8601 comparison is `webversionOlderThan()` in `session/appversion.go` (pure, unit-tested).
- **PHP (V1)** — `session.php` mirrors it via `Utils::webversionTooOld()` (parity).
- **Both FAIL OPEN**: an unset config key, a missing `webversion`, or a legacy non-ISO value never blocks anyone. With the config unset (the default), this PR blocks **nobody**.

## 🔒 Before this can be merged AND armed
1. A **post-#650 fixed app** is live in **both** the App Store and Play Store (so there's a fixed version to force users onto — allow for iOS review time).
2. Set config `app_min_webversion` = that build's `BUILD_DATE` (**B_DATE**). Setting it earlier would **lock users out**.
3. Raise `app_fd_version_ios_required` / `app_fd_version_android_required` from the current `3.2.5` to that build's `MOBILE_VERSION` (**B_VER**) — the non-blocking in-app nudge (`stores/mobile.js` `versionOutOfDate`). These are config **data** changes, set post-publish; **not in this PR**. Each must be ≤ the version actually live in its store.

## Tests
- `iznik-server-go/session/appversion_pure_test.go` — Go threshold/fail-open cases (no DB).
- `UtilsTest::testWebversionTooOld` — PHP parity cases.

### DIFF
--- a/iznik-server-go/session/appversion.go
+++ b/iznik-server-go/session/appversion.go
@@ -1,0 +1,31 @@
+package session
+
+import "time"
+
+// webversionOlderThan reports whether an ISO-8601 webversion (the BUILD_DATE that
+// every client sends on GET /session) is strictly older than minVersion, the
+// configured app_min_webversion threshold.
+//
+// It is the comparison behind the server-side "app is out of date" kill switch.
+// It fails open (returns false) whenever either value is empty or is not a
+// parseable RFC3339/ISO-8601 timestamp:
+//   - the threshold is empty until an operator arms it, so an unset config blocks
+//     nobody;
+//   - legacy clients send non-ISO webversions, which must never be blocked.
+func webversionOlderThan(webversion, minVersion string) bool {
+	if webversion == "" || minVersion == "" {
+		return false
+	}
+
+	clientTime, err := time.Parse(time.RFC3339, webversion)
+	if err != nil {
+		return false
+	}
+
+	thresholdTime, err := time.Parse(time.RFC3339, minVersion)
+	if err != nil {
+		return false
+	}
+
+	return clientTime.Before(thresholdTime)
+}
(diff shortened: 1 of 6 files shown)


edwh:
Folded into master (Go half) in commit 6b40fb21f, with a per-app split added: ModTools and the Freegle app/web are gated on separate config keys (`app_min_webversion_mt` vs `app_min_webversion`), selected by the `modtools` flag the client adds to every request - so each app can be forced to update independently (this PR used a single global threshold). webversionOlderThan + the fail-open behaviour are unchanged and unit-tested. V1 PHP parity was intentionally dropped (apps use apiv2/Go for the session). Still ships inert - blocks nobody until a config key is armed.

PR #747, opened 2026-06-14
fix(modtools/config): allow group mods to add standard messages to unprotected configs

edwh:
## Root Cause

`PostStdMsg` (Go API `/modtools/stdmsg` POST) checked `auth.IsSystemMod(myid)`, which only allows users with a system-level `Moderator`/`Support`/`Admin` systemrole. Regular group moderators have `systemrole = "User"` (they're mods at the group membership level, not the system level), so they received HTTP 403 when trying to add a new standard message button to a config.

`PatchStdMsg` and `DeleteStdMsg` correctly use `canModifyConfig(myid, configid)`, which allows the config creator or any user when the config is unprotected. `PostStdMsg` was inconsistent — it required system-level authority that regular group moderators never have.

## Evidence

`TestPostStdMsgByGroupMod` (with assert 200) **failed** on buggy code:

```
--- FAIL: TestPostStdMsgByGroupMod (0.05s)
    Error: Response status 403 is not equal to 200
FAIL	iznik-server-go/test	132.046s
```

After fix, both new tests pass along with all existing stdmsg tests.

## Fix

Removed the early `IsSystemMod` check from `PostStdMsg`. After parsing `req.Configid`, replaced it with `canModifyConfig(myid, req.Configid)` — the same check already used by PATCH and DELETE. The check runs after input parsing so the configid is available to look up permission.

## Other Call Sites

`auth.IsSystemMod` is used in 10+ other files (alert, admin, microvolunteering, etc.) where system-level access is genuinely required. The only incorrect usage was in `PostStdMsg`.

## Test

`iznik-server-go/test/stdmsg_test.go`

Command: `curl -X POST http://localhost:8081/api/tests/go`

- `TestPostStdMsgByGroupMod` — group mod (systemrole="User") can POST to unprotected config → 200 ✓
- `TestPostStdMsgByGroupModProtectedConfig` — group mod cannot POST to a protected config they didn't create → 403 ✓

Proves: fail-before (403 on unprotected), pass-after (200 on unprotected, 403 on protected).

## Discourse

https://discourse.ilovefreegle.org/t/9793/1

### DIFF
--- a/iznik-server-go/stdmsg/stdmsg.go
+++ b/iznik-server-go/stdmsg/stdmsg.go
@@ -89,10 +89,6 @@ func PostStdMsg(c *fiber.Ctx) error {
 		return c.Status(fiber.StatusUnauthorized).JSON(fiber.Map{"ret": 1, "status": "Not logged in"})
 	}
 
-	if !auth.IsSystemMod(myid) {
-		return c.Status(fiber.StatusForbidden).JSON(fiber.Map{"ret": 4, "status": "Don't have rights to create configs"})
-	}
-
 	type CreateRequest struct {
 		Configid     uint64  `json:"configid"`
 		Title        string  `json:"title"`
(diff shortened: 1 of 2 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #747) — this fix did not pass and is not ready to merge.

- **Security regression / test proves nothing**: This PR relaxes an authorization gate on a write path (the INSERT into mod_stdmsgs in PostStdMsg) from `auth.IsSystemMod(myid)` to `canModifyConfig(myid, req.Configid)`, but ships NO test proving the security lower bound — i.e. that a plain logged-in user with no moderator role and no relevant membe

PR #748, opened 2026-06-14
fix(pending/listings): reject Received/Taken outcomes on pending Wanted posts

edwh:
## Root Cause

The V2 Go API intentionally shows pending posts to their owner (unlike V1 which only showed Approved posts). The `handleOutcome` function in `iznik-server-go/message/message.go` already guarded `Withdrawn` on pending posts but had no equivalent guard for `Taken` or `Received`. This meant a user could visit "My Posts", see their still-pending WANTED, and click "Received" — which would immediately record an outcome and remove the post from listings before it had been approved by a moderator.

The `ChaseUpService::processIntendedOutcomes` PHP batch job (runs every minute, applies intended outcomes after 30 min) also had no pending check, so a race existed: record an intended outcome, then have the post go back to pending — and the batch job would apply it anyway.

## Evidence

**Before fix** — `TestPostMessageReceivedOnPendingWanted` fails:
```
FAIL: expected 400, got 200
FAIL: expected outcomeCount=0, got 1
```

**After fix** — test passes:
```
PASS: TestPostMessageReceivedOnPendingWanted (400 returned, outcomeCount=0)
Full suite: 3176 passed, 2 failed (both pre-existing, unrelated to this change)
```

## Fix

Three changes:

1. **`iznik-server-go/message/message.go` — `handleOutcome`** (primary guard): After the existing `Withdrawn`-on-pending guard (~line 3779), added an identical check for `Taken`/`Received` that returns HTTP 400 if `messages_groups` has any row with `collection = 'Pending'` for the message.

2. **`iznik-server-go/message/message.go` — `handleOutcomeIntended`** (~line 3695): Added the same pending check before recording an intended `Taken`/`Received` outcome, so the batch job never gets a chance to race.

3. **`iznik-batch/app/Services/ChaseUpService.php` — `processIntendedOutcomes`**: Defence-in-depth — before applying any `Taken`/`Received` intended outcome from `messages_outcomes_intended`, skip if the message is still in `Pending` collection. Handles the edge case where an intended was recorded before the post went pending.

All three guards mirror the same pattern already used for `Withdrawn` on pending posts.

## Other Call Sites

Grepped for other places that insert into `messages_outcomes` directly — all go through `handleOutcome`, which now has the guard. No other direct callers.

## Test

**File:** `iznik-server-go/test/message_test.go` — `TestPostMessageReceivedOnPendingWanted`

**Command:** `POST /api/message {action:"Outcome", outcome:"Received"}` as the post owner, where the post has `messages_groups.collection = 'Pending'`

**Proves:**
- Before fix: returns 200, `messages_outcomes` row created (bug)
- After fix: returns 400, no `messages_outcomes` row created

## Discourse
https://discourse.ilovefreegle.org/t/9788/7

### DIFF
--- a/iznik-batch/app/Services/ChaseUpService.php
+++ b/iznik-batch/app/Services/ChaseUpService.php
@@ -139,6 +139,20 @@ class ChaseUpService
                 continue;
             }
 
+            // Skip Taken/Received on pending posts: the post has not been approved yet,
+            // so recording a successful outcome would mark it as done prematurely.
+            // Discourse: https://discourse.ilovefreegle.org/t/9788/7
+            if (in_array($intended->outcome, ['Taken', 'Received'])) {
+                $isPending = DB::table('messages_groups')
+                    ->where('msgid', $intended->msgid)
+                    ->where('collection', MessageGroup::COLLECTION_PENDING)
+                    ->exists();
+
+                if ($isPending) {
+                    continue;
+                }
+            }
+
             if ($dryRun) {
                 Log::info("Dry run: would process intended outcome '{$intended->outcome}' for message #{$intended->msgid}");
                 $count++;
--- a/iznik-server-go/message/message.go
+++ b/iznik-server-go/message/message.go
@@ -3693,6 +3693,17 @@ func handleOutcomeIntended(c *fiber.Ctx, myid uint64, req PostMessageRequest) er
 		return fiber.NewError(fiber.StatusForbidden, "Not allowed to modify this message")
 	}
 
+	// Reject Taken/Received intended outcomes for pending posts: the batch job
+	// (ChaseUpService::processIntendedOutcomes) would otherwise apply the outcome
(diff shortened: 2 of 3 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #748) — this fix did not pass and is not ready to merge.

- **Coverage — root-cause path untested**: The fix touches THREE code paths but only one is tested. The Laravel change in ChaseUpService::processIntendedOutcomes (the pending-skip branch) has NO test — no PHPUnit test was added in this PR at all. This is the path the bug is actually about: the Go comment itself states the batch job 'would ot
- **Coverage — changed Go branch untested**: The new reject branch added to handleOutcomeIntended (message.go ~line 3696) has no test. The added test posts action 'Outcome' (handleOutcome) only; it never posts the OutcomeIntended action, so the handleOutcomeIntended guard — its own separate SELECT COUNT query and constant comparison — is entir

PR #749, opened 2026-06-14
fix(pending): auto-refresh list when ContentCheck blocks post to Spam (Discourse #9783)

edwh:
## Problem

Discourse #9783 post 6 — moderators see a red pending badge but an empty list when a post is held by a concern keyword.

**Root cause:** When ContentCheck's `processUnprocessed` runs a global concern keyword with `action='block'`, the message moves to the Spam collection. This increments `work.spam`, which the left-nav red badge counts (`pending + spam`). However, the pending page's `workType` was `['pending', 'pendingother']`, so `workdetail.total` excluded spam. The `watch(workdetail)` watcher's `JSON.stringify` comparison saw no change and skipped `getMessages`. Result: badge increases, list does not refresh.

(Note: per-group worry words use `action='flag'` and stay Pending — those already trigger correctly via `work.pending`. Only the `action='block'` path, which moves posts to Spam, was missing.)

## Fix

Two minimal changes in `[[term]].vue`:

1. `workType.value = ['pending', 'pendingother', 'spam']` — so `workdetail.total` rises when `work.spam` rises, firing `getMessages` to refresh the list.
2. `:work="['pending', 'pendingother', 'spam']"` on `ModGroupSelect` — so per-group spam counts appear in the group dropdown (cosmetic; `group.work.spam` is already populated by `groupWork.go`).

The Spam-collection messages are already included in the pending-page list response from the server (see `ListMessagesMT` / `collectionFilter`), and the client-side `allowed` array was fixed to include `'Spam'` in PR #709. This PR provides the missing auto-refresh trigger.

## Tests

Two new unit tests in `useModMessages.pendingAutoRefresh.spec.js`:

- **Documents the bug**: with `workType=['pending','pendingother']`, spam increase → `getMessages` NOT called.
- **Documents the fix**: with `workType=['pending','pendingother','spam']`, spam increase → `getMessages` IS called.

## Test plan

- [ ] CI vitest suite passes (verifies the two new unit tests)
- [ ] CI build-test-katapult passes (overall regression check)
- [ ] Manual: post a message matching a global concern keyword with `action='block'` → check that the pending-page list auto-refreshes when `work.spam` updates (within the session poll interval, ~30s)

### DIFF
--- a/iznik-nuxt3/modtools/pages/messages/pending/[[id]]/[[term]].vue
+++ b/iznik-nuxt3/modtools/pages/messages/pending/[[id]]/[[term]].vue
@@ -16,7 +16,7 @@
           v-model="groupid"
           all
           modonly
-          :work="['pending', 'pendingother']"
+          :work="['pending', 'pendingother', 'spam']"
           remember="pending"
           :url-override="urlOverride"
         />
@@ -87,7 +87,7 @@ const {
 
 summarykey.value = 'modtoolsMessagesPendingSummary'
 collection.value = 'Pending' // Pending also gets PendingOther
-workType.value = ['pending', 'pendingother']
+workType.value = ['pending', 'pendingother', 'spam']
 
 const { me, myGroups } = useMe()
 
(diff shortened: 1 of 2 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #749) — this fix did not pass and is not ready to merge.

- **Test proves nothing / not a reproduction test**: The fix is the config change in modtools/pages/messages/pending/[[id]]/[[term]].vue (workType and :work gain 'spam'), but both new tests are in the composable spec (useModMessages.pendingAutoRefresh.spec.js) and manually assign workType.value rather than importing it from the page. The 'calls getMes

PR #758, opened 2026-06-14
feat(app): share an image into Freegle to start an Offer (Android)

edwh:
## What

Adds an Android **"Share → Freegle"** target. From the gallery, camera, or any
other app's share sheet, the user can send a photo straight into a new **OFFER**
with the image already attached — no need to open the app and re-pick it.

This is the Android half of the *Share-into-Freegle* feature. The iOS Share
Extension follows in a separate PR (it needs an App Group + provisioning-profile
changes, so it's staged on its own to avoid touching production signing here).

## How

- **AndroidManifest** — `ACTION_SEND` / `ACTION_SEND_MULTIPLE` `image/*`
  intent-filters on `MainActivity`.
- **MainActivity.java** — on `onCreate`/`onNewIntent`, copies the shared
  image(s) from the intent-scoped `content://` Uri into the app cache (the Uri
  permission doesn't outlive the intent) and exposes them to the WebView via a
  tiny `window.FreegleShare` JS bridge (`consume()` returns the file paths and
  clears the queue).
- **stores/mobile.js** — `initShareIntent` pulls shared images at startup (cold
  share) and on `resume` (warm share), converts the cache paths with
  `Capacitor.convertFileSrc`, queues them in `pendingSharedImages`, and routes
  to `/give/mobile/photos`.
- **PhotoUploader.vue** — `defineExpose({ processPhoto })` so a shared photo
  runs the *same* quality-check + tus-upload path as the camera/gallery flows.
- **give/mobile/photos.vue** — on mount, attaches any queued shared image(s),
  then clears the queue.

## Risk / testing

- Pure-additive: web behaviour and existing camera/gallery flows are unchanged
  unless the native `FreegleShare` bridge is present (Android app only).
- Native (manifest + Java) can't be built locally; verified via CI Android build
  and on-device share test.
- No new dependencies.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #757, opened 2026-06-14
feat(android): app-icon long-press shortcuts (Give / Ask / Chats)

edwh:
Long-press the Freegle launcher icon → **Give**, **Ask**, **My Chats**. Each fires a `https://www.ilovefreegle.org/<path>` VIEW intent that the existing deep-link filter + `appUrlOpen` route to `/give` / `/find` / `/chats` — no new native code. (iOS shortcuts + share-into-Freegle are separate follow-ups.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #756, opened 2026-06-14
feat(app): haptic feedback (useHaptics + @capacitor/haptics)

edwh:
Adds native haptic feedback. `@capacitor/haptics` + a `useHaptics()` composable (no-op on web, lazy native import in-app; light/medium/heavy impacts + success/warning/error). First wiring: success haptic on chat reply sent from a notification. Reusable for posting, sending, swipe, etc. — the foundation for the agreed haptics work.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #759, opened 2026-06-14
feat(app): share an image into Freegle to start an Offer (iOS) [dark]

edwh:
## What

The **iOS half** of *Share-into-Freegle*. Adds an iOS share-sheet extension so a
photo can be shared straight into a new **OFFER** with the image pre-attached —
the iOS counterpart to the Android `ACTION_SEND` work in #758.

> **Stacked on #758.** This PR is based on `feat/share-into-freegle` (Android) and
> reuses its `pendingSharedImages` give-flow path. Merge #758 first (GitHub will
> retarget this PR's base to `master`), or merge them together.

## ⚠️ DARK / flag-gated — safe to merge, does nothing until enabled

Everything is a **no-op unless `ENABLE_SHARE_EXTENSION=true`**. By default the
Xcode project, the main app's entitlements and its signing/export are
**byte-for-byte unchanged**, so this **cannot affect the weekly App Store
release** until deliberately enabled. The disabled export composes to exactly
today's app(+NSE) provisioning map.

Enabling requires these **human, Apple-Developer-portal** steps first (no code
can do them — they need interactive Apple ID + 2FA):

1. Register App Group `group.org.ilovefreegle.iphone`; enable it on **both** App
   IDs (`org.ilovefreegle.iphone` and `org.ilovefreegle.iphone.ShareExtension`).
2. Pre-create the `.ShareExtension` App ID + an App Store distribution profile
   (same "create once by hand" approach as the NSE).
3. Regenerate the **main app** App Store profile so it includes the App Group.
4. Set `ENABLE_SHARE_EXTENSION=true` in the `deploy-apps` CircleCI context.

…then **on-device verification** (native code is not buildable locally).

## How

- **ShareViewController.swift** — no-UI handler: copies the shared image(s) into
  the App Group container, then opens the host app with
  `freegleshare://shared?p=<path>…` via the responder-chain `openURL:` technique.
- **ShareExtension/Info.plist** — `com.apple.share-services`, image activation
  rule, programmatic principal class (no storyboard to get wrong).
- **ShareExtension.entitlements** — App Group.
- **App/Info.plist** — registers the `freegleshare://` URL scheme.
- **add_share_extension_target.rb** — gated; adds the target and (build-time
  only, never committed) the App Group entitlement to the main app. Mirrors
  `add_nse_target.rb`.
- **Fastfile** — gated wiring + signing + a composed export map (disabled build
  == today's export).
- **stores/mobile.js** — `appUrlOpen` catches `freegleshare://shared` and queues
  the image(s) into the **same** `pendingSharedImages` → `/give/mobile/photos`
  path as Android.

## Risk / testing

- Default build path is provably unchanged (flag off). Validated: edited Fastfile
  and the new Ruby script both pass `ruby -c`; all plists are well-formed;
  `mobile.js` passes `node --check`.
- Native Swift can't be compiled locally — needs the on-device test above.
- The one runtime assumption to verify on-device: `Capacitor.convertFileSrc()`
  serving the App Group container path to the WKWebView. If it doesn't, the fix
  is to copy app-group→app-tmp in `AppDelegate` before forwarding the URL.

Part of *Share-into-Freegle*. Companion: #758 (Android).


edwh:
Superseded by #766, which consolidates the native-app feature PRs (#756 haptics, #757 app-icon shortcuts, #758 share-into-Freegle Android, #759 share-into-Freegle iOS) into a single PR rebased onto current master. Branch kept (not deleted) in case it's needed.

PR #777, opened 2026-06-17
fix(digest): exclude banned users from post emails (standalone — not in rippling merge chain)

edwh:
## Summary
A user banned from a group must not receive its posts in the unified digest (immediate or daily). Banning deletes the membership, so the membership-based recipient queries normally exclude them — but if a membership row lingers (TN sync re-add, stale row, race) they would still be emailed. Adds an explicit `users_banned` exclusion to both recipient paths:

- **Immediate** recipient query — exclude members banned from that group.
- **Daily** `getPostsForUser` — exclude groups the user is banned from.

## Tests
`test_immediate_digest_excludes_banned_user` (banned member not emailed; poster still is) and `test_daily_digest_excludes_banned_user` (banned recipient gets no daily digest). Full Laravel suite green.

## Follow-up (noted from cross-PR review)
The future reach-gated digest (PR F) selects recipients by **location**, not membership — it will need the same `users_banned` exclusion added to its reach recipient query. Not a regression here.

Not merged — humans merge.

### DIFF
--- a/iznik-batch/app/Services/UnifiedDigestService.php
+++ b/iznik-batch/app/Services/UnifiedDigestService.php
@@ -233,6 +233,13 @@ class UnifiedDigestService
             ->where('memberships.groupid', $groupid)
             ->where('memberships.emailfrequency', Membership::EMAIL_FREQUENCY_IMMEDIATE)
             ->where('memberships.collection', Membership::COLLECTION_APPROVED)
+            // A user banned from the group must never be emailed about its posts,
+            // even if a membership row still exists (re-added by TN sync, or stale).
+            ->whereNotExists(function ($q) use ($groupid) {
+                $q->select(DB::raw(1))->from('users_banned')
+                    ->whereColumn('users_banned.userid', 'memberships.userid')
+                    ->where('users_banned.groupid', $groupid);
+            })
             ->whereNull('users.deleted')
             ->where(function ($q) {
                 $q->whereNull('users.lastaccess')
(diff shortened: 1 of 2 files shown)


edwh:
On "how could a banned user's row linger?" — banning already **deletes** the membership row (see `iznik-server-go/user/authorization.go:45`: "banning deletes the memberships row"). So an `Approved` membership for a banned user should not normally exist; when it does, it's been re-created (e.g. a TrashNothing membership sync re-adding the row) or it's stale data. The durable fix is preventing that re-add at the sync/ban layer — this digest-query exclusion is reasonable defense-in-depth but treats the symptom.

On applying the same protection to rippling: the rippling **immediate-mail sending path isn't built yet** — #771 only adds the `messages_reach_notified` ledger, not the send. So there's no recipient query to patch today; the banned-user exclusion needs to be baked into that query when the rippling immediate-send is implemented (and into any rippling digest recipient selection). Flagging here so it isn't missed.

Suggest: keep this exclusion (harmless safety net), but track the root cause (why a banned user retains an Approved membership) separately, and ensure rippling's send path carries the same guard.

edwh:
Closing — confirmed against production this is **not a live leak**.

There are 23 banned-user-with-membership rows, but **none created since 2023**, and every current write path is already guarded:
- ban deletes the membership: Go `membership.go:213` ("Ban" case) and `:1312`, V1 `User::removeMembership` (`User.php:1080`), and the spam bulk-ban goes through `removeMembership` (`Spam.php:773`);
- every membership insert checks `users_banned` first: Go `addMemberToGroup` (`:1219`), TN/partner join (`:1170`), post-autojoin (`message.go:2405`), V1 `addMembership` via `isBanned()`;
- the user merge reconciles (deletes id2's memberships for the combined ban set, `user.go:2717`).

So the 23 are frozen historical rows (pre-2023, predating delete-on-ban / merge fixes), and nothing recreates them.

Adding a permanent `users_banned` anti-join to the digest recipient queries to dodge 23 dead rows is the wrong fix. The remedy is a one-off reconciliation — the ban is authoritative, so those users simply shouldn't have a membership:

```sql
DELETE m FROM memberships m
JOIN users_banned ub ON ub.userid = m.userid AND ub.groupid = m.groupid;
```

After that there's nothing to exclude and no live path recreates it. Closing in favour of that cleanup.

PR #780, opened 2026-06-17
fix(email): digest image/title link opens reply overlay instead of showing post

edwh:
## Summary

- **Bug**: Clicking the photo or title link in a Freegle digest email opens the ChatReplyPane overlay instead of showing the expandable post page (Discourse topic 9363 post 29)
- **Root cause**: `_card.blade.php` used `messageUrl` (which contains `?reply=1`) for both the thumbnail image `href` and the title link. `MessageExpanded` detects `?reply=1` on mount and auto-opens the full-screen reply pane, covering the photo so it cannot be expanded.
- **Fix**: Two-part change:
  1. `_card.blade.php`: image `href` and title `href` now use `summaryUrl` (no `?reply=1`); only the Reply button retains `messageUrl`
  2. `MessageExpanded.vue`: guard the `?reply=1` auto-open to `inModal`/`fullscreenOverlay` contexts, so any stale or mis-linked `?reply=1` URL on the standalone page never covers the photo — the Reply button is always visible there anyway

## Test plan

- [ ] New Vitest test `email link ?reply=1 on standalone page > does not auto-open reply pane when ?reply=1 is in URL on standalone page` fails on the original code and passes after the fix
- [ ] All 116 existing `MessageExpanded` Vitest tests pass
- [ ] Clicking a post image or title in a digest email now opens the standalone post page where the photo is visible and expandable
- [ ] Clicking the Reply button in a digest email still navigates to `/message/{id}?reply=1`; the Reply overlay auto-opens when the component is embedded in a modal/overlay context

### DIFF
--- a/iznik-batch/resources/views/emails/mjml/digest/_card.blade.php
+++ b/iznik-batch/resources/views/emails/mjml/digest/_card.blade.php
@@ -48,7 +48,7 @@
                  the column shrinks on mobile, no fluid-on-mobile blow-up. --}}
             <mj-image
                 src="{{ $post['thumbImageUrl'] }}"
-                href="{{ $post['messageUrl'] }}"
+                href="{{ $post['summaryUrl'] }}"
                 alt="{{ $post['itemName'] }}"
                 width="100%"
                 border-radius="4px"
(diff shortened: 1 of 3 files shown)


edwh:
Closing in favour of #778, which addresses the reported issue ([Discourse 9363, post 29](https://discourse.ilovefreegle.org/t/9363/29) — photo not expandable after an email link opens the chat-style reply pane) directly and without a regression: it makes the post's photo open the existing `MessagePhotosModal` carousel **inside** the reply pane (swipe/prev-next/pinch-zoom over all the item's photos).

This PR takes a different approach — point the email image/title at `summaryUrl` and gate the standalone `?reply=1` auto-open on `inModal || fullscreenOverlay`. That fixes photo-viewing, but it **degrades the email Reply CTA**: clicking "Reply" in an email currently auto-opens the reply pane on the standalone message page, and this would stop that (costing the user an extra click).

If we do want the email image/title to open the post view rather than the reply, the single blade change (`messageUrl` → `summaryUrl` in `_card.blade.php`) is worth salvaging on its own — but the `MessageExpanded` auto-open removal shouldn't ship. Reopen if you'd prefer this approach over #778.

PR #781, opened 2026-06-17
fix(digest): send WANTED posts immediately — only OFFERs wait for an AI image

edwh:
## Summary
- `isImmediateMessageReady()` deferred **any** attachment-less message for the full `ATTACHMENT_WAIT_DEADLINE_MINUTES` window, waiting for an AI-generated illustration.
- But only **OFFERs** get AI illustrations — a WANTED (or any non-OFFER) was needlessly held back for an image that never arrives, delaying the immediate email by the whole wait window.
- Non-OFFERs are now ready immediately.

## Why
Raised during the rippling-out side-effect review: "consider the deferred posts for wanted not just offers." Standalone digest-correctness fix, independent of the rippling merge chain.

## Test Plan
- New: `test_immediate_sends_recent_wanted_without_attachment` — a freshly-posted WANTED with no attachment is mailed immediately and the cursor advances.
- Unchanged: the OFFER defer (`test_immediate_defers_recent_message_with_no_attachment`) and after-deadline (`test_immediate_sends_unattached_message_after_deadline`) tests still pass (OFFERs still wait).

### DIFF
--- a/iznik-batch/app/Services/UnifiedDigestService.php
+++ b/iznik-batch/app/Services/UnifiedDigestService.php
@@ -401,6 +401,14 @@ class UnifiedDigestService
             }
         }
 
+        // Only OFFERs get AI-generated illustrations, so only an OFFER is worth
+        // holding back while an image is generated. A WANTED (or any non-OFFER)
+        // will never gain an attachment — send it immediately rather than
+        // needlessly deferring it for the whole attachment-wait deadline.
+        if ((string) $message->type !== Message::TYPE_OFFER) {
+            return true;
+        }
+
         $arrival = $message->mg_arrival
             ?? ($message->arrival instanceof \Carbon\Carbon ? $message->arrival : \Carbon\Carbon::parse($message->arrival ?? $message->date));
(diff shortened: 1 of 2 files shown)


edwh:
Closing as a bad fix — the premise doesn't hold.

"Only OFFERs get AI illustrations" is false for the production path: `messages:generate-illustrations` → `MessageIllustrationsService` runs every minute with **no message-type filter**, and `extractItemName()` strips the `WANTED:`/`OFFER:` prefix — so **WANTED posts already get AI illustrations** (typically within ~1 minute). The OFFER-only "WANTED uses stock illustrations" gates are in the separate **EEE** experiment (`EeeClassificationService`, `EeeFeatureExperimentCommand`), not this path.

So sending WANTEDs immediately would email them *before* their incoming illustration arrives. And the original "defer until attachment or deadline" is already capped by `ATTACHMENT_WAIT_DEADLINE_MINUTES`, so nothing is deferred "forever" — a WANTED whose AI generation fails still goes out at the deadline, exactly like an OFFER.

(Live-DB confirmation was pending — the tunnel was down — but the deadline backstop makes closing safe regardless.)

PR #785, opened 2026-06-17
fix(chitchat): cap Compressor maxWidth/maxHeight to prevent oversized uploads

edwh:
## Root Cause

`@uppy/compressor` delegates options directly to `compressorjs`, whose defaults are `maxWidth: Infinity, maxHeight: Infinity`. Both `OurUploader.vue` (chitchat/newsfeed) and `PhotoUploader.vue` (compose flow) called `.use(Compressor)` with no options, so the web upload path applied JPEG quality compression (reducing file size) but never constrained image dimensions — allowing 1536×2048 images (285 KB as reported) to reach the server at full resolution.

## Evidence

```
# Before fix — Compressor registered with no options (Infinity limits):
.use(Compressor)

# Tests FAIL on buggy code:
✗ configures Compressor with maxWidth to prevent oversized uploads (Discourse 9749)
  Expected: typeof compressorOpts.maxWidth to be 'number', got undefined
```

## Fix

Added `{ maxWidth: 1024, maxHeight: 1024 }` to both `.use(Compressor)` calls, matching the advisory `height: 1024` constraint already used on the Capacitor Camera native-app path:

- `iznik-nuxt3/components/OurUploader.vue` (chitchat/newsfeed path)
- `iznik-nuxt3/components/PhotoUploader.vue` (compose flow — same pattern)

## Other Call Sites

Both `.use(Compressor)` call sites in the codebase are now fixed. No other occurrences.

## Test

- File: `iznik-nuxt3/tests/unit/components/OurUploader.spec.js`
- Command: `npm run test:unit:run -- OurUploader`
- Proves: FAIL before fix (no options → no dimension cap), PASS after fix (98✓)
- Also: `PhotoUploader.spec.js` 89✓

## Discourse

https://discourse.ilovefreegle.org/t/9749/5

### DIFF
--- a/iznik-nuxt3/components/OurUploader.vue
+++ b/iznik-nuxt3/components/OurUploader.vue
@@ -407,7 +407,7 @@ onMounted(() => {
       uploadDataDuringCreation: true,
       retryDelays: [0, 3000, 5000, 10000, 20000],
     })
-    .use(Compressor)
+    .use(Compressor, { maxWidth: 1024, maxHeight: 1024 })
   uppy.on('file-added', (file) => {
     console.log('Added file', file)
   })
--- a/iznik-nuxt3/components/PhotoUploader.vue
+++ b/iznik-nuxt3/components/PhotoUploader.vue
@@ -707,7 +707,7 @@ onMounted(() => {
       endpoint: runtimeConfig.public.TUS_UPLOADER,
       uploadDataDuringCreation: true,
     })
-    .use(Compressor)
+    .use(Compressor, { maxWidth: 1024, maxHeight: 1024 })
 
   uppy.value.on('complete', handleUppySuccess)
   const scheduleRetry = createRetryCoalescer(() => uppy.value)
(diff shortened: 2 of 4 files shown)


edwh:
Closing — misdiagnosed. This topic is **nothing to do with image size**.

[9749](https://discourse.ilovefreegle.org/t/9749/5) "Silly niggle on chitchat" is about the **"…" (three-dots) options button in the ChitChat list sitting too close to the post text** (Jos, post #1). Two prior fixes (#612, #644) widened the gap; post #5 says *"mostly good but still the odd one."*

The `"1536×2048 285 KB"` in post #5 is just Discourse's metadata for the attached **screenshot** — not the subject of the complaint. Reading it as an image-size bug and capping the uploader's `maxWidth/maxHeight` doesn't touch the actual issue. The correct fix is CSS spacing in the ChitChat post layout so a long first line of text reserves room for the "…" button. Closing; a correct fix will follow.

## Discourse

https://discourse.ilovefreegle.org/t/9749/5

Issue #20, opened 2026-01-16
Investigate ParaTest for parallel test execution in iznik-batch


edwh:
## Background

We previously attempted to use ParaTest for parallel test execution in iznik-batch but encountered reliability issues, so we reverted to running tests serially. The tests currently run at an acceptable speed, but parallel execution could provide further improvements.

## Recent Discovery

We recently identified and fixed a root cause of flaky test failures related to view compilation:

- **Problem**: Laravel's `isExpired()` was recompiling views when source timestamps >= compiled timestamps, causing race conditions
- **Fix**: Added `VIEW_CHECK_TIMESTAMPS=false` for tests, ensuring precompiled views are used unconditionally (commit d20c539)

The ParaTest bootstrap isolation code we had (per-worker view cache directories) was actually **making things worse** - it redirected to empty directories, bypassing the precompiled views entirely.

## Investigation Needed

Now that views are properly precompiled before tests and timestamp checking is disabled:

1. **Test ParaTest reliability** - Does it work without the view isolation workarounds?
2. **Measure performance gain** - Is parallel execution significantly faster?
3. **Check for other race conditions** - Are there other shared resources that cause issues?

## To Test

```bash
# Install ParaTest
composer require --dev brianium/paratest

# Run with ParaTest
docker exec freegle-batch vendor/bin/paratest --testsuite=Unit,Feature
```

## Acceptance Criteria

- [ ] ParaTest runs reliably (no flaky failures)
- [ ] Test execution time is measurably improved
- [ ] No additional isolation workarounds needed

PR #786, opened 2026-06-17
fix(modtools): correct off-by-one badge count for newly-promoted moderator

edwh:
## Summary

- `pendingmembers` (count of pending group join requests awaiting approval) was included in `work.total`, which drives the red hamburger badge count in ModTools
- The frontend has **no left-menu red badge** for `pendingmembers` — so the hamburger showed `pendingmembers` more than the sum of visible left-menu items
- For a newly-promoted moderator whose group has any pending member applications, the hamburger count was exactly that many too high (typically 1)

## Root cause

In `iznik-server-go/session/session.go`, the `total` formula included `pendingmembers`:
```go
total := pending + spam + pendingmembers + spammembers + ...
```
But a search of the entire `iznik-nuxt3` codebase finds **zero** uses of `pendingmembers` as a badge count in any left-menu component. Every other field in `total` has a visible left-menu red badge home; `pendingmembers` does not.

## Fix

Remove `pendingmembers` from `total`. It remains in the `work` map so the frontend can display it in a dedicated badge if one is added later.

## Test plan

- [x] New failing test `TestWorkCountPendingMembersExcludedFromTotal` added: creates a mod + group with 1 pending member application, asserts `work.total` equals sum of all actionable items **excluding** `pendingmembers` — failed before fix, passes after
- [x] Updated `TestWorkCountTotalExcludesInformational` to reflect `pendingmembers` is now excluded from `total`
- [x] Full Go test suite: **3183 tests pass, 0 failures**

Fixes Discourse topic 9654 post 19.

### DIFF
--- a/iznik-server-go/session/session.go
+++ b/iznik-server-go/session/session.go
@@ -1403,7 +1403,9 @@ func GetSession(c *fiber.Ctx) error {
 
 		// Total only includes actionable work items (primary/red badge counts),
 		// not informational ones (other/blue badge counts like chatreviewother, happiness, giftaid, pendingother).
-		total := pending + spam + pendingmembers + spammembers + pendingevents +
+		// pendingmembers is intentionally excluded: the frontend has no left-menu red badge for it,
+		// so including it inflates the hamburger count beyond the sum of visible items (bug 9654).
+		total := pending + spam + spammembers + pendingevents +
 			pendingadmins + editreview + pendingvolunteering + stories +
 			spammerpendingadd + spammerpendingremove +
 			chatreview + newsletterstories + relatedmembers + housekeeping + cronjobs +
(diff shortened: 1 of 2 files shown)


edwh:
Closing — the premise isn't borne out by live data, and the reported issue was already fixed by a different change.

1. **`pendingmembers` ≈ 0 in practice** — Freegle groups don't accumulate pending member applications now, so `pendingmembers` doesn't inflate `work.total` for real moderators. The fix targets a condition that essentially doesn't occur (not verified against live data).
2. **The reported +1 was already resolved.** Discourse 9654, post #20: after *"include spam in Pending badge so total matches visible items"*, Neville reports *"The ModTools red counts seem to be correct at the moment in all places."* So the spurious +1 was the spam-not-in-pending-badge issue, already fixed — not `pendingmembers`.

Excluding `pendingmembers` from `total` is defensible for badge consistency in the abstract, but it doesn't address the reported bug (already fixed) and its triggering condition isn't present in live data. Closing.

## Discourse

https://discourse.ilovefreegle.org/t/9654/19

PR #787, opened 2026-06-17
fix(modtools_chat): show member name not group name after group-change + standard message

edwh:
## Summary

- **Bug**: After a volunteer edits a member's post to move it to a new group and sends a standard welcome message, ModTools chat showed "GroupName Volunteers" instead of the member's name.
- **Root cause**: The User2Mod SQL in `listChats()` (non-memberOnly / ModTools path) aliased `fullname` as `firstname` and hardcoded `lastname = ''`. When a member has `fullname = NULL` (but does have `firstname`/`lastname`), the naming logic fell through to the group-name fallback branch.
- **V1 parity**: this restores the V1 behaviour. V1 `ChatRoom` computes the name as `CASE WHEN fullname IS NOT NULL THEN fullname ELSE CONCAT(firstname,' ',lastname) END` (`iznik-server/include/chat/ChatRoom.php:91-92`) and shows the mod *"<member> on <group>"*, never *"<group> Volunteers"* (that suffix is only the member's own view, `:172`/`:720`). V2's `fullname AS firstname, '' AS lastname` aliasing dropped the firstname/lastname fallback — a V2 regression, not intended parity. So this is a parity restoration, not a behaviour change.
- **Fix**: Fetch real `firstname`/`lastname` columns from the `users` table in all three ModTools User2Mod SQL query paths, and add a fallback naming branch that uses `firstname + lastname` when `fullname` is empty but the member has partial name data.
- **Test**: `TestModToolsChatNameUsesFirstLastWhenFullnameNull` creates a member with `firstname`/`lastname` but NULL `fullname`, creates a User2Mod chat, and asserts the ModTools chat list shows the member's name rather than "GroupName Volunteers".

## Files changed

- `iznik-server-go/chat/chatroom.go` — fix the SQL and naming logic in `listChats()`
- `iznik-server-go/test/chat_test.go` — add regression test

## Test plan

- [x] `TestModToolsChatNameUsesFirstLastWhenFullnameNull` passes
- [x] All 3183 Go tests pass locally

## Discourse

https://discourse.ilovefreegle.org/t/9719/1

### DIFF
--- a/iznik-server-go/chat/chatroom.go
+++ b/iznik-server-go/chat/chatroom.go
@@ -722,8 +722,8 @@ func listChats(myid uint64, chattypes []string, start string, search string, onl
 				// Exclude backup mods (active:0 in membership settings) unless searching.
 				unions = append(unions,
 					"SELECT 0 AS search, user1 AS otheruid, nameshort, namefull, "+
-						"COALESCE((SELECT fullname FROM users WHERE users.id = user1), '') AS firstname, "+
-						"'' AS lastname, "+
+						"COALESCE((SELECT firstname FROM users WHERE users.id = user1), '') AS firstname, "+
+						"COALESCE((SELECT lastname FROM users WHERE users.id = user1), '') AS lastname, "+
 						"COALESCE((SELECT fullname FROM users WHERE users.id = user1), '') AS fullname, "+
 						"(SELECT deleted FROM users WHERE users.id = user1) AS otherdeleted, "+
 						atts+", c1.status, NULL AS lasttype FROM chat_rooms "+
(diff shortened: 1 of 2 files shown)


edwh:
⚠️ **Diagnosis not verified against live data — confirm before merge.**

The root cause assumes the reported member (41046202, per [Discourse 9719/1](https://discourse.ilovefreegle.org/t/9719/1)) has `fullname = NULL` but `firstname`/`lastname` populated. That hasn't been checked, and the fix is conditional on it:

- If 41046202 has `firstname`/`lastname` + NULL `fullname` → the new fallback branch fixes the "Newham Freegle Volunteers" display. ✅
- If **all three name fields are empty** (a brand-new member with no name yet — plausible for someone who just joined by posting) → the code still falls through to the `<group> + " Volunteers"` branch and **the bug persists**. The fix doesn't cover this case; you'd need a non-name fallback (email local-part, or "Member #<id>").
- If the member is `user2` rather than `user1` in the chat → V2's `user1 AS otheruid` resolves the *wrong* user entirely, which would be a different bug.

Confirm with (live DB):
```sql
SELECT id, fullname, firstname, lastname, added FROM users WHERE id = 41046202;
SELECT id, chattype, user1, user2, groupid FROM chat_rooms
WHERE chattype = 'User2Mod' AND (user1 = 41046202 OR user2 = 41046202);
```

The reporter attributes it to the group-change + standard message; that's consistent with `user1`=member having an empty name at chat-creation time, but the static `firstname`/`lastname` fallback only rescues the partial-name case — so the actual member's row is the deciding evidence.

edwh:
✅ **Verified against live data.** Member 41046202 has `fullname = NULL` with both `firstname` and `lastname` populated — exactly the case this fix targets.

Before the fix, the V2 listChats SQL (`fullname AS firstname, '' AS lastname`) discarded both name parts, so a NULL-`fullname` member fell through to the `<group> + " Volunteers"` branch — the reported "Newham Freegle Volunteers". With the fix, the firstname/lastname fallback renders the member's name. Diagnosis confirmed; #787 resolves the reported case and restores V1 parity (V1 uses `CONCAT(firstname,' ',lastname)` when `fullname` is NULL).

PR #789, opened 2026-06-17
fix(myposts): age out old pending posts by original date, not arrival

edwh:
## Root Cause

`applyExpiry` in `message/message.go` ages messages against `messages_groups.arrival` by default. The prior fix (#664 / Discourse 9481/561) overrode this to use `messages.date` for `COLLECTION_REJECTED` posts. However, it did not extend the same fix to `COLLECTION_PENDING` posts.

A post returned to the Pending queue gets a fresh `messages_groups.arrival` (today), while its `messages.date` is the original post date — potentially 69+ months ago. The age check used the recent arrival, so `daysAgo < expireTime (90)` and the post was never filtered out. It kept appearing in the member's active My Posts list.

## Evidence

```
--- FAIL: TestOldPendingMessageClassifiedAsOld (0.29s)
    message_test.go:347:
        Error:       Should be false
        Messages:    Years-old pending post should NOT appear in active (Discourse 9481/583)
    message_test.go:356:
        Error:       Should be true
        Messages:    Years-old pending post should be hasoutcome=true (Old) in non-active query
```

## Fix

Extended the `COLLECTION_REJECTED` condition in `applyExpiry` to also cover `COLLECTION_PENDING`:

```go
// Before
if m.Collection == utils.COLLECTION_REJECTED && !m.Date.IsZero() {
    ageBasis = m.Date
}

// After
if (m.Collection == utils.COLLECTION_REJECTED || m.Collection == utils.COLLECTION_PENDING) && !m.Date.IsZero() {
    ageBasis = m.Date
}
```

Fresh pending posts (where `m.Date` ≈ today) are unaffected: `daysAgo ≈ 0 < expireTime`.

## Other Call Sites

`ageBasis` is only set inside `applyExpiry`. `markExpiredMessages` already correctly excludes Pending/Rejected from the spatial-entry nullity check (line 1185). No other changes needed.

## Test

**File**: `iznik-server-go/test/message_test.go` — `TestOldPendingMessageClassifiedAsOld`

**Command**: `go test ./test/... -run TestOldPendingMessageClassifiedAsOld`

**Proves**: Fail-before (400-day-old pending post appears in active query) → Pass-after (filtered out). Mid-age (60-day) pending post still appears in active (V1 parity).

Full suite: 3184 tests pass (3183 pre-existing + 1 new).

## Discourse

https://discourse.ilovefreegle.org/t/9481/583

### DIFF
--- a/iznik-server-go/message/message.go
+++ b/iznik-server-go/message/message.go
@@ -1107,13 +1107,14 @@ func applyExpiry(db *gorm.DB, msgs []MessageSummary) []int {
 		}
 
 		// Age the post against the same expireTime V1 uses (maxagetoshow /
-		// EXPIRE_TIME = 90 days, or the repost window). For Rejected posts age by
-		// the ORIGINAL date (messages.date) rather than arrival: a rejected post's
-		// arrival can be recent while the post itself is years old, which would
-		// otherwise keep a long-dead rejected message in the member's active posts
-		// (Discourse topic 9481/561). V1 capped a member's own posts by age too.
+		// EXPIRE_TIME = 90 days, or the repost window). For Rejected/Pending posts
+		// age by the ORIGINAL date (messages.date) rather than arrival: a post
+		// returned to the pending queue gets a recent arrival while the original
+		// message can be years old, which would otherwise keep it in the member's
+		// active posts (Discourse topic 9481/561 for Rejected, 9481/583 for Pending).
+		// V1 capped a member's own posts by age too.
 		ageBasis := m.Arrival
-		if m.Collection == utils.COLLECTION_REJECTED && !m.Date.IsZero() {
+		if (m.Collection == utils.COLLECTION_REJECTED || m.Collection == utils.COLLECTION_PENDING) && !m.Date.IsZero() {
 			ageBasis = m.Date
 		}
 		daysAgo := int(now.Sub(ageBasis).Hours() / 24)
--- a/iznik-server-go/test/message_test.go
+++ b/iznik-server-go/test/message_test.go
@@ -291,6 +291,76 @@ func TestOldRejectedMessageClassifiedAsOld(t *testing.T) {
 	}
 }
 
+func TestOldPendingMessageClassifiedAsOld(t *testing.T) {
+	// A post that is 69+ months old but has been returned to the Pending queue
+	// (recent messages_groups.arrival) must still be aged out using its ORIGINAL
+	// date (messages.date), not the recent arrival. Regression from Discourse
+	// topic 9481/583: the prior fix (#664) only covered COLLECTION_REJECTED.
+	db := database.DBConn
+	prefix := uniquePrefix("oldpend")
+	groupID := CreateTestGroup(t, prefix)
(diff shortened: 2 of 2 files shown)


edwh:
⚠️ **Needs live-data verification — and it targets a different collection than the report.**

The report (Discourse 9481/583) is about **REJECTED** posts: *"I see 69-month-old **rejected** posts being shown within my active posts"*, and the reporter's own example is a rejected wanted `#70516224` that is *"now only shown in the Old list"* — i.e. the REJECTED date-aging from #664 **is working**.

This PR extends the date-basis to **PENDING** posts, which is a different collection that wasn't reported. Before merging, that needs to be shown to be a real case:

```sql
-- pending posts with an old original date but a recent (reset) arrival
SELECT COUNT(*)
FROM messages m
JOIN messages_groups mg ON mg.msgid = m.id
WHERE mg.collection = 'Pending'
  AND m.date < NOW() - INTERVAL 90 DAY
  AND mg.arrival > NOW() - INTERVAL 90 DAY;
```

- If this is **0**, the PENDING extension fixes a case that doesn't occur — close it (like #786).
- If non-zero, it's a real (if unreported) issue and the fix is reasonable — then it can ship on evidence rather than analogy to the REJECTED path.

Also worth confirming *why* a years-old post would be returned to Pending with a fresh arrival, since that's the mechanism the fix assumes.

## Discourse

https://discourse.ilovefreegle.org/t/9481/583

edwh:
Closing: the premise is incorrect, verified against the production DB for the exact posts the reporter named in Discourse 9481/583 (#70516224, #70023547, #70022383).

What the report actually says (9481/561 and /583): old **Rejected** posts from ~2020 appearing in the member's active My Posts. The reporter explicitly calls them "rejected", not pending.

Production reality for those three posts:

| id | collection | date | arrival | outcome | shows |
|---|---|---|---|---|---|
| 70022383 | Rejected | 2020-08-15 | 2020-08-15 | none | active (bug) |
| 70023547 | Rejected | 2020-08-15 | 2020-08-15 | none | active (bug) |
| 70516224 | Rejected | 2020-08-28 | 2020-08-28 | Withdrawn | Old (correct) |

This contradicts the PR's root cause on two points:
1. The posts are **Rejected**, not Pending - so extending the date-basis to `COLLECTION_PENDING` touches none of them.
2. Their **arrival equals their 2020 original date** - there is no "returned to the pending queue with a fresh arrival". Nothing was re-queued; they are old posts appearing unchanged.

The only field that differs between the post correctly in Old and the two stuck in active is the **outcome** (Withdrawn vs none) - so the active/Old split here is driven by outcome, not arrival-vs-date. The PR's test (`TestOldPendingMessageClassifiedAsOld`) simulates a fresh-arrival + old-date pending post, a scenario that does not occur for the reported posts, so it passes without fixing the reported bug.

Real bug (still open): old **Rejected** posts with **no outcome** remain in active My Posts despite #664's rejected-by-date age-out. Needs diagnosis of why that age-out does not remove them on the My Posts path (e.g. rejected posts are shown for the Edit & Resend flow regardless of age), not a pending-collection change.

PR #790, opened 2026-06-17
fix(chat-delivery): recover TN chat replies misrouted to noreply@ instead of notify-

edwh:
## Summary

- **Root cause**: Freegle chat notifications are sent FROM `noreply@ilovefreegle.org` with Reply-To set to `notify-{chatId}-{userId}@users.ilovefreegle.org`. TN sometimes sends follow-up chat-thread replies to the From address instead of the Reply-To, which causes `isDigestReply()` to fire (it returns true when envelope-to is `noreply@`). This triggers `handleDigestReply()` which sends back a "we can't tell which post you'd like to reply to" auto-response — a false error for what was a valid chat-thread reply.
- **Fix**: Add a `handleMisroutedChatReply()` check in `route()` before the digest-reply phase. When the envelope-to is `noreply@` and `In-Reply-To` references a chat notification Message-ID (`chat-{chatId}-msg-{msgId}@users.ilovefreegle.org`), we extract the chatId, look up the sender's userId in `users_emails`, validate they are a member of the chat, and route to the normal chat processing path via `processChatReply()`.
- **Refactoring**: Extracted the core logic of `handleChatNotificationReply` into `processChatReply(ParsedEmail, chatId, userId)` so both the normal notify-path and the recovery path share the same validation and message-creation logic.

## Discourse

https://discourse.ilovefreegle.org/t/9800/1

## Test plan

- [ ] New failing→passing test: `test_tn_chat_reply_to_noreply_routed_as_chat` — sends an email FROM TN user TO `noreply@` with `In-Reply-To: <chat-{chatId}-msg-99@users.ilovefreegle.org>`, asserts `TO_USER` result, no `DigestReplyNotice` sent, chat message created
- [ ] Full `IncomingMailServiceTest` suite: 156 tests, 0 failures
- [ ] Verify emails with no `In-Reply-To` still hit `handleDigestReply` as before
- [ ] Verify emails from unknown senders fall through to existing digest reply handling

### DIFF
--- a/iznik-batch/app/Services/Mail/Incoming/IncomingMailService.php
+++ b/iznik-batch/app/Services/Mail/Incoming/IncomingMailService.php
@@ -134,7 +134,15 @@ class IncomingMailService
             return $this->handleHumanReplyToBounceAddress($email);
         }
 
-        // Phase 3c: Reply to digest email (sent from noreply@).
+        // Phase 3c: Recover TN chat replies misrouted to noreply@ instead of notify-.
+        // TN sometimes sends chat-thread replies to the From address rather than the
+        // Reply-To address. Detect via In-Reply-To referencing a chat Message-ID.
+        $misroutedResult = $this->handleMisroutedChatReply($email);
+        if ($misroutedResult !== null) {
+            return $misroutedResult;
+        }
+
+        // Phase 3d: Reply to digest email (sent from noreply@).
         // Send helpful auto-response explaining how to reply to specific posts.
         if ($email->isDigestReply()) {
             return $this->handleDigestReply($email);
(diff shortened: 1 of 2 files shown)


edwh:
Closing — re-reading the report, this is a **TN-side bug that TN has already fixed**, not a Freegle defect to work around.

Discourse 9800, post #2: *"I will write to Andrew Trusty separately… Andrew has just replied. It was a bug, and he thinks he has fixed it."* — TN was misrouting the chat reply (to the From address instead of Reply-To), and TN (Andrew Trusty) fixed it on their end.

So adding a permanent Freegle-side `handleMisroutedChatReply()` recovery path to compensate for an external bug that no longer exists is unwarranted complexity in the inbound mail router. Closing.

## Discourse

https://discourse.ilovefreegle.org/t/9800/1

PR #792, opened 2026-06-17
fix(moderation): clear contentcheck_reasons on approval so IP-abuse warning does not persist

edwh:
## Root Cause

The IP-abuse (and other contentcheck) warning was shown to moderators on posts already in `collection='Approved'` because the GET `/api/message/:id` endpoint returned `contentcheck_reasons` from `messages_groups` unconditionally — regardless of whether the message was Pending or Approved.

Prior fix (commit 3bd5d19) cleared `contentcheck_reasons` in the DB at approval time, which prevents new approvals from persisting the flag. However, messages approved _before_ that fix was deployed still carry stale `contentcheck_reasons` in the DB — the root cause of the reporter's complaint: "still showing on posts approved 2 days ago."

**Complete fix (this re-attempt):** In addition to the DB-level clear at approval time, the API response builder now suppresses `ContentcheckReasons` for any `messages_groups` row whose `collection = 'Approved'`. This is retrospective: no DB back-fill is required, and the warning vanishes immediately for all already-approved historical records.

## Evidence

**Test FAILS on unfixed code (3187✓ 1✗):**
```
=== RUN   TestAlreadyApprovedMessageSuppressesContentcheckReasonsInAPIResponse
    message_test.go:900:
        Error:      Expected nil, but got: &json.RawMessage{...IpAbuse...}
        Messages:   contentcheck_reasons must not be returned for an approved group (retrospective fix)
--- FAIL: TestAlreadyApprovedMessageSuppressesContentcheckReasonsInAPIResponse (0.06s)
FAIL
```

**After fix: All 3188 tests pass (3188✓)**
```
--- PASS: TestApproveMessageClearsContentcheckReasons (0.54s)
--- PASS: TestAlreadyApprovedMessageSuppressesContentcheckReasonsInAPIResponse (0.42s)
```

## Fix

**`iznik-server-go/message/message.go`** — after fetching `messages_groups`, before building the response: iterate over message groups and set `ContentcheckReasons = nil` for any group with `collection = 'Approved'`. The warning is only relevant during pending moderation; clearing it in the response makes the fix retrospective.

**Already included from prior commit:** `handleApprove` sets `contentcheck_reasons = NULL` in the UPDATE so future approvals keep the DB clean. `AutoApproveService.php` also clears it on the 48h auto-approve path.

## Other Call Sites

`contentcheck_reasons` is fetched from `messages_groups` in exactly one SQL query in the Go API (line 449 of `message.go`); the suppression loop applies immediately after. No other Go call sites.

PHP batch services (`ContentCheckService`, `AutoApproveService`) already clear `contentcheck_reasons` on their own approval paths.

## Test

| Test | File | Proves |
|------|------|--------|
| `TestAlreadyApprovedMessageSuppressesContentcheckReasonsInAPIResponse` | `iznik-server-go/test/message_test.go` | Retrospective fix: GET response omits `contentcheck_reasons` for an already-approved message that still has stale data in the DB |
| `TestApproveMessageClearsContentcheckReasons` | `iznik-server-go/test/message_test.go` | Forward-path fix: approving via API clears `contentcheck_reasons` in DB |

Run: `curl -s -X POST http://localhost:8081/api/tests/go -d '{"filter":"TestAlreadyApproved|TestApproveMessageClears"}'`

## Discourse
https://discourse.ilovefreegle.org/t/9768/5

### DIFF
--- a/iznik-batch/app/Services/AutoApproveService.php
+++ b/iznik-batch/app/Services/AutoApproveService.php
@@ -196,6 +196,7 @@ class AutoApproveService
         // V1 approve(): UPDATE messages_groups SET collection='Approved', approvedby=whoAmId(),
         // approvedat=NOW(), arrival=NOW() WHERE msgid=? AND groupid=? AND collection!='Approved'
         // V1 whoAmId() returns NULL in cron context (no session).
+        // Also clear contentcheck_reasons so IP-abuse warnings are not shown after auto-approval.
         DB::table('messages_groups')
             ->where('msgid', $candidate->msgid)
             ->where('groupid', $groupid)
@@ -205,6 +206,7 @@ class AutoApproveService
                 'approvedby' => null,
                 'approvedat' => now(),
                 'arrival' => now(),
+                'contentcheck_reasons' => null,
             ]);
 
         // V1 autoapprove() log: type=Message, subtype=Autoapproved.
--- a/iznik-server-go/message/message.go
+++ b/iznik-server-go/message/message.go
@@ -1757,7 +1757,9 @@ func handleApprove(c *fiber.Ctx, myid uint64, req PostMessageRequest) error {
 	// Move to Approved with arrival=NOW() so immediate-email recipients get it.
 	// Guard against double-approve by requiring collection != Approved.
 	// Restrict to groups the caller is authorised for.
-	if result := db.Exec("UPDATE messages_groups SET collection = ?, approvedby = ?, approvedat = NOW(), arrival = NOW() WHERE msgid = ? AND groupid IN ? AND collection != ?",
+	// Clear contentcheck_reasons so IP-abuse and other warnings are not shown to
+	// moderators after a post has been manually approved.
+	if result := db.Exec("UPDATE messages_groups SET collection = ?, approvedby = ?, approvedat = NOW(), arrival = NOW(), contentcheck_reasons = NULL WHERE msgid = ? AND groupid IN ? AND collection != ?",
 		utils.COLLECTION_APPROVED, myid, req.ID, authorizedGroups, utils.COLLECTION_APPROVED); result.Error != nil {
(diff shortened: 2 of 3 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #792) — this fix did not pass and is not ready to merge.

- **Partial implementation**: No backfill for rows already in collection='Approved'. Both UPDATEs only fire on the Pending→Approved transition: message.go guards `WHERE ... AND collection != ?` (utils.COLLECTION_APPROVED) and AutoApproveService.php carries the documented `collection!='Approved'` guard. Already-approved messages 
- **Incomplete diff**: The diff covers PHP auto-approve (AutoApproveService) and Go manual-approve (apiv2 handleApprove) but the inline comments reference 'V1 approve()' — a V1 PHP iznik-server approve path that also sets collection='Approved' and is not modified here. Moderators in ModTools historically approve through t

edwh:
Closing — wrong layer, same pattern as #777: a **permanent per-request suppression** to mask **stale historical data**.

The stale `contentcheck_reasons` only exist on messages approved *before* the approval-time clear (commit `3bd5d19`) shipped — new approvals are already clean (forward path + `AutoApproveService`). So adding code that strips `ContentcheckReasons` from the GET response for every `Approved` group, on every request, indefinitely — to hide a fixed, non-growing set of pre-fix rows — taxes the hot path forever for a one-off data problem.

The right fix is a one-off DB cleanup:

```sql
UPDATE messages_groups
   SET contentcheck_reasons = NULL
 WHERE collection = 'Approved'
   AND contentcheck_reasons IS NOT NULL;
```

After that the stale warnings are gone at the source, `3bd5d19` keeps new approvals clean, and no per-request suppression is needed. Closing in favour of that cleanup.

## Discourse

https://discourse.ilovefreegle.org/t/9768/5

PR #793, opened 2026-06-18
fix(digest): wrap summary links in <div> for Outlook Classic line-break rendering

edwh:
## Root Cause

The \"In this digest\" summary index uses `\$summaryLink` CSS applied directly to `<a>` elements (including `display: block; margin-bottom: 4px`). Outlook Classic's Word rendering engine ignores `display:block` on inline elements like `<a>`, so all summary items ran together on the same line with no visible separation.

Affected template: `iznik-batch/resources/views/emails/mjml/digest/unified.blade.php` lines 239–257 (both the visible list and the `<details>` overflow block).

## Evidence

Before fix (origin/master, line 249):
```html
<a href="..." style="color: ...; text-decoration: none; display: block; margin-bottom: 4px;">Alpha</a>
<a href="..." style="color: ...; text-decoration: none; display: block; margin-bottom: 4px;">Bravo</a>
```
Outlook Classic renders these inline — items run together.

After fix:
```html
<div style="margin-bottom: 4px;"><a href="..." style="color: ...; text-decoration: none;">Alpha</a></div>
<div style="margin-bottom: 4px;"><a href="..." style="color: ...; text-decoration: none;">Bravo</a></div>
```
Outlook Classic respects `<div>` as a block element — each item on its own line.

Regression test asserts `/<div[^>]*>\s*<a\s[^>]*href=.../` — fails on master, passes on fix branch.

## Fix

- Removed `display: block; margin-bottom: 4px;` from `\$summaryLink` inline style (the `<a>` no longer carries the block layout)
- Wrapped each summary `<a>` in `<div style="margin-bottom: 4px;">` — Outlook Classic's Word engine treats `<div>` as block-level, creating a line break after each item
- Applied to both the visible summary section and the collapsed `<details>` overflow block
- Also fixes `summaryViewData()` fixture to include `postedToText`/`metaText`/`isOwnPost`/`accentColor` so direct view rendering covers the full MJML template

## Other Call Sites

The `\$summaryLink` pattern (bare `<a>` with `display:block`) is used only in `unified.blade.php` (two foreach loops). The AMP digest uses `<a class="summary-link">` with a stylesheet CSS class — AMP is never sent to Outlook Classic, so no change needed there. No other email templates use the same pattern.

## Test

`iznik-batch/tests/Unit/Mail/UnifiedDigestSummaryTest.php`
Test: `test_html_summary_links_wrapped_in_div_for_outlook_compatibility`
Command: `POST http://localhost:8081/api/tests/laravel {"filter": "UnifiedDigestSummaryTest"}`

Proves: renders the MJML blade directly (MJML content inside `<mj-text>` passes through verbatim), then asserts each visible summary link is preceded by `<div>`. Fails on master (bare `<a>`), passes on fix branch.

## Discourse

https://discourse.ilovefreegle.org/t/9363/31

### DIFF
--- a/iznik-batch/resources/views/emails/mjml/digest/unified.blade.php
+++ b/iznik-batch/resources/views/emails/mjml/digest/unified.blade.php
@@ -236,7 +236,7 @@
             $summaryPosts = collect($posts);
             $summaryVisible = \App\Mail\Digest\DigestStyle::SUMMARY_VISIBLE_LINES;
             $summaryHidden = $summaryPosts->slice($summaryVisible);
-            $summaryLink = 'color: ' . \App\Mail\Digest\DigestStyle::OFFER_GREEN . '; text-decoration: none; display: block; margin-bottom: 4px;';
+            $summaryLink = 'color: ' . \App\Mail\Digest\DigestStyle::OFFER_GREEN . '; text-decoration: none;';
         @endphp
         @if($summaryPosts->count() >= 2)
         <mj-section background-color="#ffffff" padding="16px 20px 4px">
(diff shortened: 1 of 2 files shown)


edwh:

Issue #32, opened 2026-01-25
feat: Unified digest - consolidate per-group emails into one


edwh:
## Summary
Consolidate email notifications from per-group digests into a single unified "Freegle Digest".

## Background
This work was planned in `plans/active/digestchanges.md` (feature/remove-email-wallpaper branch).

### 1. One Digest Instead of Many
Users receive a single "Freegle Digest" containing posts from all their communities, with duplicates removed.

### 2. Deduplicated Posts
Cross-posted items appear only once, with indication of which groups it's posted to.

### 3. Remove Email Frequency Popup
Based on analysis, 97% of users choose the daily default. Remove friction-adding popup.

### 4. Simplified Email Settings
One dropdown that applies to all communities instead of per-group settings.

### 5. New Subject Line Format
From: `[GroupName] What's New (3 messages)`
To: `5 new posts near you - Sofa, Coffee table...`

### 6. Unified Sender Address
Single noreply@ilovefreegle.org sender for better deliverability.

## Why This Matters
1. **Simplifying the codebase** - Per-group digest system is complex and hard to maintain
2. **Enabling rippling out** - Foundation for future algorithm showing relevant posts to more freeglers

## Branch
`feature/remove-email-wallpaper` in FreegleDocker contains the planning document.

## Plan Document
See `plans/active/digestchanges.md` for detailed implementation plan including:
- Database migration strategy
- Email template changes
- Settings page updates
- Rollout plan

edwh:
Implemented: UnifiedDigestService.php, UnifiedDigest.php mailable, SendUnifiedDigestCommand.php. Running in production.

Issue #31, opened 2026-01-25
fix: Remove email wallpaper background


edwh:
## Summary
Update email templates to use clean solid green (#338808) background instead of wallpaper image.

## Background
This work was started in the `feature/remove-email-wallpaper` branch and is ready for PR.

## Changes Made
- Delete unused `partials/header.blade.php` (contained wallpaper reference)
- Update `components/header.blade.php` with clean solid green background
- Remove `wallpaper_url` from `config/freegle.php`
- Update CLAUDE.md to remove wallpaper documentation

## Benefits
- Cleaner, more modern email design
- Reduces external image dependencies
- Faster email loading
- Better accessibility with solid background colors

## Files Changed
- `iznik-batch/config/freegle.php`
- `iznik-batch/resources/views/emails/mjml/components/header.blade.php`
- `iznik-batch/resources/views/emails/mjml/partials/header.blade.php` (deleted)

## Branch
`feature/remove-email-wallpaper` in FreegleDocker

edwh:
**Progress update (2026-04-14):** Email templates now use solid green background (no wallpaper). Stale wallpaper_url config entry remains in config/freegle.php (line 27) — can be removed to fully close this.

edwh:
This issue describes work that is in the FreegleDocker repository, not Iznik. The feature/remove-email-wallpaper branch exists but is far behind master. 

Suggest:
1. Create a fresh PR from master with the wallpaper removal changes (clean up MJML templates, remove wallpaper config, etc.)
2. This is blocking on code review and merge in FreegleDocker, not Iznik

Consider moving the issue to FreegleDocker if the work should be tracked there.

edwh:
Checked against master 2026-07-31. MJML pipeline is clean; four legacy templates and two config/doc references are not. Not closing.

**Done**
- No template under `resources/views/emails/mjml/` references the wallpaper.
- `partials/header.blade.php` is clean: solid `bg-success` section. Note the issue says to delete this file, which is now wrong - it is included by 10+ templates (`charity-signup`, `fbl/notification`, `housekeeper/results`, `session/*`, `admin/*`, `donation/*`, `newsfeed/*`). Deleting it would break them.

**Outstanding**

Four non-MJML templates still hardcode `https://www.ilovefreegle.org/wallpaper.png`:
- `resources/views/emails/noticeboard/thanks.blade.php` (lines 163, 166, 168) - live, sent via `App\Mail\Noticeboard\NoticeboardThankMail`
[...]

PR #794, opened 2026-06-18
fix(digest): wrap summary links in <div> for Outlook Classic line-break rendering

edwh:
## Root Cause

The "In this digest" summary index uses \`\$summaryLink\` CSS applied directly to `<a>` elements (including `display: block; margin-bottom: 4px`). Outlook Classic's Word rendering engine ignores `display:block` on inline elements like `<a>`, so all summary items ran together on the same line with no visible separation.

Affected template: `iznik-batch/resources/views/emails/mjml/digest/unified.blade.php` lines 239–257 (both the visible list and the `<details>` overflow block).

## Evidence

Before fix (origin/master, line 239/249):
```
$summaryLink = '... display: block; margin-bottom: 4px;';
<a href="..." style="{{ $summaryLink }}">Alpha</a>
<a href="..." style="{{ $summaryLink }}">Bravo</a>
```
Outlook Classic ignores `display:block` on `<a>` — items render inline, no line breaks.

After fix:
```
$summaryLink = '... text-decoration: none;';  // display:block removed
<div style="margin-bottom: 4px;"><a href="...">Alpha</a></div>
<div style="margin-bottom: 4px;"><a href="...">Bravo</a></div>
```
Outlook Classic respects `<div>` as block-level — each item on its own line.

Regression test regex `/<div[^>]*>\s*<a\s[^>]*href=.../` fails on master, passes on this branch.

## Fix

- Removed `display: block; margin-bottom: 4px;` from `\$summaryLink` inline `<a>` style
- Wrapped each summary `<a>` in `<div style="margin-bottom: 4px;">` — Outlook Classic's Word engine treats `<div>` as block-level
- Applied to both the visible summary section and the `<details>` overflow block
- Adds missing `postedToText`/`metaText`/`isOwnPost`/`accentColor` keys to `summaryViewData()` so direct view rendering covers the full MJML template

## Other Call Sites

The bare-`<a>`-with-`display:block` pattern existed only in `unified.blade.php` (two foreach loops). The AMP digest uses a CSS class — AMP is never sent to Outlook Classic, no change needed. No other email templates use the same pattern.

## Test

`iznik-batch/tests/Unit/Mail/UnifiedDigestSummaryTest.php`
Test: `test_html_summary_links_wrapped_in_div_for_outlook_compatibility`
Command: `POST http://localhost:8081/api/tests/laravel {"filter": "UnifiedDigestSummaryTest"}`
Proves: renders the MJML blade directly (content inside `<mj-text>` passes through verbatim), asserts each visible summary link is preceded by `<div>`. Fails before fix, passes after.

## Discourse

https://discourse.ilovefreegle.org/t/9363/31


edwh:
⚠️ **Unverified — needs the screenshot before merge.** This was written from the text report alone.

Topic 9363 ends at post #32, which is *"Screenshot please."* — Edward's reply to the report (post #31). No before/after image of the broken rendering was ever provided.

Two things the evidence would settle:

1. **Which client.** The reporter says *"In M365 Outlook the summary doesn't display properly … seems OK in other clients, including New Outlook."* Microsoft 365 ships the **classic desktop** Outlook (Word engine) *alongside* New Outlook, so "M365 Outlook" here is most likely the classic desktop — which would match this PR's `display:block`-ignored theory. But that's inference, not confirmation, and the Word-engine behaviour is specific to that client.
2. **Whether the fix works.** Wrapping `<a>` in `<div>` is standard Outlook email practice and plausible, but there's no proof it renders correctly in the actual affected client.

The "## Evidence" section is code + a regex assertion — not evidence of the Outlook behaviour or the reporter's actual problem. Recommend holding until the requested screenshot confirms both the rendering and the client.

## Discourse

https://discourse.ilovefreegle.org/t/9363/31

PR #795, opened 2026-06-18
fix(contentcheck): use item name for subject-repeat length guard (V1 parity)

edwh:
## Root Cause

`checkSubjectRepeat` in `ContentCheckService` used the **full message subject** (`"Offer: Test"` = 11 chars) for its `< 10` char skip-guard. Common test subjects like `"Offer: Test"` exceed the threshold, so they accumulate across groups whenever mods/testers post. Once ≥ 30 distinct groups have received that exact subject in the past 7 days, every new `"Offer: Test"` post is falsely flagged and left Pending.

V1's `Spam.php` used `getPrunedSubject()` (item name only, stripping type prefix) for the length check — `"Test"` is 4 chars < 10 → always skipped. The Laravel port checked the full subject instead, losing that protection.

## Evidence

**Bug path** — calling without `$itemName` (as the old code did) with 30+ groups, `"Offer: Test"` (11 chars) passes the length guard and the repeat check fires:
```
$buggyPathResult = $this->service->checkSubjectRepeat($subject, $newMsgId);
assertNotNull($buggyPathResult);  // PASSES — confirms the bug is real
assertEquals('SubjectRepeat', $buggyPathResult['check']);
```

**Fix path** — calling with `$itemName='Test'` (4 chars < 10), the length guard fires immediately:
```
$fixedPathResult = $this->service->checkSubjectRepeat($subject, $newMsgId, 'Test');
assertNull($fixedPathResult);  // PASSES — legitimate mod test-post not blocked
```

## Fix

`checkSubjectRepeat` accepts an optional `$itemName` parameter. When provided, it guards on item name length instead of the full subject. `checkMessage` passes `$itemName` (already fetched for other checks). Short item names like `"Test"` (4 chars < 10) skip the repeat check entirely, matching V1 behaviour.

**How this differs from the rejected approach**: The original PR #795 test only asserted the fixed path — the item-name guard fired immediately so the 30-group DB setup was never exercised and the bug was never proven to exist. This re-attempt adds a `assertNotNull` on the OLD code path (no `$itemName`) which proves the 30-group state actually triggers the flag, before asserting the fix prevents it.

## Other Call Sites

`checkSubjectRepeat` is only called from `checkMessage` (line 151). No other call sites.

## Test

**File**: `iznik-batch/tests/Feature/Message/ContentCheckTest.php`  
**Method**: `test_subject_repeat_not_flagged_for_short_item_name_test_post`  
**Command**: `curl -X POST http://localhost:8081/api/tests/laravel -d '{"filter":"ContentCheckTest"}'`  
**Proves**: fail-before (buggyPathResult assertNotNull) → pass-after (fixedPathResult assertNull); all 149 ContentCheckTest tests pass.

## Discourse

https://discourse.ilovefreegle.org/t/9788/28

### DIFF
--- a/iznik-batch/app/Services/ContentCheckService.php
+++ b/iznik-batch/app/Services/ContentCheckService.php
@@ -148,7 +148,7 @@ class ContentCheckService
         if ($r = $this->checkLanguage($subject, $textbody)) {
             $reasons[] = $r;
         }
-        if ($r = $this->checkSubjectRepeat($subject, $msgid)) {
+        if ($r = $this->checkSubjectRepeat($subject, $msgid, $itemName)) {
             $reasons[] = $r;
         }
         if ($r = $this->checkKnownSpammer($textbody)) {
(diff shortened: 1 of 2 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #795) — this fix did not pass and is not ready to merge.

- **Test proves nothing / reproduction not demonstrated**: test_subject_repeat_not_flagged_for_short_item_name_test_post does not prove the bug existed before the fix. With the fix applied, the new `strlen($textToCheck) < 10` guard fires immediately on the item name 'Test' (4 chars) and returns null BEFORE any repeat-counting query runs — so the entire 30-g

PR #801, opened 2026-06-18
fix(chitchat): reserve right-side space for thread-menu button (Discourse #9749)

edwh:
Re-attempt of #785 (closed due to force-push invalidating the PR state; branch is the same).

## Root Cause

The thread-level "…" dropdown in `NewsThread.vue` was positioned with `float-end` (CSS `float: inline-end`), relying on float-layout to push adjacent text left. Float-shortening only affects line-boxes *vertically adjacent* to the float element. For post types whose preceding content (e.g. a CommunityEvent type-badge at ~28 px) is shorter than the button (~36 px), the title `<h3>` text begins at the same vertical level as the float and is not shortened — the text runs underneath the button.

## Evidence

Discourse post https://discourse.ilovefreegle.org/t/9749/5 shows a ChitChat thread where a long first line of text overlaps the "…" button because the float mechanism does not reserve space on that line.

## Fix

Replace float-based positioning with an absolutely-positioned button inside a `position: relative; padding-right: 3rem` wrapper:

- Added `class="news-component-wrapper"` to the outer `<div v-if="isNewsComponent">`.
- Removed `float-end` from the `<b-dropdown>` → `class="thread-menu"`.
- CSS:
  ```scss
  .news-component-wrapper {
    position: relative;
    padding-right: 3rem;
  }
  .thread-menu {
    position: absolute;
    top: 0;
    right: 0;
  }
  ```

The `padding-right` reserve applies to every line of content regardless of vertical position, so no text can reach the button area.

## Other Call Sites

`NewsReply.vue` already uses `position: absolute; top: -0.25rem; right: -1rem` for its reply menu and is unaffected.

## Test

Added to `NewsThread.spec.js` (inside `describe('dropdown menu', ...)`):

```js
it('thread-menu is not floated and wrapper reserves right-side space for the button (Discourse #9749)', async () => {
  const wrapper = await createWrapper()
  const dropdown = wrapper.find('.b-dropdown')
  expect(dropdown.classes()).not.toContain('float-end')
  expect(wrapper.find('.news-component-wrapper').exists()).toBe(true)
})
```

AssertFlip confirmed: step 3a passed on buggy code (`.toContain('float-end')`), step 3b failed on buggy code, passes after fix. Full suite: 102 tests pass.

## Discourse

https://discourse.ilovefreegle.org/t/9749/5

### DIFF
--- a/iznik-nuxt3/components/NewsThread.vue
+++ b/iznik-nuxt3/components/NewsThread.vue
@@ -13,10 +13,10 @@
             <span v-else>the system</span>
             and is only visible to volunteers and the person who posted it.
           </div>
-          <div v-if="isNewsComponent">
+          <div v-if="isNewsComponent" class="news-component-wrapper">
             <b-dropdown
               lazy
-              class="float-end thread-menu"
+              class="thread-menu"
               right
               variant="link"
               no-caret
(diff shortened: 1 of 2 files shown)


edwh:
Closing in favour of #798 — same Discourse issue (9749), but this targets the wrong component.

I verified on dev-live against real ChitChat data: the **top-level** posts this PR changes (`NewsThread`) do **not** collide — the "…" sits on the header row (username · time · location) with the body below, so there's already clear space. The actual *"mostly good but still the odd one"* from post #5 is on the **replies** (the inline *name + text* layout), which is what #798 fixes (`NewsReply` → `.reply-content` padding), browser-verified (PacificTrashVortex/Jos replies now wrap before the dots).

This PR also re-uses the closed #785 branch and applies `padding-right: 3rem` to every top-level post body — visibly narrowing them — for a collision that isn't occurring, and wasn't verified in a browser. Closing; #798 is the verified fix.

## Discourse

https://discourse.ilovefreegle.org/t/9749/5

PR #814, opened 2026-06-18
Multi-Group Messages

fnnbrr:
## TODO

- Run the migration to make `messages_spatial` per-group manually on prod.
  - See [iznik-batch/database/migrations/2026_06_17_000001_make_messages_spatial_per_group.php](https://github.com/Freegle/Iznik/blob/messages-multiple-groups/iznik-batch/database/migrations/2026_06_17_000001_make_messages_spatial_per_group.php)


edwh:
@fnnbrr I've merged the mahoosive rippling out changes into master, ready for deploy next week.  This (sorry) overlaps with this PR.  You've got some stuff in here I needed to have, which is good timing.

One specific: I want to push the other way on `groupid` in `messages_spatial`, and remove it.  Having the groupid feels wrong since:
* the spatial table is supposed to be about the message data
* per-group status lives in `messages_groups`.  

Looking at the places that rely on the existing `groupid` I think they should be using JOINs with `messages_groups`, and each of them highlights somewhere that needs attention for multiple groups.

So I've created a PR which adapts 814 along these lines: https://github.com/Freegle/Iznik/pull/827.  Can you see whether you can pick up that one?  I've not merged it into master till you've had a look but it does pass CI.

PR #844, opened 2026-06-22
feat(modtools): per-post rippling reach map ("Who can see this?")

edwh:
## Summary

Adds a moderator-facing **"Who can see this?"** button on OFFER/WANTED posts that opens a map showing the area a post can currently be seen in, based on its rippling-out reach. This closes a gap where the reach was computed/stored but never surfaced to moderators.

- **Go API** — `GET /message/:id/reach` returns the post's current `rippling_reach` polygon as GeoJSON, plus the schedule cursor (`tick`/`totalticks`), `status`, and `nextexpansionat`. Mod-of-group (or Admin/Support) only; returns `rippling:false` (not 404) when the post isn't rippling yet, so the UI can say so cleanly.
- **ModTools** — `ModMessageReachMap` modal renders the reach polygon + origin marker on a Leaflet map (reusing the existing `l-geo-json`/`useMap` pattern). It shows "Step N of M", the status, and **"Next ripple out due in ~X hours"**.
  - We don't (cheaply) compute the *shape* of the next reach, but the *timing* is known (`next_expansion_at`), so we surface that.
- Wired into `ModMessage` via a small "Who can see this?" link, shown only for OFFER/WANTED.

## Code Quality Review

- Reuses existing conventions: `IsModOfGroup`/`IsAdminOrSupport` for auth (mirrors other message endpoints), `useOurModal` for the modal (mirrors `RipplingExplanationModal`), `l-geo-json` + `useMap` for the map (mirrors `MessageMap`/`ModGroupPostVisibility`).
- GeoJSON is produced server-side via `ST_AsGeoJSON` and passed straight to Leaflet; raw column names are selected so gorm's snake_case→struct mapping binds correctly.
- No new dependencies.

## Future Improvements

- Once a cheap "next tick" polygon is available, draw the next reach as a lighter faded overlay (the timing is already shown).

## Test Plan

- **Go** (full suite green, 3224 passed / 0 failed): new `TestMessageReachAsMod` (mod sees Polygon GeoJSON + tick), `TestMessageReachForbiddenForNonMod` (403), `TestMessageReachNoReachRow` (`rippling:false`).
- **Component specs** (5 passed): not-rippling notice, map + details render, next-due timing label, fully-rippled state, show/hide.
- **`ModMessage` spec**: 80 passed (added a stub for the new child).
- **Live smoke**: endpoint serves on apiv2-live against the production DB (401 without auth, routes correctly).

### DIFF
--- a/docs/superpowers/plans/2026-06-22-digest-rippling-score-ordering.md
+++ b/docs/superpowers/plans/2026-06-22-digest-rippling-score-ordering.md
@@ -1,0 +1,867 @@
+# Rippling score-ordering in the unified digest — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Order the unified daily digest's live posts by the `/rippling` digest-preview score (closeness + freshness + underexposure + home-group anchor) instead of strictly by arrival time.
+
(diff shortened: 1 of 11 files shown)


edwh:
Superseded by #845. The static reach-map approach here was replaced (per review) with a full-screen explorer-based reach modal, plus the related rippling member/mod gap closures. The unused `GET /message/:id/reach` endpoint introduced here was removed in #845 (the modal computes reach client-side via the routing server).

PR #860, opened 2026-06-23
fix(rippling-explorer): correct misleading TN tooltip

edwh:
## Root Cause

The "~X would be notified" sidebar bar in the rippling explorer always showed an italic note: _"TrashNothing members use a separate algorithm"_. This is misleading because:

- TN members who are also Freegle users **are** counted in the notification figure via the same rippling algorithm as everyone else (they appear in `users_approxlocs`)
- The actual TN-specific behaviour (TN-originated posts with `tnpostid` are not rippled into new groups) relates to **post sources**, not notification recipients - irrelevant to this count bar
- The note was shown for **all** posts, not just TN-originated ones, adding confusion with no benefit

## Evidence

Test failure on buggy code (Step 3b):

```
FAIL  tests/unit/composables/rippling-freegler-bar.spec.js > rippling/freeglerBar
       > does not mention TrashNothing or a separate algorithm (misleading tooltip fixed)
AssertionError: expected '...' not to contain 'TrashNothing'
     35|     expect(html).not.toContain('TrashNothing')
```

## Fix

- Extract the bar's HTML-building logic to `modtools/composables/rippling/freeglerBar.js` (pure function, no DOM/closure dependencies, matches the pattern of `pie.js` / `scoring.js`)
- Remove the misleading note from the generated HTML
- `renderFreeglerBar` in `setupRipplingExplorer.js` now delegates to `buildFreeglerBarHTML`

## Other Call Sites

No other user-visible TrashNothing references found in source (`grep -rn "TrashNothing" --include="*.js" --include="*.vue"` returns only the code comment at line 439 and the new test).

## Test

**File**: `tests/unit/composables/rippling-freegler-bar.spec.js`  
**Command**: `npx vitest run tests/unit/composables/rippling-freegler-bar.spec.js`  
**Proves**: 7 tests; the "does not mention TrashNothing" test failed before the fix and passes after.

## Discourse

https://discourse.ilovefreegle.org/t/9808/56

### DIFF
--- a/iznik-nuxt3/modtools/composables/rippling/freeglerBar.js
+++ b/iznik-nuxt3/modtools/composables/rippling/freeglerBar.js
@@ -1,0 +1,21 @@
+// Builds the HTML for the "~X would be notified" sidebar bar.
+// Returns null when the bar should be hidden (zero estimate or no located total).
+export function buildFreeglerBarHTML(
+  estimatedInsideLocated,
+  totalLocated,
+  unlocatedFraction
+) {
+  if (!(estimatedInsideLocated > 0 && totalLocated > 0)) return null
+  const totalEstimate = Math.round(
+    estimatedInsideLocated / (1 - unlocatedFraction)
+  )
+  const unlocatedShare = totalEstimate - estimatedInsideLocated
+  return (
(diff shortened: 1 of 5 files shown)


edwh:
Closing per maintainer review (@edwh): this is a bad fix and should not be merged. If the underlying "misleading TN tooltip" issue (Discourse 9808/56) still needs addressing, it should be superseded by a corrected approach rather than merged as-is.

PR #874, opened 2026-06-24
fix(ripple): only block re-rippling for groups rippled-into then left

edwh:
## Problem

The rippling expander blocked rippling a poster into **any** group they had a `Group/Left` log for. Because that guard fired on *any* prior departure - a manual leave, mod-remove, unban, "moved house" - a single past leave **permanently** barred that user's future posts from ever rippling into the group again, even though the leave had nothing to do with rippling.

The intended rule is narrower: only a group the poster was **rippled into** (a `Group/Joined` log with `text='Rippled'`) and then **left** should count as an opt-out. An ordinary membership they joined normally and later left must **not** block rippling.

## Why the marker exists

We always join-then-post, so every rippled membership writes a distinct join log: `addPosterMembershipToRippledGroups` records `logs(type='Group', subtype='Joined', text='Rippled')`. That marker is exactly what tells a rippling auto-join apart from a manual/auto one.

## Fix — "most recent join wins"

All three `Group/Left` guards in `ExpandService` now suppress **only** when the poster's **most recent** `Group/Joined` log for the group is a ripple-join (`text='Rippled'`) **and** they then left it (a `Left` after it). Concretely the predicate requires:
- a `Joined/Rippled` log `lj`, **and**
- `NOT EXISTS` any later `Joined` log (so `lj` is the latest join → "most recent join wins"), **and**
- `EXISTS` a `Left` with `id > lj.id` (they left after that ripple-join).

| Site | Function | Effect |
|---|---|---|
| A | `rippleIntoNewGroups` | whether the **post** ripples into the group |
| B | `addPosterMembershipToRippledGroups` | whether the **membership** is added |
| C | `pullRippledPostsFromLeftGroups` | whether a rippled post is **pulled** when the poster left |

Site C is kept consistent on purpose: if only A+B were relaxed, a post would ripple into a group the user *normally* left and then C would immediately pull it back out (churn).

No changes needed in Go (`iznik-server-go` writes `Left` logs but never reads them to gate rippling) or V1 PHP. Confirmed by a full-repo audit.

## Behaviour after the fix

| Scenario (oldest→newest) | Before | After |
|---|---|---|
| No prior history | ripples | ripples |
| Manual/auto join → left | **blocked** | **ripples (bug fixed)** |
| Rippled in → left | blocked | blocked (the opt-out) |
| Rippled in → still a member | ripples | ripples |
| **Rippled in → left → manually rejoined → left again** | blocked | **ripples (most-recent-join-wins)** |

The last row is the refinement: because the latest join was manual, the earlier rippled opt-out no longer applies.

## Tests (`ExpandServiceTest`, 36 passed locally)

- `test_rippled_in_then_left_is_not_rejoined_and_post_not_rippled_in` (genuine opt-out → suppressed)
- `test_ordinary_left_does_not_block_rippling` (manual/auto leave → ripples)
- `test_manual_rejoin_after_rippled_left_allows_rippling` (most-recent-join-wins)
- `test_ordinary_left_does_not_pull_a_freshly_rippled_post` (no pull-churn; also covers a Left predating the ripple-join)
- Existing `test_leaving_a_rippled_group_pulls_the_post` still passes

**Verification:** an adversarial sweep of 42 log-history scenarios (multi-cycle, boundary, NULL-text, id-ordering) found **0 divergences** between the SQL and the most-recent-join-wins intent.

## Index/perf

The new correlated subqueries filter `logs` by `user = ? AND groupid = ?` first - the same access pattern the original guard used - served by the existing `user`/`group` indexes; `type`/`subtype`/`text` are cheap residual filters on a user's handful of per-group rows. Runs in the batch expander, not a request path. No new index or migration.

### DIFF
--- a/iznik-batch/app/Services/Ripple/ExpandService.php
+++ b/iznik-batch/app/Services/Ripple/ExpandService.php
@@ -421,9 +421,22 @@ class ExpandService
                        SELECT 1 FROM messages_groups mg WHERE mg.msgid = ? AND mg.groupid = g.id
                    )
                    AND NOT EXISTS (
-                       SELECT 1 FROM logs l
-                       WHERE l.user = m.fromuser AND l.groupid = g.id
-                         AND l.type = 'Group' AND l.subtype = 'Left'
+                       -- Only a group the poster was RIPPLED into (a Group/Joined log with
+                       -- text='Rippled') and then LEFT (a later Group/Left log) is barred from
+                       -- re-rippling - that pairing is the poster opting out of the ripple. A group
+                       -- they joined manually/ordinarily and later left must NOT block rippling.
+                       -- ll.id > lj.id requires the leave to post-date the ripple-join, so a group
+                       -- left long before being rippled in is not treated as opted-out. Sites B/C
+                       -- apply the identical rule.
+                       SELECT 1 FROM logs lj
+                       WHERE lj.user = m.fromuser AND lj.groupid = g.id
+                         AND lj.type = 'Group' AND lj.subtype = 'Joined' AND lj.text = 'Rippled'
+                         AND EXISTS (
+                             SELECT 1 FROM logs ll
+                             WHERE ll.user = lj.user AND ll.groupid = lj.groupid
+                               AND ll.type = 'Group' AND ll.subtype = 'Left'
+                               AND ll.id > lj.id
+                         )
                    )",
                 [$msgid, $msgid, $reachWkt, $msgid]
             );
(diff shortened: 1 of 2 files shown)


edwh:
**Index/perf check (resolves the only open question):** the new correlated subqueries filter `logs` by `user = ? AND groupid = ?` first, which is the *same* access pattern the original `Group/Left` guard already used in production. The `logs` table has single-column indexes on `user` and `group`(groupid) (plus composites on `(type,subtype)` and `(timestamp,type,subtype)`); `type`/`subtype`/`text='Rippled'` are cheap residual filters on the handful of rows a user has for a given group. `text` is `mediumText` (unindexed) but only ever evaluated on those already-narrowed rows. This runs in the batch expander, not a request path. **No new index or migration is required.**

PR #884, opened 2026-06-25
fix(rippling): one repost reminder & chase-up per item, not per group

edwh:
## Problem

Rippling-out puts a single post on many groups (`messages_groups.rippled_in = 1`). Auto-reposting runs per group, which is correct for the *repost* (it keeps the item fresh in each community), but the poster-facing **notification emails** were not deduplicated across groups:

- **Auto-repost warning ("Will Repost: …")** — `AutoRepostService` stamped `lastautopostwarning` per group (`WHERE msgid = ? AND groupid = ?`), so a widely-rippled item could email the poster once per group, per cycle. Because rippled-in rows get a fresh `arrival` as the reach expands over hours/days, their repost windows are *staggered*, so even a simple "stamp all groups" 24h dedup would still leak repeat emails days apart.
- **Chase-up ("What happened to: …")** — `ChaseUpService` already stamped `lastchaseup` on every group, but shared the same staggered-window hole (a rippled row reaching max-reposts later could re-chase).

## Key insight

Every action on these emails (`…/mypost/{msgid}/completed|withdraw|promise`) is recorded against the **msgid**, and both the auto-repost and chase-up candidate queries exclude any message that has an outcome/promise. So one response already stops reposting and chasing on **every** group — one email per item loses nothing.

## Fix

Anchor both poster-facing notifications to the post's **home posting** (`rippled_in = 0`), leaving the repost itself per-group.

- **Warning** — only a `rippled_in = 0` row sends it; rippled rows still repost silently. When sent, stamp `lastautopostwarning` across **every** group of the message (revert to V1's `WHERE msgid = ?`).
- **Chase-up** — only a `rippled_in = 0` row may initiate it; keep the existing all-groups `lastchaseup` stamp. Closes the staggered-window hole.
- **Reposting stays per-group** — unchanged (`test_multiple_groups_each_repost_independently` preserved).

## Docs — deep-dive accuracy pass

Verified every rippling claim in both guides against the code and corrected:

**Moderators:**
- Rippled-in posts are **approved on arrival by default** (`rippled_in_pending_hours = 0`), not "~1h in pending"; documented the optional mod-veto window.
- Reach button label corrected to **"View rippling reach"**.
- Edit warning corrected to include the group count.
- **TrashNothing** is currently **excluded** from rippling (was: "behaves exactly like any other post").
- Held replies also release when a post **finishes spreading** (not only when it reaches the member).
- New "repost reminders & chase-ups happen once per item" section.

**Members:**
- Reply-saturation note (a post stops spreading once enough people nearby reply).
- "Reminders about your post" once-per-item section + summary-table row.

(The audit also flagged the "What's New" subject-cap as reverted — that was the main checkout's unrelated uncommitted WIP; this branch, from committed HEAD, has the fix, so the docs are accurate here.)

## Tests

4 new (all green in a filtered Unit run — 81 cases, 0 failed):
- `AutoRepostServiceTest::test_warning_sent_once_across_rippled_groups`
- `AutoRepostServiceTest::test_rippled_in_row_reposts_without_warning`
- `ChaseUpServiceTest::test_rippled_item_chased_up_once_from_home`
- `ChaseUpServiceTest::test_rippled_only_item_is_not_chased_up`

Design spec: `docs/superpowers/specs/2026-06-25-repost-reminder-dedup-rippling-design.md`

### DIFF
--- a/RIPPLING-OUT-FOR-MEMBERS.md
+++ b/RIPPLING-OUT-FOR-MEMBERS.md
@@ -15,6 +15,9 @@ The aim is to keep freegling as local as possible. Your neighbours get first cha
 which means less driving and a fairer go for everyone close by. If nobody nearby wants
 it, it reaches further afield automatically - you do not need to do anything extra.
 
+If enough people nearby reply to your post, it stops spreading further - it already has
+plenty of interest, so there is no need to show it to people further away.
+
 ---
 
 ## What happens when your post ripples into a new community?
@@ -78,6 +81,17 @@ count. Nothing is missing - everything is always there on the browse page.
 
 ---
 
+## Reminders about your post
+
+If your item is still available after a while, we remind you shortly before we
+automatically repost it ("Will Repost: ..."), and later we check in to ask what happened
+("What happened to: ..."). Even though your post may be live in several communities at
+once through rippling, you only ever get **one** reminder and **one** check-in each time -
+not one per community. Whatever you choose - mark it taken, withdraw it, or promise it to
+someone - applies to your post everywhere it has reached, so a single tap settles it.
+
+---
+
 ## Staying in control
 
 ### Changing your email settings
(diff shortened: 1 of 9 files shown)


edwh:
Cherry-picked `879494a62` (already on master) onto this branch to fix CI: the job-click attribution change (`bfb31b29a`) added a `link` field to `jobStore.log()`, but `JobOne.spec.js`/`JobMosaicTile.spec.js` here still asserted the old `{ id }` signature and failed the frontend unit suite. This branch was cut before the spec fix landed on master. No functional change.

PR #907, opened 2026-06-26
fix(modtools): hide a group's work queue from backup mods (explicit-groupid path)

edwh:
## Problem
A moderator (Torbrexbones / Derek, Fife) reported seeing the **Edits** queue in ModTools for groups he is only a **backup** mod on — "never used to be the case".

## Root cause
`GET /apiv2/modtools/messages?collection=Edit` resolves groups two ways in `ListMessagesMT`:
- **all-groups (`groupid=0`)** → `GetActiveModGroupIDs`, which excludes backup mods (`settings.active=0`). **Safe.**
- **explicit `groupid`** → `IsModOfGroup`, a **role-only** check with **no active filter**. A backup mod still has role `Moderator`/`Owner`, so they pass and get that group's edit-review queue.

The ModTools group selector (`ModGroupSelect`, `remember="edits"`) lists backup groups and re-sends a remembered backup group's id on mount, hitting the explicit-groupid path. Verified on prod: Derek is `Owner` with `active:0` on `freegle_seleicestershire` (253451), which had a live `reviewrequired=1` edit returned to him. Prod apiv2 = current master tip, so this is a master bug, not a stale deploy. The work-count badges (`session.go`/`groupWork.go`) were already correct — they use `isActiveModForGroup`.

## Fix
- **`user.IsActiveModOfGroup(uid, gid)`** — role **and** active, with the Admin/Support shortcut `IsModOfGroup` has.
- **`ListMessagesMT` + `ListMessages`** — the explicit-`groupid` work-collection path returns an **empty list** for backup mods (graceful, not 403), matching the `groupid=0` path and the badges. Applies to all work collections (Edit/Pending/Spam/Rejected).
- **`GetActiveModGroupIDs`** gains the legacy **`showmessages`** fallback via a shared `activeModSettingsSQL` predicate, matching `isActiveModForGroup` — closing a secondary gap where a mod marked backup via `showmessages=0` (16 prod memberships) leaked via the all-groups path too.

## Testing
`modtools_edits_backup_test.go`: an active mod **sees** the edit (explicit groupid + all-groups); a backup mod — via **`active=0`** and legacy **`showmessages=0`** — **does not**, on both routes. **Full Go suite passes locally.**

https://claude.ai/code/session_012NjYtwJBGymZyZgZbc59oS

### DIFF
--- a/iznik-server-go/message/message_list.go
+++ b/iznik-server-go/message/message_list.go
@@ -107,6 +107,12 @@ func ListMessages(c *fiber.Ctx) error {
 			if !user.IsModOfGroup(myid, groupid) {
 				return fiber.NewError(fiber.StatusForbidden, "Not a moderator for this group")
 			}
+			// Backup mods (settings.active=0, or legacy showmessages=0) have stepped back and must
+			// not see a group's work queue (Pending/Spam/Edit/Rejected) even via an explicit
+			// groupid - mirrors the groupid=0 path and the work-count badges.
+			if !user.IsActiveModOfGroup(myid, groupid) {
+				return c.JSON(ListMessagesResponse{Messages: []ListMessageItem{}})
+			}
 		}
 		groupIDs = []uint64{groupid}
 	}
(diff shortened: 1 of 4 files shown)


edwh:
Closing without merge — the diagnosis here was wrong, and the real fix has landed on master as `6f56ac259`.

## Why this was a misdiagnosis
This PR attributed Derek/Torbrexbones (Fife) seeing the Edit queue for unexpected groups to the **explicit-groupid path** letting **backup** mods through (`IsModOfGroup` being role-only). Re-checking prod Loki disproved that: **22 of his 23** `collection=Edit` requests (25-26 Jun) used `groupid=0` — the all-groups path, which already excludes backup groups via `GetActiveModGroupIDs` (`settings.active != 0`). The single explicit request was for a group he's **active** on. The "backup leak" this PR reproduced (`groupid=253451`) was only reachable by manually hitting that groupid — not via the path he actually uses.

## The real cause — rippling
A post rippled **into** a group gets an Approved `messages_groups` row there with `rippled_in=1`. The Edit-collection query in `ListMessagesMT` joins `messages_edits -> messages_groups` by msgid with **no `rippled_in` filter**, so an edit on a rippled-in post surfaced in **every receiving group's** Edit queue — including for **active** mods via the all-groups path — even though the edit belongs to the post's **origin** group only. New since rippling added `rippled_in` (migration `2026_06_18_000004`), hence "never used to be the case". Same bug class as the IP-abuse fix (`4b6d7b3c3`, `WHERE rippled_in=0`).

## Fix on master (`6f56ac259`)
One line: `AND mg.rippled_in = 0` on the Edit query's `messages_groups` join only (NOT Pending — rippled-in veto-window rows are meant to show there). Plus a regression test: origin-group mod sees the edit (all-groups + explicit groupid); a mod active only on the rippled-into group does not, on both routes. Full Go suite green (3287✓).

## Left open deliberately
The **secondary** gap this PR did find is real: on the **explicit-groupid** path, `IsModOfGroup` is role-only, so a backup mod can pull up a specific group's work queue if they request that groupid directly. It is **not** Derek's issue and isn't addressed by the master fix. If we want to close it, it should be a separate, clearly-scoped PR rather than riding on the rippling fix.

PR #956, opened 2026-07-04
feat(digest): rank daily digest posts by personal relevance (flagged, with holdout)

edwh:
Merge after #947 — this branch is based on it so its CI carries the srvx package-lock fix. It is otherwise independent of the similar-posts and wanted-match work.

## What this does

Personal relevance becomes a sort-ranking signal inside the existing daily digest rather than a separate email, covering the job of the standalone V1 "Relevant" mail.

**`DigestRelevanceService`** ranks a member's daily-digest posts by vector similarity to what they care about: the subject embeddings of their own posts from the last `OWN_POST_DAYS` (60) days and the posts they viewed in the last `VIEW_DAYS` (30), newest first and capped at `MAX_INTERESTS` (40) vectors. A post scores its best cosine against any interest vector. It reuses the same `nomic-embed-text-v1.5` 256-dim embeddings as search, decoded in PHP with `unpack('g')` against the Go writer's little-endian encoding.

**`UnifiedDigestService::sendDigestToUser`** calls `rank()` after scoring and distance filtering, and before dedup — so the ranked order decides which post dedup keeps as the representative, and what `metadata.post_msgids` records. That is the field the digest click-by-position dashboard reads, so measurement needs no new tracking.

### What ranking deliberately leaves alone

The digest already makes three ordering promises that relevance must not quietly undo.

- **The two-nearest pin.** `pinClosestTwo` puts the two closest posts at the top to answer "I keep seeing posts far away", choosing only among posts the recipient has not already seen. It marks those posts `_pinned`, and `rank()` holds them at the front and reorders only the tail.
- **The seen-post sink.** Already-seen posts are demoted by `seen_penalty`. A post the member viewed within `VIEW_DAYS` is also one of their interest vectors, so it would self-match at cosine 1.0 and retake the top slot. `rank()` passes the digest's own msgids to `interests()`, which excludes them, so no post is its own interest vector.
- **Which posts appear at all.** `UnifiedDigest` truncates the prepared list at `DigestStyle::DIGEST_POST_CAP` (65), so in a longer digest the order decides what is shown, not merely where. A post with no embedding is therefore treated as unknown relevance rather than low relevance: it holds its position while the embedded posts are permuted among the slots they already occupied. Sinking it would delete it from a long digest for want of an embedding — content loss driven by pipeline coverage, falling hardest on the newest posts.

### Flag, holdout and measurement

Gated on `FEATURE_DIGEST_RELEVANCE`, default off, since it changes email ordering. A 10% holdout (`userid % 10 == 0`) always keeps the arrival order as a control. With the flag off, in the holdout, or with no interest signal, `rank()` returns its input unchanged and the digest is byte-identical to the unranked one.

`rank()` reports through `didRank()` whether it actually reordered, and `UnifiedDigest` records that as `metadata.relevance_ranked`. The dashboard's ranked arm filters on that recorded value rather than on "not in the holdout", which would also sweep in every member who had no interest signal: they received an identical unranked digest and would drag the measured effect toward zero. This mirrors the recommendations funnel, which derives its arms from recorded impressions. `ModSysAdminDigestClicks` and the Go `emailtracking` digest-positions endpoint take the resulting `cohort=ranked|holdout` filter.

### Send cost

`interests()` runs once per recipient. It filters `messages_likes` on `userid`, `type` and `timestamp`, which the existing indexes cannot serve efficiently — `messages_likes_source_ts_user` leads on `source`, which this query does not constrain — so MySQL reads every like row a member has ever had to keep a few dozen. A `messages_likes (userid, type, timestamp)` index lets it stop early. `messages_likes` is ~75M rows and hot and production is Galera under TOI, so the migration ships with a guarded companion `_migration.sql`, matching how the other `messages_likes` index is deployed, for node-by-node application under RSU.

Decoded embeddings are also memoised across recipients for the life of a run, since neighbouring members see largely the same posts, with misses cached so an unembedded post is not re-queried per recipient. The memo is bounded and flushed at the start of each run.

### V1 "Relevant" email

The V1 tree, including `scripts/cron/relevant.php` (the "Any of these take your fancy?" mail) and `include/mail/Relevant.php`, is already gone from master, so there is no code here to remove. `MIGRATION-STATUS.md` marks that job retired and records where its purpose is served instead, which closes the last non-search dependency on the `messages_index` keyword index that Phase 5 of the rollout retires. No deploy action.

## Decision criterion

After four weeks with `FEATURE_DIGEST_RELEVANCE` on: keep the ranking if the ranked cohort's overall digest CTR is at least the holdout's, the expected signature being a steeper position curve with clicks concentrated at the top.

## Tests

- `DigestRelevanceServiceTest` covers the Go↔PHP embedding blob round-trip; ordering by relevance; each path that must leave the order untouched (flag off, holdout member, no interest signal); the pin surviving ranking; an already-viewed post not self-matching to the top; an unembedded post holding its position; `didRank()` reporting; and the cross-recipient embedding memo.
- `TestDigestPositions_CohortSplit` covers the cohort partition, including that a recipient outside the holdout whose digest was never reranked stays out of the ranked arm.
- The ModTools digest-clicks spec covers the `cohort` filter parameter.

## Rollback

The feature is off by default and does nothing without `FEATURE_DIGEST_RELEVANCE`. The index is additive and separately reversible.

### DIFF
--- a/.circleci/continue-config.yml
+++ b/.circleci/continue-config.yml
@@ -49,7 +49,15 @@ orbs:
   # and downloads anyway; it now also falls back to the gradle-deps cache that
   # build-android-debug/-dev keep warm from master, which already contains the
   # unpacked distribution.
-  freegle: freegle/tests@1.1.375
+  #
+  # 1.1.377 retries `npm ci` in run-vitest-tests. package.json resolves
+  # cordova-plugin-calendar to a GitHub tarball URL, not a registry package, so
+  # npm's own --fetch-retries (registry fetches only) never covers it. Pipeline
+  # #10740 job 31779 lost `npm ci` outright to one ECONNRESET fetching that
+  # tarball, which failed Vitest and reded the whole build-and-test check. Wrapped
+  # in the same 3-attempt/15s shell retry already used for the Nuxt docker-build
+  # ECONNRESET case above.
+  freegle: freegle/tests@1.1.377
 
 jobs:
   mark-ci-passed:
--- a/.circleci/orb/freegle-tests.yml
+++ b/.circleci/orb/freegle-tests.yml
@@ -1468,7 +1468,20 @@ commands:
             fi
 
             echo "📦 Installing npm dependencies..."
-            npm ci --prefer-offline --no-audit
+            # package.json pulls cordova-plugin-calendar straight from a GitHub tarball
+            # URL rather than the npm registry, so npm's own --fetch-retries (registry
+            # fetches only) does not cover it — an ECONNRESET there aborts npm ci outright.
+            # Retry at the shell level, same pattern used for the Nuxt docker-build retry
+            # elsewhere in this file.
+            for attempt in 1 2 3; do
+              npm ci --prefer-offline --no-audit && break
+              if [ $attempt -eq 3 ]; then
+                echo "❌ npm ci failed after 3 attempts"
+                exit 1
+              fi
+              echo "⚠️ npm ci attempt $attempt failed (possible transient network error, e.g. ECONNRESET fetching a GitHub tarball dependency), retrying in 15s..."
+              sleep 15
+            done
 
             echo "🔧 Running nuxi prepare..."
             npx nuxi prepare
@@ -2559,7 +2572,20 @@ jobs:
                 fi
                 cd "$NUXT_DIR"
                 echo "running" > /tmp/vitest-result
-                npm ci --prefer-offline --no-audit
+                # package.json pulls cordova-plugin-calendar straight from a GitHub tarball
+                # URL rather than the npm registry, so npm's own --fetch-retries (registry
+                # fetches only) does not cover it — an ECONNRESET there aborts npm ci outright.
+                # Retry at the shell level, same pattern as the Nuxt docker-build retry above.
+                for attempt in 1 2 3; do
+                  npm ci --prefer-offline --no-audit && break
+                  if [ $attempt -eq 3 ]; then
+                    echo "fail" > /tmp/vitest-result
+                    echo "❌ npm ci failed after 3 attempts"
+                    exit 1
+                  fi
+                  echo "⚠️ npm ci attempt $attempt failed (possible transient network error, e.g. ECONNRESET fetching a GitHub tarball dependency), retrying in 15s..."
+                  sleep 15
+                done
                 npx nuxi prepare
                 npm run test:unit:run -- --reporter=verbose --coverage 2>&1 | tee /tmp/vitest-output.log
                 EXIT_CODE=${PIPESTATUS[0]}
--- a/iznik-batch/MIGRATION-STATUS.md
+++ b/iznik-batch/MIGRATION-STATUS.md
@@ -315,7 +315,7 @@ These original scripts need to be migrated to Laravel artisan commands:
 |--------|------|----------|-------------|
 | ~~`chat_chaseup_expected.php`~~ | ~~06:00~~ | ~~Medium~~ | ~~Chat expected response chase-up~~ — **Migrated: `chats:chaseup-expected`** (WAITING FOR REPLY user2user notification; scheduler disabled pending go-live) |
 | ~~`birthday.php`~~ | ~~12:00~~ | ~~Low~~ | ~~Birthday notifications~~ — **Migrated: `birthday:send-emails`** |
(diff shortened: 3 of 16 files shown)


edwh:
Closing this. The mechanism has now been measured directly against the outcome it exists to move, and it does not move it.

## The measurement

Rather than run the proposed four-week live experiment, we replayed this branch's own code against historical digest clicks. `DigestRelevanceService::interests()` and `maxCosine()` were used verbatim off the branch, each recipient's interest set was reconstructed as it stood at send time (own posts 60d, viewed posts 30d with `pageview=1`, newest-first, capped at 40, digest's own msgids excluded), and the clicked post was scored against the posts sitting un-clicked in the same digest. The within-digest comparison controls for member and for digest composition; a shuffle control against other members' interest sets controls for the fact that all Freegle items resemble each other.

Sample: `UnifiedDigestDaily`, 2026-07-28 to 2026-08-18, 2,126 digests with a click, 2,884 deduped click events, 1,291 members, 1,823 events usable after requiring a non-empty interest set and a clicked-post embedding, 82,025 clicked-vs-not-clicked pairs.

| | |
|---|---|
| Clicked post, median cosine to interest set | 0.6757 |
| Not-clicked posts, same digest | 0.6759 |
| Within-digest AUC (0.5 = no signal) | 0.5268 |
| Shuffle control | 0.5129 |
| Top-1 hit rate | 5.70% (random 5.33%, shuffled 4.55%) |

The clicked post scores 0.0002 *lower* than its un-clicked stablemates. The signal clears the generic-similarity floor by 0.0139 of AUC.

By arm:

- **Views (arm 2): dead.** AUC 0.5228, on the 0.5129 shuffle floor. This independently reproduces, on clicks, the null already recorded for replies in `docs/developers/reference/first-reply.md` ("Post views are NOT a signal", commit e0160a37c): best cosine 0.596 real vs 0.597 shuffled, n=1,142. That measurement used repeat views, the stronger population; `interests()` takes all views, ~98% of which are single (0.29% reply conversion).
- **Own posts (arm 1): the best number here, and still not usable.** AUC 0.5377 on n=799. This arm was genuinely untested before, so it was fair to hold it open. Measured, it is directionally better than views and still close to a coin flip, sitting in the 0.60 to 0.75 cosine band where top-1 precision is 11%. Every shipped consumer of these embeddings is floored at `MinMatchedPostScore` 0.85, where precision is 0.96; `rank()` applies no floor at all.

Separately, 35% of sampled click events had no qualifying interest set, so `rank()` would have returned those digests unchanged regardless.

## The proposed experiment would not have resolved it either

The `ranked` cohort is conditioned on an interest signal having fired; the holdout is the unconditioned 10%. Measured on today's digests, with the feature off and no reordering applied, members who would qualify for the ranked arm click at 4.06% and members who would not click at 0.74% (n=5,287 over 7 days).

The decision criterion is "keep the ranking if the ranked cohort's CTR is at least the holdout's". That 3 to 6x composition gap passes it on its own, whatever the ranking does. The four-week, 10/90 design also only resolves a ~19% relative lift, which is not an effect size that reordering alone produces; a 50/50 split would reach 12%, so the allocation is not efficient either.

## Ordering parity with browse

A secondary reason not to enable this. `iznik-server-go/isochrone/score.go` is explicitly "a pure, DB-free mirror" of `DigestPostScorer.php`, same four terms and same default weights, so browse and the digest deliberately order by one shared score today. `rank()`'s comparator sorts on `maxCosine` alone, so for every embedded post it discards `_score` rather than blending with it, taking the seen penalty's effect on relative order with it. And because `rank()` runs before the `DIGEST_POST_CAP` truncation, on the ~6.7% of digests that exceed 65 posts it changes which posts appear, not just their order.

## If this is revived

Not as a whole-tail reranker over a recency union. The only shape the evidence supports is the one the scout already uses: the member's own open **Wanted** matched cross-type against **Offers**, floored at 0.85, routed through the existing matched-posts and scout machinery rather than as a second, weaker, digest-only relevance layer with its own query and migration. Same-type matching is the wrong direction anyway (`first-reply.md`: "somebody else wanting what you want is competition rather than a lead"). It can be validated by the offline replay above before it goes anywhere near a cohort split.

## Worth keeping from this work

- The offline click-replay method itself. It answered in one pass, at no risk, what the live experiment was going to spend a month and a hot-table Galera migration on.
- One finding that stands on its own: the digest click dashboard counts only `pN`/Reply clicks and ignores `vN` card clicks, undercounting real digest engagement by roughly 15x. Worth fixing regardless of this PR.

None of this is a criticism of how the branch was built. The pin hold, the unembedded-position hold, the holdout, and the `relevance_ranked` cohort recording all do exactly what the description says. The objection is empirical.

PR #967, opened 2026-07-05
fix(microvolunteering): stop re-notifying users who already reviewed a post (Discourse 9856)

edwh:
Fixes the "microvolunteer post to check keeps re-appearing / badge never clears" bug: https://discourse.ilovefreegle.org/t/9856/3

## Root Cause
This bug has **two distinct surfaces**, and the original fix only addressed one of them:

1. **Review selection** (which post is next to review) - `getApprovedMessageChallenge` / `getPendingMessageChallenge` in `iznik-server-go/microvolunteering/microvolunteering.go` already `LEFT JOIN microactions ... WHERE microactions.id IS NULL`, so an already-reviewed message is correctly excluded from being re-selected. This side was never broken.
2. **Notification badge/count** (`notification.Count` in `iznik-server-go/notification/notification.go`) counts unseen `users_notifications` rows. The previous commit on this branch (`MicrovolunteeringNotifyService::pickCandidates`) stops a **new** duplicate `Exhort` notification from being created for a user who already reviewed a message - but it does nothing for a notification that **already exists**. That row stays `seen=0` forever, so the badge never clears, and tapping its `/microvolunteering/message/<id>` link re-presents the already-reviewed post (the "repeat post" Jacky saw) - independent of the review-selection query, because that page fetches the message directly rather than going through the "next challenge" endpoint.

Confirmed against production (read-only, via the V2 live-DB tunnel): **81 currently-stuck unseen `Exhort` notifications** exist right now where the recipient already has a `CheckMessage` microaction for that message, and **100% of them (81/81) were created after the user's review** - i.e. exactly the notifications the previous fix prevents going forward, but cannot clear retroactively.

## Evidence
New test `test_stale_notification_for_already_reviewed_message_is_cleared` in `MicrovolunteeringNotifyCommandTest`:
- Characterized the bug first: pre-insert a `CheckMessage` microaction plus a stale unseen `Exhort` notification for the same message/user, ran `microvolunteering:notify`, confirmed the notification stayed `seen=0` on the unfixed code.
- Inverted to assert `seen=1`; confirmed it fails on the unfixed code (`Failed asserting that a row ... matches ... "seen": 1. Found ... "seen": 0`).
- Full pre-fix run of the class: **10 passed, 1 failed** (the new test) - no regressions from adding the test.

## Fix
Added `MicrovolunteeringNotifyService::markStaleReviewedNotificationsSeen()`, called once per `notifyForMessages()` run (independent of the day's candidate-message set, since a stuck notification can point at a message older than the 1-day candidate window). It's a single `UPDATE users_notifications un INNER JOIN microactions ma ON ma.userid = un.touser AND ma.actiontype = 'CheckMessage' AND un.url = CONCAT(...) SET un.seen = 1 WHERE un.type = 'Exhort' AND un.seen = 0` - mirrors the exact join already used (and tested) in `loadReviewedByMessage`. Dry-run mode counts instead of writing. The cleared count is now surfaced in `$stats` and the command's info line.

This makes the badge/count consistent with the review-selection exclusion: once a message is excluded from re-selection because the user reviewed it, any notification still pointing at it is also cleared.

## Other Call Sites
Grepped for other `users_notifications` writes/reads touching Exhort/microvolunteering: only `MicrovolunteeringNotifyService` (Laravel, creates them) and the Go `PostResponse` handler (`iznik-server-go/microvolunteering/microvolunteering.go`, marks the *current* vote's notification seen at vote time) touch this type. No other call site needed changing.

## Test
- File: `iznik-batch/tests/Feature/AI/MicrovolunteeringNotifyCommandTest.php`
- Command: `php artisan test --filter=MicrovolunteeringNotifyCommandTest`
- Proves fail-before/pass-after: characterized RED against unfixed code (seen stays 0), fix makes it GREEN (seen becomes 1).

> Note: this environment currently has several concurrent worktree sessions sharing the same `iznik_batch_test` database (each running its own `migrate:fresh`), which repeatedly aborted the post-fix verification run here with concurrent-DDL errors (`Table 'iznik_batch_test.groups' doesn't exist` mid-migration, from another session's simultaneous wipe) - the same class of local-run blocker noted in the previous commit on this branch. The pre-fix baseline run completed cleanly (10 passed, 1 correctly failed), and the fix is a minimal, lint-clean, single-query extension of the already-tested `loadReviewedByMessage` pattern, validated directly against production data. Relying on CI for the definitive full-suite green run.

## Discourse
https://discourse.ilovefreegle.org/t/9856/3

### DIFF
--- a/iznik-batch/app/Services/MicrovolunteeringNotifyService.php
+++ b/iznik-batch/app/Services/MicrovolunteeringNotifyService.php
@@ -30,10 +30,14 @@ class MicrovolunteeringNotifyService
     /** @var array<int, bool> userid => true for users with any Exhort microvolunteering notification today */
     private array $alreadyNotifiedToday = [];
 
+    /** @var array<int, array<int, bool>> msgid => {userid => true} for users who already reviewed that message */
+    private array $reviewedByMessage = [];
+
     public function notifyForMessages(bool $dryRun = false): array
     {
         $this->eligibleCache       = [];
         $this->alreadyNotifiedToday = $this->loadAlreadyNotifiedToday();
+        $this->reviewedByMessage    = [];
 
         $stats = [
             'messages_considered' => 0,
@@ -58,6 +62,16 @@ class MicrovolunteeringNotifyService
         ", [self::NOTIFICATION_TYPE]);
 
         $stats['messages_considered'] = count($msgs);
+
+        // Users who have already recorded a CheckMessage microaction for a message
+        // must never be re-notified about it. Without this, a rippling post - whose
+        // messages_groups.arrival is refreshed each time it ripples into a new group -
+        // keeps re-entering the "arrival within 1 day" gate and re-lights the same
+        // person's "post to check" badge for ever, even after they have reviewed it.
+        // See Discourse 9856.
+        $this->reviewedByMessage = $this->loadReviewedByMessage(
+            array_values(array_unique(array_map(fn ($m) => (int) $m->id, $msgs)))
+        );
 
         Log::info("MicrovolunteeringNotify: considering " . count($msgs) . " messages");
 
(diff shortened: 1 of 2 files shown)


edwh:
Verified locally on a freshly-rebuilt env: `MicrovolunteeringNotifyCommandTest` — **10 tests, 20 assertions, 0 failures** (incl. the two new re-notification tests). The earlier local blocker was a mis-bound dev container, now fixed.

PR #972, opened 2026-07-06
feat(routing): reachable-node group set for water/toll-correct ripple targeting

edwh:
Implements the routing-server side of `plans/2026-07-06-ripple-reach-crosses-water.md` (TDD).

## Why
The rippling reach **polygon** is a lossy smoothing of the reachable road-node set (Dijkstra) and over-approximates — it bridges rivers (the ~0.8 km Thames at Tilbury), so posts ripple into groups you can't drive to (canonical case: msgid 120898144 Corringham → Gravesend). The exact, correct signal — *which groups contain a reachable node* — is computed then thrown away. A reached node can't cross a river (no edge does) or use an excluded toll, so this is threshold-free and correct for both severance and tolls.

## What (routing server only)
- `reachable_groups.go`: `reachableGroupIDs(g, reached, groups)` — bbox + ray-cast each reached node against group polygons; dedupes a MULTIPOLYGON group split across parts. `pointInPolygon` promoted from the test file to production.
- **`GET /v1/reachable-groups?lat=&lng=&minutes=&mode=`** and a **`reachable_group_ids`** field on **`/v1/ripple-schedule`** — computed once per reach from the node set already in hand (cheap: one bbox+PIP pass, no extra Dijkstra). Empty serialises as `[]` (not `null`) so batch can tell a new server from an old one.
- **The smoothed polygon is untouched** and still drives map display / digest scoring / nearby-reach.

## Tests
Routing suite green (75). New: node→group classification (separated clusters / empty / MULTIPOLYGON dedupe), the handler (missing-param 400, present-list 200), and the schedule-field roundtrip (incl. `[]` vs `null`).

## Follow-ups (separate PRs, per the plan)
- **Batch gate**: `ExpandService::rippleIntoNewGroups` AND-gated on `reachable_group_ids` behind `freegle.ripple.reachable_gate` (default off) + the `rippling_reach.reachable_group_ids` migration + `removeGroupsNoLongerReached` extension; then the `ripple:retract-unreachable` backfill (throttled, one row at a time).
- **Display-polygon refinement (your "better display" ask)**: assessed as **defer**. Once targeting is node-gated the polygon crossing water is cosmetic, and it feeds 5+ consumers (map, digest, nearby-reach PR #921, retract, ripple-eval), so changing `IsochronePolygon` carries real regression risk. Lowest-risk future option: skip the morphological closing for drive mode, guarded by a two-cluster regression test — noted for a follow-up.

### DIFF
--- a/iznik-routing-go/polygon_test.go
+++ b/iznik-routing-go/polygon_test.go
@@ -48,18 +48,4 @@ func TestAutoResolution(t *testing.T) {
 	t.Logf("walk15 res=%.5f drive15 res=%.5f", walkRes, driveRes)
 }
 
-// pointInPolygon uses ray casting. Coords are [lng, lat].
-func pointInPolygon(lng, lat float64, ring [][2]float64) bool {
-	inside := false
-	n := len(ring)
-	j := n - 1
-	for i := 0; i < n; i++ {
-		xi, yi := ring[i][0], ring[i][1]
-		xj, yj := ring[j][0], ring[j][1]
-		if ((yi > lat) != (yj > lat)) && (lng < (xj-xi)*(lat-yi)/(yj-yi)+xi) {
-			inside = !inside
-		}
-		j = i
-	}
-	return inside
-}
+// pointInPolygon now lives in reachable_groups.go (production code), reused here.
--- a/iznik-routing-go/reachable_groups.go
+++ b/iznik-routing-go/reachable_groups.go
@@ -1,0 +1,124 @@
+package main
+
+import (
+	"math"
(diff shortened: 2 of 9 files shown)


edwh:
Superseded: the reached-node-in-group-polygon gate has been replaced by a member-based reachability gate (a group is reached only if an active member living inside the group's own polygon has a road-reachable street node), plus per-tick reachable_group_ids in the ripple schedule and an exact no-smoothing display polygon. That work will arrive as a fresh PR; the branch here is retained for reference.

PR #973, opened 2026-07-06
feat(ripple): gate ripple targeting on road-reachable groups (batch consumer)

edwh:
Consumer side of `plans/2026-07-06-ripple-reach-crosses-water.md` — pairs with the routing server in #972.

## Why
The rippling reach **polygon** is a lossy smoothing of the reachable road-node set and over-approximates, bridging rivers (the ~0.8 km Thames at Tilbury) so posts ripple into groups you can't actually drive to. #972 makes the routing server return the exact signal — **which groups contain a reachable road node** (`reachable_group_ids`). This PR makes the batch consume it to gate ripple targeting.

## What
- **ReachService::parseScheduleResponse** now carries `reachable_group_ids` (empty when an older server omits it, so the gate can tell "computed zero" from "not available").
- **Migration**: `rippling_reach.reachable_group_ids` JSON column (Laravel + idempotent prod SQL). NULL on pre-rollout rows.
- **Config flag** `freegle.ripple.reachable_gate` (`RIPPLE_REACHABLE_GATE`, **default off**, independent of `RIPPLE_ENABLED`). Enable only once #972's routing server is deployed.
- **ExpandService::rippleIntoNewGroups** — when the gate is on AND a non-empty reachable set is present, the target SELECT is AND-gated with `g.id IN (…server-sourced ints…)` on top of the existing polygon `ST_Intersects` prefilter. Threaded through both ripple paths (fresh schedule at init; cached column on later ticks). The reach row is populated on both writes.
- **ExpandService::retractOutOfReachCopies** — with the gate on, also retracts a rippled-in group the polygon still covers but which is no longer in the node-reachable set, so a reach that overshot water self-heals. Guarded by `reachable_group_ids IS NOT NULL`, so pre-rollout rows are only ever retracted by the existing polygon rule.

Gate off (default): every query is byte-for-byte the pre-feature behaviour — the added clause is omitted entirely.

## Tests
- ReachServiceTest: `reachable_group_ids` parsed; defaults to `[]` when absent (RED confirmed, then GREEN).
- ExpandServiceTest: gate **blocks** a polygon-covered group not in the set, **allows** one in the set, and **off** ripples the intersecting group regardless; retraction **removes** a rippled group outside the set and **keeps** one inside.

## Rollout
1. Deploy #972 (routing server returns `reachable_group_ids`).
2. Deploy this (column + code, gate off — no behaviour change; reach rows start recording the set).
3. Flip `RIPPLE_REACHABLE_GATE=true` once enough rows carry a set; the retraction backfill then self-heals existing over-water copies.

### DIFF
--- a/iznik-batch/app/Services/Ripple/ExpandService.php
+++ b/iznik-batch/app/Services/Ripple/ExpandService.php
@@ -224,12 +224,13 @@ class ExpandService
             DB::statement(
                 'UPDATE rippling_reach
                     SET polygon = ST_GeomFromText(?, ' . self::SRID . '),
-                        schedule = ?, total_freeglers = ?, max_drive_min = ?,
+                        schedule = ?, reachable_group_ids = ?, total_freeglers = ?, max_drive_min = ?,
                         updated_at = updated_at
                   WHERE msgid = ?',
                 [
                     $storeWkt,
                     json_encode($ticks),
+                    json_encode($schedule['reachable_group_ids'] ?? []),
                     (int) $schedule['total_freeglers'],
                     $schedule['max_drive_min'],
                     $row->msgid,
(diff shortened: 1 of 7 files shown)


edwh:
Superseded: the routing-server signal this consumed has been replaced by a member-based gate with per-tick reachable_group_ids (see #972's closing note). The batch consumer will be reworked against the per-tick contract together with the reach-polygon/membership backfill, as a fresh PR. Branch retained for reference.

PR #978, opened 2026-07-06
feat(rippling): member-based reach gate, per-tick targeting, exact display polygon

edwh:
## Summary

Rippling targeting no longer trusts polygon overlap. A group receives a rippled post only if **at least one active member (90-day, Approved) who lives inside that group's own polygon can actually drive to the post** - their postcode-centre street node is in the Dijkstra reached set. This kills three failure classes verified on live data:

- **Water/toll severance**: a Corringham offer no longer targets Gravesend across the Thames (no member's street is reachable; the Dartford crossing is toll-excluded).
- **Empty-sliver targeting**: nicking a corner of a sprawling catchment polygon no longer ripples a copy nobody will see (e.g. High Wycombe: 468 in-polygon actives, none within the budget by road).
- **Privacy-blur jumps**: a blurred point landing mid-river can snap to the wrong bank; a postcode centre sits on its own street and cannot (server-side only - member locations never leave the server).

### Routing server (iznik-routing-go)
- `/v1/reachable-groups`: the member-based gate. Member query drives off the `users_approxlocs` spatial index (`STRAIGHT_JOIN` + `MBRContains`): 21s → 0.9s estuary / 3.5s London.
- `/v1/ripple-schedule`: every tick carries its own `reachable_group_ids` (thresholding the single Dijkstra - no extra routing). New `polygons=0` slim form for the batch: **6KB vs 24MB** on a London call and ~7s faster, keeping `ripple:expand` inside its minute budget at peak (6,877 new reaches/day on live).
- **The displayed polygon is exactly what is reached**: cell size derived from the road network (p75 reached-edge length, ~33-330m - the old budget-scaled ~1.1km cells were wider than the Thames), reached nodes *and* road segments stamped (water is never stamped), one cell of growth, and **no simplification** - Douglas-Peucker removed server-side. A regression test pins the river-bridging case.
- `/demo` removed (ModTools `/rippling` replaces it).

### Rippling Explorer (iznik-nuxt3 modtools)
- Group tint/list driven by the gate (max-extent fetch per pin + per-tick ids on animation frames): what tints is exactly what targeting decides, at every tick.
- Client-side Chaikin corner-smoothing removed - the rounded curve bulged the boundary across rivers.

### Batch (iznik-batch)
- Includes the reachable-gate consumer (from the closed #973, review-hardened) adapted to the slim contract: `ReachService` requests/parses slim schedules and fetches single tick polygons via `catchmentWkt()`; init/advance/recompute resolve tick polygons on demand; targeting and the stored reachable set use the **current tick's** ids.
- **`ripple:rebuild-reach`**: one-off idempotent backfill - rebuilds every active reach polygon/schedule under the new algorithm and retracts rippled-in copies plus the ripple-created memberships in groups no longer covered. Dry-run first, sliceable via `--limit`, preserves `updated_at` so no mail fires. (Distinct from `ripple:backfill`, the pre-go-live seeder.)
- Everything ships dark: `RIPPLE_REACHABLE_GATE` default OFF; with it off, targeting/retraction are byte-for-byte the old behaviour.

### Measured live impact (40-post sample, ripple-eligible posts, current-tick budgets)
- Median groups/post 8 → 5; 34/40 lose ≥1 group; removals concentrate in dense conurbations = withdrawal of coarse-polygon overshoot. Verified cases all correct (Stirling: nearest resident beyond 30-min budget, returns at 45; Forth crossing reachable both ways; Gravesend excluded per-tick).
- TN posts (46% of reaches) never rippled into new groups and still don't - the per-post guards are untouched.
- Exposure noted: 3 groups nationally have zero in-polygon actives (2 test groups + Outer Hebrides), 6 are fragile with 1-4 - all remote/tiny.

### Docs
- `RIPPLING-ALGORITHM.md`: technical reference (isochrone/Dijkstra explained, polygon pipeline, extent governor, targeting incl. rejected approaches, mechanics, config, data model). Moderator guide cross-links it.

## Code Quality Review
- Targeting decisions and display come from the same per-tick data - no divergence between what moderators see and what the batch does.
- No magic constants introduced: cell size is network-derived; the gate is threshold-free (≥1 member); the only fixed number (90-day activity) reuses the existing group-actives definition.
- Backwards compatibility at every seam: old servers ignore `polygons=0`; old stored schedules (with WKT, without per-tick ids) still advance; pre-rollout reach rows (NULL `reachable_group_ids`) are never id-retracted.
- Known trade-offs: unsimplified polygons are ~4x the vertices (display-exactness paid deliberately; the slim schedule keeps them out of storage); the gate leans on group polygon quality (documented, measured, tail is tiny).

## Future Improvements
- Advance-time polygon fetch adds one 0.7s catchment call per tick advance (~62k/day at current volume) - fine concurrently, but a per-tick polygon cache by (origin, budget) would remove most of them.
- Schedule freegler *counting* (extent governor pacing) still uses blurred points via `within_coords`; the gate does not. Unifying them is possible but changes pacing behaviour, so deliberately not done here.
- Group polygon quality audit (e.g. Queensferry: 11 of 535 actives in-polygon) as a separate data task.

## Test Plan
- Go: full suite green (78 tests via status API), including new tests for the member gate (severance drop, ≥1-member sufficiency, offshore snap), per-tick thresholding, slim schedule form, and the river-bridging display regression.
- Laravel: full suite green via status API, including new ReachService slim-parse/catchment tests and ExpandService slim-init/slim-advance/rebuild-reach (dry-run, live retraction, idempotency) tests.
- Vitest: full suite green (13,854), geometry spec pinned to the no-smoothing contract.
- Live validation on the UK graph + prod DB (read-only): gate excludes Gravesend/Dartford from both test origins and from every schedule tick; monotonic per-tick growth; origin always inside the drawn polygon; Gravesend town outside it.


edwh:
Opened against the wrong repo by tooling (branch lives in FreegleDocker); superseded by the FreegleDocker PR.

PR #986, opened 2026-07-06
fix(birthday): include email-only mods in group birthday volunteer list

edwh:
## Problem

The "Happy Birthday to X Freegle!" email lists the group's volunteers, but `BirthdayService::getActiveVolunteers()` filtered mods on `users.lastaccess` within a year. `lastaccess` only updates on an authenticated web/app login - a mod who moderates purely by email reply never refreshes it, so they were silently dropped from the volunteer line every year while appearing as an active volunteer everywhere else on the site. Reported by support for Oxford's birthday email (one mod consistently missing).

## Fix

Drop the `lastaccess` filter to match the canonical volunteer query (`iznik-server-go/group/groupVolunteer.go`), which has no such filter - the same source of truth this method's comment already claims to mirror. Role, collection, deleted and `showmod` filtering are unchanged.

## Tests

- Adds the first direct test of `getActiveVolunteers`: a mod with `lastaccess` two years ago is included; `showmod=false` is still respected.
- Full local Laravel suite: **4733 passed** (worktree, status API).

https://claude.ai/code/session_01E5jyvZJ7xJB58tbrVjj86L

### DIFF
--- a/iznik-batch/app/Services/BirthdayService.php
+++ b/iznik-batch/app/Services/BirthdayService.php
@@ -102,8 +102,6 @@ class BirthdayService
 
     private function getActiveVolunteers(int $groupId): array
     {
-        $oneYearAgo = now()->subYear()->format('Y-m-d H:i:s');
-
         // Match the source of truth used by the modtools "Show me as a volunteer"
         // toggle, the Go API's group volunteer list (iznik-server-go
         // group/groupVolunteer.go:45), and every other on-site volunteer display:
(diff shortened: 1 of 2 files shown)


edwh:
Closing: wrong root cause. The Oxford report (Discourse https://discourse.ilovefreegle.org/t/happy-birthday-to-oxford-freegle/9386) was actually fixed by 453614648 (2 Jun 2026), which switched the volunteer list from users.publishconsent (defaults 0, never written by the UI toggle) to settings.showmod - that's why John was missing despite 'Show me' being set. The lastaccess-within-a-year filter this PR removed is deliberate V1 parity ('active volunteers', matching Donations.php logic, see scripts/fix/fix_list_birthday_moderators.php), so removing it would have changed intended behaviour.

PR #996, opened 2026-07-07
perf(ripple): checked-once-forever negative memoization for proximity notes

edwh:
## Summary

- `ripple:proximity-notes` re-computed "not quicker" / "unreachable" rows on every 5-minute run for the whole 8-day candidate window, because a `rippling_proximity` row is only written when `quicker=true`. For the 2026-07-06 group-21521 Sentry storm that was 406 of 635 rows recomputed forever - a standing tax of up to ~12 CPU-hours/day of routing work (200-row cap x ~754ms x 288 runs/day).
- New `rippling_proximity_checked` marker table (Laravel migration + idempotent production SQL): a row means this (msgid, groupid) rippled-in copy got a **definitive** answer and is never re-queried. A separate table rather than nullable p/q columns because iznik-server-go scans `rippling_proximity.p/q` as non-null strings.
- `ReachService::groupProximity` widened to a tri-state (`PROX_OK` / `PROX_UNREACHABLE` / `PROX_ERROR`): failed calls (timeout, non-2xx, exception - e.g. the routing server mid-PBF-rebuild) are never memoized and retry next run, so a routing restart cannot permanently suppress notes.
- The command excludes marked rows (keeping the existing note-row exclusion for pre-marker rows), writes the marker only on definitive outcomes (note written / not quicker / unreachable - not on geocode gaps), and purges markers older than 14 days (candidates only span 8 days of arrivals, so checked-once-forever still holds while the table stays small).
- Includes `plans/routing-performance-step-change.md` - the full routing/isochrone performance design study this is Phase 0 of.

## Code Quality Review

- Tri-state contract documented on the constants and method; the only caller is `ProximityNotesCommand` (verified by grep across the monorepo).
- Marker writes use `insertOrIgnore` inside the existing per-row try/catch, so a mid-run failure degrades to "retry next run", never to a wrong tombstone.
- Purge is a single indexed DELETE per run (`checked_at` index) and can never expire a marker that could still match a candidate (14d > 8d arrival window).
- FKs cascade with messages/groups, mirroring `rippling_proximity`.

## Future Improvements

Later phases of the design study (see the plan doc): dedup the per-freegler nearest-node loop in the routing server's `/v1/ripple-schedule` (measured as the real hot-path cost), flat-epoch Dijkstra core, superset isochrone caching, weak-component snap fix, mmap graph snapshot.

## Test Plan

- Red-first TDD: the 4 new tests + 1 extended assertion failed before the implementation (verified), pass after.
- New tests in `ExpandServiceTest`: marker on not-quicker + never re-queried (HTTP request count is flat across runs); marker on unreachable + never re-queried; **no** marker on HTTP 500 and the row IS retried (request count grows); purge removes only expired markers; quicker path writes note + marker.
- Full Laravel suite via the status API: **4751 tests passed** (includes fresh `migrate:fresh` exercising the new migration).

### DIFF
--- a/iznik-batch/app/Console/Commands/Ripple/ProximityNotesCommand.php
+++ b/iznik-batch/app/Console/Commands/Ripple/ProximityNotesCommand.php
@@ -15,9 +15,17 @@ use Illuminate\Support\Facades\Log;
  * yet have a rippling_proximity row, resolves P (nearest in-group point to the offer) and Q
  * (furthest in-group point from P) to place names, and stores them only when quicker=true.
  *
- * Idempotent (skips rows that already have a note), bounded per run (--limit), and disabled by a
- * dedicated flag (freegle.ripple.proximity_notes) so it can be turned off without touching the
- * master RIPPLE_ENABLED switch.
+ * Negative memoization (Phase 0, plans/routing-performance-step-change.md): every DEFINITIVE
+ * routing answer — note written, not quicker, or unreachable within budget — also writes a
+ * checked-once-forever marker to rippling_proximity_checked, and marked rows are excluded from
+ * future runs. Without this, "no note needed" rows (the majority) were recomputed every 5-minute
+ * run for the whole 8-day window: up to ~12 CPU-hours/day of wasted routing work (the 2026-07-06
+ * group-21521 Sentry storm's standing tax). Failed calls (PROX_ERROR: timeout/non-2xx/mid-restart)
+ * are NOT marked, so those rows retry next run.
+ *
+ * Idempotent (skips rows that already have a note or a marker), bounded per run (--limit), and
+ * disabled by a dedicated flag (freegle.ripple.proximity_notes) so it can be turned off without
+ * touching the master RIPPLE_ENABLED switch.
  */
 class ProximityNotesCommand extends Command
 {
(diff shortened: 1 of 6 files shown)


edwh:
**Note on the red "Coveralls - playwright" status (-0.02%)** - investigated, not code-caused:

- This PR's diff contains no playwright-instrumented code (4 PHP files, 1 SQL, 1 md); the branch's frontend is byte-identical to its parent `4ccf8dd6`.
- Coveralls' own data for this build: aggregate coverage **+0.006%**, diff coverage **92%** (23/25 new lines covered). Laravel flag (the code this PR touches) **increased**.
- The playwright coverage metric is run-nondeterministic: master's three consecutive builds today moved -0.009% / +0.005% / -0.004%, including `93538da48` which changed only a Laravel migration (zero instrumented lines). E2E-covered branches vary with runtime conditions (ad fill, One Tap prompts), so a ±0.02% flap happens on identical code.
- All 151 Playwright tests passed on this run; the status is a ratio comparison, not a test failure.

Durable fix for this recurring flap (it hits unrelated PRs regularly) would be a small change-threshold on the Coveralls repo settings (admin web UI) or making ad-slot rendering deterministic in CI - both outside this PR's scope and a maintainer call.

PR #1006, opened 2026-07-08
feat(infra): cgroup v2 slice prioritisation for the batch-prod host

edwh:
## What

Work-conserving cgroup v2 priority tiers for the batch-prod host, so user-facing services are guaranteed resources under contention while batch can still use 100% of an idle box - replacing "nothing" (the stack currently has zero resource controls of any kind).

- **`interactive.slice`** (CPUWeight=800, MemoryLow=4G, IOWeight=500): `embedding-sidecar` (live apiv2 semantic search), `postfix`.
- **`batch.slice`** (CPUWeight=25, MemoryHigh=10G, MemoryMax=14G, IOWeight=50): `batch-prod`, `spatial`, `spatial-knn`.
- Everything else stays default (system.slice, weight 100). Under full contention CPU splits ~86/11/3; idle capacity stays fully usable by anyone.
- Wiring via new `docker-compose.override.batchprod.yml` - the prod host opts in through `COMPOSE_FILE`; **dev and CI never load it** (dev profile-set service list verified byte-identical to master).
- Deployment + verification runbook: `plans/2026-07-08-batch-host-cgroup-slices.md`. **Merging this changes nothing on prod** - deployment is a separate human-gated step (install slice units, add override to COMPOSE_FILE, recreate 5 containers).

## Why now

Measured on the host 2026-07-08: load 13.6 on 8 cores with `batch-prod` at 428% CPU, 12G/25G swap in active churn, PSI ~25% sustained CPU stall. Host prerequisites verified live: cgroup v2, systemd cgroup driver, controllers delegated. This is also the dress rehearsal for the same pattern on the edge VM (PR #894), where it collapses the plan's VM-A/VM-B split (image cache vs tile-render storms) into one VM.

## Also in this PR

- `spatial`/`spatial-knn` gain the **`production` profile**. The batch host (`COMPOSE_PROFILES=backend,production,mail`) only got them because Compose v5 auto-enables hard `depends_on` targets across profiles; older `docker-compose` binaries reject the whole project as "invalid compose project" (breaking local tooling against the prod profile set). Explicit inclusion fixes validation and documents reality. Deliberately **not** `required: false` - that would make Compose v5 skip starting spatial on prod, recreating the 2026-07-03 digest hang.
- Reboot runbook correction: `spatial`/`spatial-knn` are `restart: "no"` (come back via `compose up`, not the docker daemon), not `unless-stopped` as the table claimed.

## Validation

- Strict `docker-compose` v2.35 config matrix: prod profile set now resolves (previously invalid) to exactly the 11 services prod runs today, with and without the override; `cgroup_parent` renders on the intended 5 services; `mem_reservation` renders.
- Dev profile set (`frontend,database,backend,dev,monitoring`): resolved service list identical to master. Yesterday override unaffected (doesn't enable `production`).
- `systemd-analyze verify` clean on both slice units.

### DIFF
--- a/batch.slice
+++ b/batch.slice
@@ -1,0 +1,20 @@
+# Background/batch tier on the batch-prod host (see interactive.slice for
+# install + rationale). batch-prod is the box's dominant CPU consumer; this
+# slice lets it soak up idle capacity while guaranteeing it yields to the
+# interactive tier under contention.
+#
+# CPUWeight=25 keeps ~3% of CPU guaranteed under full contention so scheduled
+# jobs (digests) still make progress. For a true scavenger class that gets
+# CPU only when nothing else wants it, use CPUWeight=idle instead.
+#
+# MemoryHigh throttles-and-reclaims batch before it drives the host into
+# swap; MemoryMax means a runaway job OOMs inside this cgroup instead of
+# taking out the interactive tier. Tune against `free -h` and
+# /proc/pressure/memory after observing real usage (batch-prod RSS was
+# ~2.7G, spatial ~5.8G when these defaults were chosen, host RAM 23G).
+
+[Slice]
+CPUWeight=25
+MemoryHigh=10G
+MemoryMax=14G
+IOWeight=50
--- a/docker-compose.override.batchprod.yml
+++ b/docker-compose.override.batchprod.yml
@@ -1,0 +1,39 @@
+# Batch-prod host ONLY - not used in dev or CI.
+#
+# Opt in on the prod batch host by setting in its .env:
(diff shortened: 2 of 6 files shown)


edwh:
The red "Coveralls - playwright" check on this PR is the documented coverage nondeterminism, not anything in this diff (which contains no product code). Per-line Coveralls comparison against the master baseline shows exactly one flipped line: SpinButton.vue's `onConfirmClosed` body, covered in master's lucky sample but not in this run's. The durable fix (v8-ignore + a deterministic unit spec, mirroring PR #910) is in its own PR; once that merges, a rebase/rerun here clears the check.

PR #1007, opened 2026-07-08
fix(coverage): stop SpinButton onConfirmClosed flipping Playwright coverage

edwh:
## What

Kills the recurring spurious "Coveralls - playwright: Coverage decreased" red check on unrelated PRs, by fixing the one remaining nondeterministic line.

Diagnosed with per-line Coveralls data (method from the PR #910 investigation): diffing master build 80471664 against PR #1006's build 80472473 shows exactly one file changed - `SpinButton.vue`, 10/11 → 9/11 - and per-line data pinpoints it to `onConfirmClosed`'s body (line 207). Master's baseline run happened to include a run dismissing a SpinButton confirm dialog; the PR's run didn't. Whether any e2e run takes the dismiss-rather-than-confirm path varies run to run. (PR #910 predicted a residual flipper and guessed LoginModal - it turned out to be a second line in SpinButton itself.)

## Fix (two parts, mirroring #910)

- `/* v8 ignore next */` on the line so monocart/Playwright coverage stops sampling it. vitest uses istanbul, which ignores v8 comments, so unit coverage still counts it.
- A new unit spec covering the dismiss path deterministically (stub emits `hidden`, asserts the modal unmounts, `showConfirm` resets, and no `handle` is emitted). Follows the proven `PostPhoto.spec.js` pattern for stubbing the async-declared ConfirmModal.

A one-time small baseline shift lands on this commit, then the flip-flopping stops.

eslint + prettier clean on both files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014V1ihELPDZSe3CWT461fyK

PR #1009, opened 2026-07-08
fix(reply): overlay silently fails to open if the message fetch rejects

edwh:
## What

`ChatReplyPane` does a top-level `await messageStore.fetch()` in async setup under `<Suspense>`. If that fetch rejects - e.g. a transient connectivity blip - the Suspense never resolves, `.reply-overlay` never enters the DOM, and the user's Reply click is a **silent no-op**: no overlay, no error, no retry.

## How it was found

Root-caused from PR #1007's CI run: the standalone-page CTA reachability check timed out waiting for `.reply-overlay` ("element(s) not found", i.e. never mounted - not hidden) immediately after the app's own console logged `Gone offline TypeError: Failed to fetch`. 150/151 passed; the identical product code passed the same suite two hours earlier on PR #1006's run. The trigger is environmental (an API blip mid-run), but the defect - an unguarded top-level await under Suspense - is real, and would bite a real user replying during a connection blip.

## Fix

Guard the await: on failure, log a client action (`chatreplypane_fetch_failed`) and render with cached store data. The page that opened the pane has almost always just fetched the message, and the composer gates itself if data is genuinely missing. Includes a regression unit test: fetch rejects → overlay still mounts.

eslint + prettier clean.

### DIFF
--- a/iznik-nuxt3/components/ChatReplyPane.vue
+++ b/iznik-nuxt3/components/ChatReplyPane.vue
@@ -346,8 +346,17 @@ const newUserModal = ref(null)
 const replyToPostChatButton = ref(null)
 const emailValidatorRef = ref(null)
 
-// Fetch the message data
-await messageStore.fetch(props.messageId)
+// Fetch the message data. Guarded: this is a top-level await in async setup
+// under <Suspense>, so an unhandled rejection (e.g. a transient connectivity
+// blip) would mean the overlay never mounts and the Reply click is a silent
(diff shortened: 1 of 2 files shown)


edwh:
Audited the other Suspense sites for the same defect class: MessageList (x3) and MyPostsPostsList wrap OurMessage/MessageSummary, which have no top-level awaits in setup - so ChatReplyPane was the only component where a rejected fetch could leave a Suspense permanently unresolved. No broader sweep needed.

PR #1012, opened 2026-07-08
test(e2e): wait for hydration before bulk register-interest interactions

edwh:
## What

Fixes the bulk-offer e2e failure on master pipeline #8990 (the first post-#894 build): the empty-submit assertion clicked `register-interest` and timed out waiting for the `register-error` banner.

## Root cause

`registerInterest()` in the spec never waits for Nuxt hydration. Every element it waits on before clicking - the pick toggles, item photos, and the register button itself - renders **server-side**, so their visibility proves nothing about interactivity. On a loaded CI box (this run had a parallel worker hammering the same containers) the click landed on inert pre-hydration DOM: no Vue handler ran, `bulkTriedRegister` was never set, and the banner - whose render is gated on that ref - never appeared. 67.5s timeout with "element(s) not found".

Static analysis rules out the alternatives: with nothing picked, `BulkItemsInterest`'s `validationError` computed is unconditionally non-empty and emitted immediately, and `registerBulkInterest()` shows the banner on every path; and `getByTestId` had already located the button, so `isBulk` was true. The only path to "banner never enters the DOM" is the handler never running.

## Fix

`await waitForNuxtHydration(page)` after navigation in `registerInterest()` - the exact pattern the reply-to-chat CTA spec already uses before interacting. Covers r1's empty-submit check plus all subsequent toggle/slot/submit interactions for both repliers.

eslint + prettier clean. Spec-only change; CI executes it (local Playwright runs test the main checkout tree, not this branch).

### DIFF
--- a/iznik-nuxt3/tests/e2e/test-bulk-offer-flow.spec.js
+++ b/iznik-nuxt3/tests/e2e/test-bulk-offer-flow.spec.js
@@ -12,6 +12,7 @@ const path = require('path')
 const { test, expect } = require('./fixtures')
 const { timeouts } = require('./config')
 const { loginViaHomepage, logoutIfLoggedIn } = require('./utils/user')
+const { waitForNuxtHydration } = require('./utils/reply-helpers')
 
 const ASSET = (f) => path.join(__dirname, 'assets', f)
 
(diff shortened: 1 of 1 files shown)


edwh:
The two red Coveralls delta-gates are verified per-line against the current master baseline (jobs 184384497/184384494 vs this PR's 184385766/184385763): laravel = AutoRepostService 164/172→163/172, i.e. line 184, the ORDER BY RAND() 1-in-3 coverage roll fixed in #1011; playwright = SpinButton 10/11→9/11, the onConfirmClosed dismiss-path roll fixed in #1007. All four suites pass. Once #1011 and #1007 merge, a rerun here compares against deterministic baselines and both contexts go green.

PR #1013, opened 2026-07-08
Migrate TN messages sync from email to API

fnnbrr:
## Test plan

WIP test to run both email and API message ingestion can be run by attaching a shell to the `iznik-batch` container and running `php artisan tn:parity-check`. The `--date-min` option is recommended with an ISO timestamp of about an hour ago to avoid processing an unnecessarily large number of posts.

## CI

Master merged in. Two things were failing, neither of them visible from the job summary:

- The build died in the **ORM migration inventory ratchet (Gate 1)** before any suite ran, which is why the artifacts showed `result=running` and read as a Laravel failure. That gate no longer exists — orb `1.1.373` drops it, since `.claude/check-raw-sql.sh` keeps raw SQL out at authoring time. This branch was pinned to an older orb; merging master retires it.
- With the suites actually running, one real failure remained: `GroupPostIngestionServiceTest::test_different_subject_at_same_location_does_not_trigger_repost`, fixed here. `findClosestPostcodeId` returns an id from the spatial index, which is a separate store from the database being written; writing it into `users.lastlocation` (a foreign key) throws when no such row exists, and because that happens inside `createMessage` the whole post is dropped. The id is now verified before use, with a log line when it is stale — the same failure is possible in production whenever the index outlives a purged location.


edwh:
## Adversarial review

**PR description:** the Summary section is still the template placeholder. This PR needs a plain-English summary of what it does and how the cutover works.

### Confirmed bugs (verified against the code)

1. **Worry words read the wrong table.** `GroupPostIngestionService::subjectContainsWorryWords()` (line 647) queries the legacy `worrywords` snapshot with `type === 'Allowed'`. The email path was explicitly fixed (Discourse #9944, ac3f80c82) to read `concern_keywords` (`category === 'allowed'`) because moderator whitelist edits only reach that table. A phrase whitelisted via ModSupportConcernKeywords has no effect on TN API posts, which are held as "worry words" while the same content arriving by email is let through. This re-introduces a bug that was already fixed once.

2. **Unmoderated posters get instant approval.** The email path routes DEFAULT/UNMODERATED posting statuses to Pending so `messages:contentcheck` gates them, with a comment saying they are deliberately NOT approved on arrival. `GroupPostIngestionService::ingest()` (line 291) maps DEFAULT/UNMODERATED straight to `approved`, sets collection=Approved and indexes immediately, skipping the content check entirely. A parity test acknowledges this as "the already-known email-pends/API-approves divergence", but nothing in the service documents it as intended, unlike every other deliberate divergence in the file.

3. **The consent field can never be populated in production.** `freegle_group_ids` was hand-added only to the generated `Post` model's constructor (`PublicApi/lib/Model/Post.php:364`). `ObjectSerializer::deserialize()` builds models via `new Post()` with no data and only sets properties listed in `openAPITypes()`/`attributeMap()`, which omit the field. So on the live API path `$post['freegle_group_ids']` is always null, `?? []` masks it, and every TN post is stamped `mod_messaging_allowed = false` regardless of actual consent (`PostSyncer.php:260`). Tests pass because fixtures are plain arrays that bypass deserialization. The hand edit is also unprotected by `.openapi-generator-ignore`, so regenerating the client silently deletes it.

### Other findings

- `TnEmailRoutingGate::isTrashNothingGroupPost()` (line 77) excludes chat replies and volunteers/auto mail but not the earlier `route()` phases (bounce, digest reply, dropped sender, auto-reply, self-sent, known spammer). Post-cutover, TN mail matching those is archived and skipped as `SkippedTnApi` instead of getting that handling, contradicting the class's own docblock.
- `postAlreadyExists()` (line 628) has no `deleted` filter, unlike the email path's `findLiveTnMessage()`. Once a TN-ingested message is deleted, any resync/backfill of that post is permanently reported as duplicate and dropped.
- `docs/developers/reference/rippling-algorithm.md:378` says the rippling TN exclusion is gated on `FREEGLE_TN_INGEST_POSTS_VIA_API`, but `ExpandService` never reads that flag; the actual guard is the duplicate-based `sharesTnPostId` check. The new `ExpandServiceTest` sets the flag but it is decorative; the test passes because no live duplicate is seeded. During the cutover window a live unmerged email-era duplicate silently blocks the API-ingested message from rippling into other groups.
- Discarding every `group_id`-bearing copy (`GroupPostIngestionService.php:189`) assumes rippling substitutes for genuine multi-group placement. Rippling is capped and can be opted out of, so an item genuinely cross-posted to two Freegle-mapped groups can end up visible in only one.
- `EmailReplaySyncer::ensureUserExists()` (line 373) resolves groups via `Group::where('nameshort', $groupId)` where `$groupId` is TN's numeric group id from the post-log CSV. `EmailApiParityTest` has to seed groups with numeric nameshort values for this to work; against real CSV data every replayed email would report "Post to unknown group", invalidating the parity check on anything but synthetic fixtures.
- `CoverageVerifier::isPlaceable()` (line 183) trusts the archived email's coordinates header and short-circuits to unplaceable when it is missing or malformed, instead of checking the live API data. That is the exact coverage gap the verifier exists to catch.
- `UserChangesSyncer` (lines 68, 83): `!empty()` on `reply_time` and `about_me` drops a legitimate 0 reply time and a cleared bio.
- `RatingsSyncer` (lines 44, 58): unguarded `$rating['date']` access can warn and disturb the sync watermark; `if ($rating['rating'])` treats a genuine zero rating as a deletion.
- `createMessage()` (line 426) only preserves the TN post date when `$date` is a DateTime; all fixture/array-shaped posts pass ISO strings and silently get `now()`.
- The 2 req/s TN rate limit is enforced only in `PostSyncer`; `RatingsSyncer` and `UserChangesSyncer` hit the same API unthrottled in the same run with the same key, so a large backlog risks 429s or key suspension that then breaks the posts sync.
- `lastaccess`/`lastlocation` writes (`GroupPostIngestionService.php:234`, 418) use `DB::table('users')->update()` on an already-loaded Eloquent User, bypassing model events and Laravel Auditing, against the repo convention; sibling classes in this PR use `->save()`.
- `tn:parity-check` goes inert at the exact moment the flag disables the email path, while the 926-line duplicated ingestion logic stays. Every future fix to `handleGroupPost`/`createGroupPostMessage` has to be manually mirrored, with nothing automated to flag a miss.

### Dead code / leftovers

The superseded designs recorded in the plan really are gone: no trace of the repost/crosspost heuristic (`bumpAsRepost`, `haversineMeters`, radius constants), `PostRoutingLogger`, `FREEGLE_TN_SKIP_EMAIL_ROUTING`, or `dedupeApiCrosspostsAndReposts`, and `TNSyncCommand` was genuinely slimmed into the syncer classes. What remains:

- Generator cruft in `PublicApi/`: `git_push.sh`, `.travis.yml`, and ~40 files of generated docs and test stubs serve no purpose in this repo.
- The old `logEvent('tn-sync', 'post-*')` batch events still fire alongside the new routed stream. If keeping both is deliberate, the class docblock claiming the service "does NOT log to Loki itself" should say so; the plan said these would be reshaped.
- ~40 `TN-SYNC-TRACE` lines were added inside `IncomingMailService` despite the plan's rule that the email path stays untouched. They, and the replay/parity harness, are cutover scaffolding that will persist as noise unless removal is planned.
- Unrelated changes riding the branch, worth splitting out: the digest `seen_penalty` duplicate-config-key fix, the status-nuxt node 20 to 22 bump, and the yesterday 2FA admin API docs.

The cutover flag should not be flipped with the three confirmed bugs unfixed; all three fail silently.

PR #1031, opened 2026-07-10
Persist the in-chat compose draft across chat switches

edwh:
## What
The auto-save-draft feature that shipped covers only the **post reply pane**. The **in-chat compose box** (`ChatFooter.vue`) held the typed text in a plain local `ref`, so switching directly to another chat lost it.

Reported on Discourse: [/t/08-07-2026-freegle-code-changes-summary-ai-generated/9884/2](https://discourse.ilovefreegle.org/t/08-07-2026-freegle-code-changes-summary-ai-generated/9884/2) —
> "If I start a chat reply and then click on another chat (a realistic thing to do if I intend to copy & paste), the original draft has **not** been saved."

## How
- New `stores/chatdraft.js` — a small persisted store keyed by chat id (`{ [chatid]: { text, savedAt } }`), mirroring the reply pane's existing draft pattern in `stores/reply.js`. Persisted to `localStorage`, so drafts also survive a reload. Drafts older than 24h are discarded.
- `ChatFooter.vue` wiring:
  - **restore** the saved draft when a chat opens (only into an empty box, never clobbering live typing);
  - **save** debounced (500ms) as you type;
  - **flush** on unmount, so a draft typed right before switching isn't lost to the pending debounce;
  - **clear** on a successful send.

## Tests
`tests/unit/stores/chatdraft.spec.js` — save/restore, per-chat isolation, clear, empty/whitespace clears the draft, and 24h staleness pruning. Full local component-test run green (14241 passing, including these 7 new ones).

## Verification note
Store logic is unit-tested and eslint is clean; `/chats` renders 200 with the change deployed (no fatal error). The full click-through of restore-on-switch on dev-live is pending — the dev stack restarted and `apiv2-live`'s prod-DB tunnel is currently down (homepage/API 404s), so live chat data won't load. I'll complete the end-to-end browser check once the dev environment is back.


edwh:
Browser-verified end-to-end on dev-live now that the environment is back: typed a draft in a chat → it persisted to localStorage (`chatdraft` keyed by chat id) after the debounce → reloaded the page → the compose box repopulated with the exact draft text. That exercises the same re-mount + restore path a chat switch uses. Combined with the store unit tests, the fix is confirmed working.

PR #1046, opened 2026-07-12
fix(ai-images): Regenerate doesn't handle multi-item post names

edwh:
## Root Cause
`buildImagePrompt()` in `iznik-server-go/aiimage/aiimage.go` embeds `ai_images.name` verbatim as the subject of a `"single isolated ... single object only"` prompt, with no parsing for compound names like "sofa and a bed". The Regenerate button (`POST /api/admin/ai-images/:id/regenerate`) calls this exact function with the stored name, so it reproduces the same defective image every time — it never picks up on "and" to detect there are multiple items.

## Evidence
Failing test before the fix (`TestBuildImagePrompt_MultiItemName_DoesNotForceSingleObjectOnly`):
```
=== RUN   TestBuildImagePrompt_MultiItemName_DoesNotForceSingleObjectOnly
    aiimage_unit_test.go:285:
        Error Trace:  aiimage_unit_test.go:285
        Error:        "Product illustration: single isolated sofa and a bed centered on plain dark green background. Style: friendly cartoon white line drawing, moderate shading, cute and quirky, UK audience. The object sits alone on a simple surface or floats in space. Simple illustration style, clean lines, single object only. This is a common UK household or everyday item. Use American English terminology." should not contain "single object only"
        Test:         TestBuildImagePrompt_MultiItemName_DoesNotForceSingleObjectOnly
        Messages:     a compound item name must not be forced into a 'single object only' prompt
--- FAIL: TestBuildImagePrompt_MultiItemName_DoesNotForceSingleObjectOnly (0.00s)
FAIL
```
Confirmed against production data: 2,467 of 45,302 `ai_images` rows have `name LIKE '% and %'` — this is a common, non-edge-case shape for post names, not a rare occurrence.

## Fix
Added `splitItems()`, which splits a name on the word "and" (word-boundary, case-insensitive). `buildImagePrompt()` now checks the item count: for a single item, behaviour is unchanged (the existing "single isolated ... single object only" prompt). For 2+ items, it builds a prompt asking the model to show every item together, instead of a prompt that both names multiple things and simultaneously insists on drawing only one — the contradiction that was causing the model to draw just the first item.

This fixes the Regenerate flow specifically, since it always reads the stored `ai_images.name`, which is unaffected by the fact that the *original* cron-generated image (via `iznik-batch`'s `MessageIllustrationsService`/`PollinationsService`) may still only depict one item — that's a separate batch-job code path not touched by this PR.

## Other Call Sites
`iznik-batch/app/Services/PollinationsService.php` (`buildMessagePrompt`, `buildJobPrompt`) has the same `"single object only"` pattern, used only for the initial cron-generated illustration, not the Regenerate flow. Left out of scope for this PR — the reporter's concrete complaint was specifically that Regenerate doesn't pick up on "and", which is now fixed.

## Test
`iznik-server-go/aiimage/aiimage_unit_test.go`:
- `TestBuildImagePrompt_MultiItemName_DoesNotForceSingleObjectOnly` — fails before the fix (shown above), passes after.
- `TestBuildImagePrompt_SingleItemName_StillForcesSingleObjectOnly` — guards against regressing single-item behaviour.

Run: `go test ./aiimage/...` (pure function, no DB). Full package suite passes, including all pre-existing tests.

## Discourse
https://discourse.ilovefreegle.org/t/9630/53

### DIFF
--- a/iznik-server-go/aiimage/aiimage.go
+++ b/iznik-server-go/aiimage/aiimage.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #1046) — this fix did not pass and is not ready to merge.

- **Partial implementation / new regression**: The multi-item branch triggers on ANY name matching `(?i)\s+and\s+`, with no guard against names where " and " is part of a single item's name. Common Freegle inputs regress: "Black and Decker drill", "Pride and Prejudice" (books/DVDs/board games are routinely posted), "salt and pepper set", "cup an
- **Partial implementation**: Root cause is that the prompt hardcodes a single-object framing ("single isolated", "The object sits alone", "single object only") regardless of how many items the stored name lists. The fix detects only the literal separator " and ". A name of "sofa & bed", "sofa, bed, table", "sofa + bed", "sofa p

PR #1062, opened 2026-07-13
fix(rippling): scope MessageHistory approvedby to the group being viewed

edwh:
## Root Cause
`MessageHistory.vue`'s `approvedby` computed looped over **every** group a message belongs to and kept overwriting one shared result with whichever group's `approvedby` it saw last. Once rippling made posts multi-group, this meant a rippled-in copy (legitimately auto-approved, `approvedby IS NULL` on that group - by design, to keep moderation load off receiving groups) displayed the **origin** group's human-approving moderator's name instead, even in the per-group (`onlyGroupid`) view that the rest of the component already scopes correctly (added in 4a2b3dda5). A moderator viewing the destination group therefore saw "Approved by <name>" attribution implying a human had approved the post on *their* group, when nobody there had.

This is a display-only bug, not a moderation bypass: the rippled-in copy's own moderation state (`collection=Approved`, `approvedby=NULL`) is correct and intentional per the rippling design (posts approved on their origin propagate without per-group re-approval). Confirmed via Discourse 9886 ("Why do rippled out posts from members all come through as moderated members?") that this misattribution is a routine, widely-seen symptom of rippling being fully live, not a one-off.

## Evidence
Live production DB (read-only, numeric ids only):
```sql
SELECT COUNT(DISTINCT rippled.msgid) AS affected_messages
FROM messages_groups rippled
JOIN messages_groups origin
  ON origin.msgid = rippled.msgid AND origin.groupid <> rippled.groupid
  AND origin.approvedby IS NOT NULL
WHERE rippled.rippled_in = 1 AND rippled.collection = 'Approved'
  AND rippled.approvedby IS NULL AND rippled.deleted = 0
  AND rippled.arrival > DATE_SUB(NOW(), INTERVAL 14 DAY);
-- affected_messages = 10289
```
Example: msgid 120172510 rippled from origin group 21529 (approvedby=12279097) into destination groups 21250/21347/21516 (all `approvedby IS NULL`, `rippled_in=1`) - exactly the pattern that triggers the misattribution.

Failing test before fix (`tests/unit/components/MessageHistory.spec.js`):
```
FAIL  ... rippled-in copy attribution (Discourse 9890) > does not attribute the rippled-in copy to the origin group approving mod
AssertionError: expected '...' not to contain 'Approved by'
Received: "...Approved by Mod User..."
```

## Fix
Replaced the message-wide `approvedby` computed with a per-row `approvedByFor(group)` function, called against the specific `group` being rendered in the `v-for`. A group's row now only shows "Approved by X" when *that group's own* `messages_groups.approvedby` is set - never borrowed from a sibling group on the same message.

## Other Call Sites
`grep -rn "approvedby" components/ modtools/` - the only other reference is the pre-fetch loop (lines ~172-179) that warms the `userStore` cache for every group's approver; it doesn't render anything and fetching a few extra user records is harmless, so it was left as-is.

## Test
- File: `iznik-nuxt3/tests/unit/components/MessageHistory.spec.js`
- Command: `POST /api/tests/vitest {"filter":"MessageHistory.spec.js"}` via the status API
- Proves fail-before (misattributed "Approved by" shown for the un-approved destination group) / pass-after (only the origin group's own row shows its approver; the destination group shows none)
- Full `npm run test` (vitest) suite: 14270 passed, 0 failed after the fix

## Discourse
https://discourse.ilovefreegle.org/t/9890/1


edwh:
Closing after re-diagnosing the report against the actual screenshot in the ticket.

## What was actually reported
Discourse [/t/9890/1](https://discourse.ilovefreegle.org/t/9890/1) — *"Rippled out post not approved by anyone but has come through as a moderated member"*. The screenshot is ModTools `/messages/approved` showing:

- OFFER #118852950 rippled into **Uttlesford Freegle**, with the blue banner: *"This post was rippled in from a neighbouring community and **automatically approved** onto Uttlesford Freegle to keep moderation load low."*
- The member panel showing the poster (jacqui rat) with moderation status **"Moderated"**.

The confusion is: the post went to **Approved** (not Pending) even though the member shows as **Moderated**. That is **correct by design** — rippled-in posts auto-approve regardless of the poster's per-group moderation status, exactly as the banner explains. In the related [/t/9886/3](https://discourse.ilovefreegle.org/t/9886/3) the reporter also partly retracted (*"Err may have this wrong"*).

## Why this PR doesn't address it
This PR rescopes MessageHistory's **"Approved by X"** attribution line. But:

- The reporter explicitly said the post was *"not approved by anyone"* — i.e. she saw **no** approver line. The "Approved by X" misattribution this PR fixes is the opposite of what she experienced, and that line isn't even present in the reported `/messages/approved` view.
- The element she actually flagged — the **member "Moderated" status dropdown** on a rippled-in poster — is untouched by this PR (the diff is MessageHistory.vue only).

So the reported behaviour is working-as-designed (no status bug), and this change fixes a different, unreported thing. Closing as a mis-diagnosis.

## Note
The per-group `approvedByFor(group)` scoping here is a reasonable tidy-up on its own (a message-wide `approvedby` computed could borrow a sibling group's approver), but it is orthogonal to 9890 and not worth carrying as "the fix" for this ticket. It can be reconsidered independently if desired. If we want to act on the real report, the change would be to how a rippled-in poster's moderation status is presented to the destination mod (per 9886's request to honour the member's origin-group setting) — a separate piece of work.

PR #1065, opened 2026-07-13
fix(ripple): don't show cross-group held-by wording on an already-approved copy

edwh:
## Root Cause
`ModMessage.vue` and `ModMessageButtons.vue` read the message-level `messages.heldby` field (kept only for backwards compatibility, predating rippling) instead of the per-group `messages_groups.heldby`. That legacy field gets set whenever **any** group holds **any** copy of a rippled post. So a mod who approved a post on their own group, which then rippled to another group and was held/rejected there, saw the post on their own (Approved, unheld) copy displayed as "Held by X. Please check with them before releasing it" - nonsensical, since their copy was never held and had already been released. The same global field also drove the Hold/Release button toggle in `ModMessageButtons.vue`, so a mod could be offered "Release" for a copy that wasn't actually held on their group.

Confirmed live in production (read-only, numeric id only): message id 120950218 is `Approved` on 8 groups while `Held` on a 9th (rippled-to) group, with the legacy `messages.heldby` field set globally - reproducing exactly the reporter's scenario.

This is a **re-attempt**: the first attempt (same root cause) was blocked in adversarial review for the three gaps below. Each is now resolved with concrete evidence, not assertion.

## Evidence (failing-test output before the fix)
AssertFlip - reverted the production-code fix (kept the new tests) and re-ran; every new test for the fixed behaviour fails on the pre-fix code, and nothing else regresses:
```
Test Files  4 failed | 2 passed (6)
     Tests  15 failed | 219 passed (234)

 FAIL  tests/unit/composables/rippleStatus.spec.js > findMessageGroupRow > (5 tests)
       TypeError: (0 , findMessageGroupRow) is not a function
 FAIL  tests/unit/composables/rippleStatus.spec.js > messageGroupHeldBy > (7 tests)
       TypeError: (0 , messageGroupHeldBy) is not a function
 FAIL  tests/unit/components/modtools/ModMessage.spec.js > Cross-group ripple hold (Discourse 9904)
       > fails safe (still shows the banner, still hides the buttons) when the context group cannot be matched to any row
       AssertionError: expected 'OFFER: Test Item...' to contain 'Held by'
 FAIL  tests/unit/components/modtools/ModMessageButtons.spec.js > pending message buttons
       > fails safe (offers release, not hold) when groupid matches no row but another group holds its copy
       AssertionError: expected true to be false
 FAIL  tests/unit/components/ModMessageButton.spec.js > confirmButton (Discourse 9904 follow-up)
       > does not require confirmation when a DIFFERENT group holds its own copy
```
All 15 pass again once the fix is restored; full local run after restoring: `Test Files 670 passed (670) / Tests 14300 passed | 4 skipped (14304)` (whole `tests/unit/` suite, no regressions).

## Fix
- `composables/rippleStatus.js`: added two shared helpers, `findMessageGroupRow(groups, groupid)` (tolerant numeric/string groupid coercion) and `messageGroupHeldBy(groups, groupid)` (the per-group `heldby` for the exact row, **failing SAFE** to "is ANY copy of this post held" when the exact row can't be identified, rather than silently returning unheld).
- `ModMessage.vue`: `heldbyId`/`heldbyName` now derive from `messageGroupHeldBy(message.groups, currentGroupid)` instead of `contextGroup.value?.heldby` (which itself falls back to an arbitrary `groups[0]` for unrelated display purposes and was not safe to reuse for the hold check). `contextGroup` and `rippleProximity` now use the shared `findMessageGroupRow` for their own exact-match lookups.
- `ModMessageButtons.vue`: `heldByThisGroup` now calls the same `messageGroupHeldBy` helper instead of its own inline `.find()`.
- `ModMessageButton.vue`: `confirmButton` (extra confirmation before an action) now also uses `messageGroupHeldBy` instead of the legacy `message.value?.heldby` - see Other Call Sites below.

## Other Call Sites
Grepped `heldby` across `iznik-nuxt3`/`modtools` for every message-related consumer of the field:
- `ModMessage.vue` (`heldbyId`/`heldbyName`) - fixed.
- `ModMessageButtons.vue` (`heldByThisGroup`, gates the Hold/Release button) - fixed.
- `ModMessageButton.vue` (`confirmButton`) - fixed in this re-attempt (previously left as-is; now uses the same shared helper so it can't drift from the other two).
- `ModMessages.vue` and the pending-queue views do not reference `heldby` at all - no held-state filter exists there to check.
- `hasCollection()` in `ModMessageButtons.vue` (used for `pending`/`approved`/`spam`) checks whether **any** group of the message has a given collection, not specifically the group being administered - a related but separate scoping issue with broader implications. Left out of scope to keep this fix focused.
- `member.heldby` / `membership.heldby` / `admin.heldby` (`ModMember.vue`, `ModMemberButtons.vue`, `ModMemberReviewActions.vue`, `ModAdmin.vue`) are unrelated: those rows are already inherently per-group/per-object with no rippling/multi-group concept, so they don't share this bug.

## Test
- Files: `iznik-nuxt3/tests/unit/composables/rippleStatus.spec.js`, `tests/unit/components/modtools/ModMessage.spec.js`, `tests/unit/components/modtools/ModMessageButtons.spec.js`, `tests/unit/components/modtools/ModMessageButtonsHoldRelease.spec.js` (new), `tests/unit/components/ModMessageButton.spec.js`, `tests/unit/components/modtools/ModMessageButton.spec.js`, `tests/unit/stores/message.spec.js`.
- Command: `npx vitest run tests/unit/` (whole unit suite) - `670 passed (670)` files, `14300 passed | 4 skipped` tests.
- Proves fail-before/pass-after per the Evidence section above (AssertFlip: reverted just the four production files via `git stash`, confirmed the new tests fail, restored, confirmed they pass).

## Review Blockers Resolved
**Blocker 1 (data contract unproven)** - apiv2 does return `groups[].heldby`, traced end-to-end, not just asserted by fixtures:
- Go struct: `iznik-server-go/message/messageGroup.go:27` - `Heldby *uint64 \`json:"heldby,omitempty"\`` on `MessageGroup` (table `messages_groups`).
- SQL select: `iznik-server-go/message/message.go:487` - `SELECT groupid, msgid, arrival, collection, autoreposts, approvedby, heldby, spamtype, ... FROM messages_groups WHERE msgid = ? AND deleted = 0`, scanned into `messageGroups`, assigned via `message.MessageGroups = messageGroups` (line 618).
- Serialisation: `Message.MessageGroups []MessageGroup \`gorm:"-" json:"groups"\`` (message.go:212) - so the JSON `groups` array is exactly these rows, `heldby` included.
- Call path: `GetMessages` (the `/api/message/:ids` handler) calls `GetMessagesByIds`, which runs the select above (message.go:377-392). The frontend's `messageStore.fetch`/`fetchMT` call `$getv2('/message/' + id)` (`api/MessageAPI.js:5-11`) - i.e. this exact endpoint.
- No backend change was needed - the field was already there; the bug was purely in which field the frontend read.

**Blocker 2 (write path untouched)** - `messageStore.hold()`/`release()` (`stores/message.js:801-814`, unmodified by this fix) already call the API then **fully refetch the message** via `api.message.fetchMT({id})` and replace `this.list[message.id] = message`. Since `message` in every consuming component is `computed(() => messageStore.byId(props.messageid))` and `byId` is a reactive Pinia getter (`(state) => (id) => state.list[id]`), the per-group `heldby` - and every computed reading it - updates automatically, no separate refetch required. Proved with:
- A store-level test (`tests/unit/stores/message.spec.js`) asserting `store.byId(id).groups[0].heldby` reflects the post-refetch value after `hold()`/`release()`.
- A full UI-level test (`ModMessageButtonsHoldRelease.spec.js`, new) using the **real** Pinia message store (only `~/api` is mocked) and the **real**, unstubbed `ModMessageButton`: clicking the rendered Hold button calls `messageStore.hold()`, and once the mocked refetch resolves the button set flips from Hold to Release (and vice versa for Release), with no full page reload.

**Blocker 3 (fail-open guard)** - `messageGroupHeldBy` (see Fix) explicitly fails SAFE: if the group being administered can't be matched to any row in `message.groups` (e.g. `currentGroupid`/`props.groupid` doesn't line up with any group - stale context, message still loading), it falls back to `groups.find(g => g.heldby)` - "is ANY copy of this post held" - rather than silently returning null/unheld. Covered by dedicated tests for `contextGroup === undefined`/no-match:
- `rippleStatus.spec.js`: `messageGroupHeldBy` unit tests for "falls back to any held copy when the groupid matches no row" and the null/no-match/no-groups cases.
- `ModMessage.spec.js`: "fails safe (still shows the banner, still hides the buttons) when the context group cannot be matched to any row".
- `ModMessageButtons.spec.js`: "fails safe (offers release, not hold) when groupid matches no row but another group holds its copy".

## Discourse

### DIFF
--- a/iznik-nuxt3/composables/rippleStatus.js
+++ b/iznik-nuxt3/composables/rippleStatus.js
@@ -148,3 +148,51 @@ export function isHomeGroup(group, groups) {
   const gid = group.groupid ?? group.id
   return gid != null && parseInt(gid) === homeId
 }
+
+/**
+ * Find the messages_groups row for a specific group, tolerant of a numeric or
+ * numeric-string groupid on either side (API rows and component props aren't
+ * guaranteed to agree on type). Centralised so the coercion can't drift between
+ * components and have them silently disagree about which row is "the" group
+ * being administered.
+ *
+ * @param {Array<{groupid:number|string}>} groups message.groups
+ * @param {number|string} groupid
+ * @returns {object|undefined}
+ */
+export function findMessageGroupRow(groups, groupid) {
+  if (!Array.isArray(groups) || groupid === null || groupid === undefined) {
+    return undefined
+  }
+  const gid = parseInt(groupid)
+  if (Number.isNaN(gid)) return undefined
+  return groups.find((g) => g && parseInt(g.groupid) === gid)
+}
+
+/**
+ * The per-group hold state (messages_groups.heldby) for the group being
+ * administered - NOT messages.heldby, the legacy message-level field that gets
+ * set whenever ANY group holds ANY copy of a rippled post (Discourse 9904).
+ * Returns the raw heldby value: a numeric user id (Go API) or a
+ * {id, displayname} object (legacy PHP API), or null if this copy isn't held.
+ *
+ * Fails SAFE: if the specific group row can't be pinned down (message still
+ * loading, or groupid doesn't match any row this post is on), falls back to
+ * "is ANY copy of this post held" rather than silently reporting unheld - a
+ * false "not held" would let a mod release/reject a copy someone else is
+ * actively holding on another group, which is worse than an unnecessary
+ * confirmation.
+ *
+ * @param {Array<{groupid:number|string, heldby?:*}>} groups message.groups
+ * @param {number|string} groupid the group being administered
+ * @returns {*} heldby value, or null
+ */
+export function messageGroupHeldBy(groups, groupid) {
+  if (!Array.isArray(groups) || !groups.length) return null
+
+  const exact = findMessageGroupRow(groups, groupid)
+  if (exact) return exact.heldby || null
+
+  const heldRow = groups.find((g) => g && g.heldby)
+  return heldRow ? heldRow.heldby : null
+}
--- a/iznik-nuxt3/modtools/components/ModMessage.vue
+++ b/iznik-nuxt3/modtools/components/ModMessage.vue
@@ -323,7 +323,7 @@
                 at
                 {{ datetimeshort(message.outcomes[0].timestamp) }}
               </NoticeMessage>
-              <div v-if="message.heldby">
+              <div v-if="heldbyId">
                 <NoticeMessage variant="warning" class="mb-2">
                   <p v-if="me.id === heldbyId">
                     You held this. Other people will see a warning to check with
@@ -721,7 +721,7 @@
         </b-row>
       </b-card-body>
       <b-card-footer v-if="!noactions && expanded">
-        <div v-if="message.heldby && heldbyId !== myid">
+        <div v-if="heldbyId && heldbyId !== myid">
           This message is held by someone else. The buttons are hidden so you
           don't click them by accident. Please check with them before releasing
           the message.
@@ -734,11 +734,7 @@
           This message needs editing so that we know where it is.
         </NoticeMessage>
         <div
-          v-if="
-            pending &&
-            (!message.heldby || (message.heldby && heldbyId === myid)) &&
-            !editing
-          "
+          v-if="pending && (!heldbyId || heldbyId === myid) && !editing"
           class="text-end mb-1"
         >
           <b-button variant="danger" @click="spamReport">
@@ -746,10 +742,7 @@
           </b-button>
         </div>
         <ModMessageButtons
-          v-if="
-            (!message.heldby || (message.heldby && heldbyId === myid)) &&
-            !editing
-          "
+          v-if="(!heldbyId || heldbyId === myid) && !editing"
           :messageid="message.id"
           :groupid="currentGroupid"
           :modconfigid="configid"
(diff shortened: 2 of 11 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #1065) — this fix did not pass and is not ready to merge.

- **Incomplete diff / unverified API contract**: The fix re-sources hold state from the per-group row (`contextGroup.value?.heldby` in ModMessage.vue, `groups.find(...).heldby` in ModMessageButtons.vue), but the diff contains ZERO backend/API changes and no test that exercises a real API payload. Nothing here proves that `messages_groups.heldby` i
- **Fail-open on unresolved group context**: Both new guards silently degrade to 'not held' when the group can't be resolved. ModMessageButtons: `if (!groups || !props.groupid) return false` -> Hold button shown. ModMessage: `contextGroup.value?.heldby` -> `heldbyId === null` -> the footer 'This message is held by someone else' block and the M

edwh:
Auto-closed by the monitor's adversarial review (PR #1065) — this fix did not pass and is not ready to merge.

- **Partial implementation / unswept sibling gate**: ModMessageButtons.vue received exactly ONE gate change: the Hold button's `v-if="!message.heldby"` → `v-if="!heldByThisGroup"`. A `!message.heldby` gate on a Hold button virtually always has a mirrored `v-if="message.heldby"` gate on the Release/unhold button (ModMessage's own banner text — 'check w

PR #1068, opened 2026-07-14
fix(modtools): don't show cross-group held-by wording on an already-approved copy

edwh:
## Root Cause
ModTools read the legacy, cross-group `messages.heldby` field to decide whether the message being viewed was held - instead of the per-group `messages_groups.heldby` row for the group actually being moderated. That legacy field is set whenever **any** group holds **any** copy of a rippled post, so a mod who approved a post on their own group, which then rippled to another group and was held/rejected there, saw their own (Approved, unheld) copy display "Held by X. Please check with them before releasing it" - nonsensical, since their copy was never held. The same unscoped field also drove the Hold/Release button toggle and the "confirm before acting" check, so a mod could in principle be offered Release for a copy that was never held on their group.

This is a fresh re-attempt of a previously-blocked PR (#1065, closed). That review found: the Release/unhold gate in `ModMessageButtons.vue` was left on the legacy field (so Hold+Release could render together); the heldby→id coercion was duplicated and inconsistent between components; `ModMessageButtons.vue` had no direct test coverage; the fail-safe fallback named an arbitrary, possibly-unreachable mod; and the Go API still served the cross-group field, so every reader (present and future) inherited the bug. All five are addressed below.

## Evidence
AssertFlip - reverted the four production files (composable + 3 components), kept only the new/updated tests, ran the affected specs:
```
Test Files  4 failed | 1 passed (5)
     Tests  12 failed | 221 passed (233)

 FAIL  tests/unit/composables/rippleStatus.spec.js > messageGroupHeldById (Discourse 9904) > returns null when this group's own row is not held, even though another group holds its copy
 FAIL  tests/unit/composables/rippleStatus.spec.js > messageGroupHeldById (Discourse 9904) > returns the holder id when this group's own row is held
 FAIL  tests/unit/composables/rippleStatus.spec.js > messageGroupHeldById (Discourse 9904) > coerces a numeric-string groupid and a numeric-string heldby identically
 FAIL  tests/unit/composables/rippleStatus.spec.js > messageGroupHeldById (Discourse 9904) > coerces an object-shaped heldby ({id}) to its numeric id
 FAIL  tests/unit/composables/rippleStatus.spec.js > messageGroupHeldById (Discourse 9904) > returns undefined (not a guessed holder) when no row matches the groupid
 FAIL  tests/unit/composables/rippleStatus.spec.js > messageGroupHeldById (Discourse 9904) > returns undefined for missing/empty groups or unresolvable groupid
 FAIL  tests/unit/components/modtools/ModMessage.spec.js > ModMessage > Held message > does not show a held banner when this group is unheld even though the legacy cross-group message.heldby is set by a rippled-to group holding its own copy (Discourse 9904)
 FAIL  tests/unit/components/modtools/ModMessageButton.spec.js > ModMessageButton > confirmButton computed > returns true when message is held and action is not spam or delete
 FAIL  tests/unit/components/modtools/ModMessageButton.spec.js > ModMessageButton > confirmButton computed > returns false when message is held but action is spam
 FAIL  tests/unit/components/modtools/ModMessageButton.spec.js > ModMessageButton > confirmButton computed > returns false when message is held but action is delete
 FAIL  tests/unit/components/modtools/ModMessageButton.spec.js > ModMessageButton > confirmButton computed > does not confirm when a DIFFERENT group holds its own copy (Discourse 9904)
 FAIL  tests/unit/components/modtools/ModMessageButtons.spec.js > ModMessageButtons > pending message buttons > offers Hold (not Release) when a DIFFERENT group holds its own copy - the exact Discourse 9904 scenario
```
All fail on the pre-fix (reverted) code; all pass once the fix is restored. Full local run after restoring: `npx vitest run tests/unit/` → `Test Files 669 passed (669)` / `Tests 14284 passed | 4 skipped (14288)` - no regressions.

Go side: `go build ./...` and `go vet ./message/...` are clean, and the new Go test (`TestGetMessageOmitsLegacyCrossGroupHeldby`) compiles (`go test ./test/... -run TestGetMessageOmitsLegacyCrossGroupHeldby -c`). I could not run the Go suite against a live DB in this sandboxed session (no bound MySQL test instance was safely reachable without touching another active worktree's shared containers) - CI will run the full Go suite on this PR before merge, per the usual gate.

## Fix
- **`composables/rippleStatus.js`**: added `messageGroupHeldById(message, groupid)` - the single place that finds the matching `messages_groups` row for the group in context and coerces `heldby` (numeric, numeric-string, or `{id}` object) to a plain user id. Returns `null` when that group's row is confirmed unheld, `undefined` when the row can't be identified at all (never guesses).
- **`ModMessage.vue`**: `heldbyId`/`heldbyName`/the held banner/the button-hiding logic all now derive from `messageGroupHeldById(message.value, currentGroupid.value)` instead of `message.heldby`. Added a narrow, distinct fallback (`heldByUnknownGroup`) for when the group's own row genuinely can't be resolved (e.g. still loading) but the legacy field says it's held somewhere - shows generic "Held on another group" copy, naming no one, rather than an arbitrary (possibly unreachable) mod.
- **`ModMessageButtons.vue`**: the Hold/Release toggle (`heldByThisGroup`) now uses the same helper. Hold and Release are rendered via `v-if`/`v-else` on one boolean, so they're structurally mutually exclusive - fixes the previous attempt's gap where Release stayed on the legacy field.
- **`ModMessageButton.vue`**: `confirmButton` (extra confirm-before-acting dialog) now uses the same helper too, so it can't drift from the other two.
- **`iznik-server-go/message/message.go`**: the single-message `Message.Heldby` field (the cross-group `messages.heldby` column) is no longer serialised (`json:"-"`) - root-caused server-side so every current and future JSON consumer is forced onto the correctly-scoped `groups[].heldby`.

## Other Call Sites
Grepped `heldby` across `iznik-nuxt3` (not just `modtools`) for every message-related consumer:
- `ModMessage.vue` (banner text, button-hiding, `heldbyId`/`heldbyName`, the held-user fetch in `onMounted`) - fixed.
- `ModMessageButtons.vue` (`heldByThisGroup`, gates Hold vs Release) - fixed.
- `ModMessageButton.vue` (`confirmButton`) - fixed.
- `ModMessages.vue` and the pending-queue list views don't reference `heldby` at all.
- `member.heldby` / `membership.heldby` / `admin.heldby` (`ModMember.vue`, `ModMemberButtons.vue`, `ModMemberReviewActions.vue`, `ModAdmin.vue`, `stores/spammer.js`, `stores/member.js`) are unrelated: those rows are inherently per-object with no rippling/multi-group concept, so they don't share this bug - left untouched.
- Go: `message_list.go`'s `MessageGroupInfo.Heldby` (used by the list/queue endpoint) was already correctly per-group - no change needed there. Only the single-message `Message.Heldby` (used by `GET /api/message/:ids`, which the frontend calls via `messageStore.fetch`/`fetchMT`) carried the cross-group leak.

## Test
- Files: `iznik-nuxt3/tests/unit/composables/rippleStatus.spec.js`, `tests/unit/components/modtools/ModMessage.spec.js`, `tests/unit/components/modtools/ModMessageButtons.spec.js`, `tests/unit/components/modtools/ModMessageButton.spec.js`, `iznik-server-go/test/message_test.go` (`TestGetMessageOmitsLegacyCrossGroupHeldby`).
- Command: `npx vitest run tests/unit/` (whole unit suite) - `669 passed (669)` files, `14284 passed | 4 skipped` tests.
- Proves fail-before/pass-after via AssertFlip (see Evidence above).

## Discourse

### DIFF
--- a/iznik-nuxt3/composables/rippleStatus.js
+++ b/iznik-nuxt3/composables/rippleStatus.js
@@ -148,3 +148,44 @@ export function isHomeGroup(group, groups) {
   const gid = group.groupid ?? group.id
   return gid != null && parseInt(gid) === homeId
 }
+
+function normaliseHeldById(h) {
+  if (h === null || h === undefined) return null
+  if (typeof h === 'object') return h.id ?? null
+  const n = parseInt(h)
+  return Number.isNaN(n) ? null : n
+}
+
+/**
+ * The id of the moderator holding THIS message on THIS group - i.e. the group's own
+ * messages_groups row - never the legacy cross-group messages.heldby, which is set
+ * whenever ANY group holds ANY copy of a rippled post. Reading that legacy field leaked a
+ * hold placed on one group's copy into every other group's view of the same message: a mod
+ * saw their own already-approved copy display "Held by X, check with them before releasing
+ * it" even though their copy was never held (Discourse 9904).
+ *
+ * This is the ONLY place that should read/coerce a message-level heldby - every consumer
+ * (the held-by banner, the Hold/Release button toggle, the confirm-before-acting check)
+ * must call this rather than inspecting message.heldby or a group row directly, so they
+ * can't drift out of sync with each other.
+ *
+ * Returns:
+ *  - a numeric user id if the group's own copy is held
+ *  - null if a matching group row exists and is definitely NOT held
+ *  - undefined if the group's row can't be identified at all (message still loading, or
+ *    groupid doesn't correspond to any group the post is on) - callers must not guess who
+ *    holds it in this case, since that mod may be on a group the reader can't act on.
+ *
+ * @param {{groups?: Array<{groupid:number|string, heldby?:number|string|{id:number}|null}>}} message
+ * @param {number|string} groupid
+ * @returns {number|null|undefined}
+ */
+export function messageGroupHeldById(message, groupid) {
+  const groups = message?.groups
+  if (!Array.isArray(groups) || !groups.length) return undefined
+  const gid = parseInt(groupid)
+  if (Number.isNaN(gid)) return undefined
+  const row = groups.find((g) => parseInt(g.groupid) === gid)
+  if (!row) return undefined
+  return normaliseHeldById(row.heldby)
+}
--- a/iznik-nuxt3/modtools/components/ModMessage.vue
+++ b/iznik-nuxt3/modtools/components/ModMessage.vue
@@ -323,7 +323,7 @@
                 at
                 {{ datetimeshort(message.outcomes[0].timestamp) }}
               </NoticeMessage>
-              <div v-if="message.heldby">
+              <div v-if="heldByThisGroup">
                 <NoticeMessage variant="warning" class="mb-2">
                   <p v-if="me.id === heldbyId">
                     You held this. Other people will see a warning to check with
@@ -342,6 +342,13 @@
                     release
                     label="Release"
                   />
+                </NoticeMessage>
+              </div>
+              <div v-else-if="heldByUnknownGroup">
+                <NoticeMessage variant="warning" class="mb-2">
+                  <p>
+                    Held on another group. Please check before releasing it.
+                  </p>
                 </NoticeMessage>
               </div>
             </div>
@@ -721,7 +728,7 @@
         </b-row>
       </b-card-body>
       <b-card-footer v-if="!noactions && expanded">
-        <div v-if="message.heldby && heldbyId !== myid">
+        <div v-if="heldAndNotByMe">
           This message is held by someone else. The buttons are hidden so you
           don't click them by accident. Please check with them before releasing
           the message.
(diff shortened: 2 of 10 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #1068) — this fix did not pass and is not ready to merge.

- **Incomplete diff / unverified API contract**: The entire fix pivots on `groups[].heldby` existing on the message payload, but nothing in the diff adds, serialises, or even cites that field. No Go/API change, no type/fixture change, no store change — the only places `groups[].heldby` appears are hand-written test mocks (ModMessage.spec.js lines 
- **Regression: holder locked out of their own held message**: `heldAndNotByMe` returns true whenever `heldByUnknownGroup` is true, with no check of who the legacy holder is. Previously the footer/buttons guard was `message.heldby && heldbyId === myid` → the mod who held the message still saw ModMessageButtons (including Release). Now, if the group row can't be
- **Safety regression: confirm guard fails open**: ModMessageButton.vue `confirmButton` is now `typeof messageGroupHeldById(message.value, groupid.value) === 'number'`. The composable's own docblock says callers 'must not guess' on `undefined` — but this caller silently treats `undefined` (row unidentifiable / `groupid` unresolved so `parseInt` → Na

PR #1069, opened 2026-07-14
fix(chat): dedupe retried chat message sends with a client idempotency key

edwh:
## Root Cause

Two combined problems, both real:

1. **`useFetchRetry` retried non-idempotent writes on an ambiguous outcome.** `fetchRetry()` (`composables/useFetchRetry.js`), which backs every mutating call through `BaseAPI.$postv2`/`$putv2`/`$patchv2`/`$delv2`, retried *any* request on a network error, a 5xx, or a 2xx it couldn't parse - regardless of method. For a chat send (`ChatAPI.send()` → `POST /chat/{id}/message`), those outcomes don't prove the write failed: the POST may already have been committed server-side, and retrying resends the identical POST.
2. **The server-side write had no idempotency protection**, so any duplicate POST - client retry, double-click, or a second tab - created a second, genuinely-persisted `chat_messages` row. Both rows briefly showed to the sender (topic 9913: two copies of a just-sent chat message) until the pre-existing `cleanup:chat-duplicates` cron (`PurgeService::deduplicateChatMessages`) later removed one - matching the report's "there is now only one copy" follow-up.

This is a **re-attempt of PR #1063**, which was blocked by adversarial review for: silently swallowing legitimate identical resends via a 60s content-hash dedupe (a genuine second "ok"/"?" would have been dropped); a non-atomic SELECT-then-INSERT that didn't actually close the concurrent-retry race; no caller-side handling of the new fail-fast behaviour; and an unchecked `db.Raw(...).Scan()` read that could hit a lagging replica. This PR replaces the content-hash approach entirely with a **client-generated idempotency key** enforced by a DB-level unique index, fixing all four points (see Fix).

## Evidence

`composables/useFetchRetry.js` had no method awareness at all - it retried a `POST` exactly like a `GET` on any 5xx/network-error/unparseable-2xx. New regression test (`tests/unit/composables/useFetchRetry.spec.js`) run against the **pre-fix** code:

```
FAIL  useFetchRetry > mutating requests are not retried on an ambiguous outcome (#9913) > does not retry a POST on a network error ...
FAIL  useFetchRetry > ... does not retry a POST that returns a 200 it cannot parse as JSON ...
FAIL  useFetchRetry > ... does not retry a POST on a 500
FAIL  useFetchRetry > ... does not retry a PUT/PATCH/DELETE on a network error (x3)
FAIL  useFetchRetry > ... treats a lower-case method string as mutating too
FAIL  useFetchRetry > ... derives the method from a Request-like input ...
FAIL  useFetchRetry > ... does not block on connectivity before rejecting a mutating request — Test timed out in 30000ms
 Tests  9 failed | 21 passed (30)
```
All 30 pass after the fix (`isSafeMethod` restricting auto-retry to GET/HEAD/OPTIONS).

On the Go side, master's `CreateChatMessage` calls `db.Create(&payload)` unconditionally with no idempotency field on the struct at all, so two sequential POSTs with identical content **always** create two separate autoincrement rows today - that's what the new `TestCreateChatMessageIdempotencyKeyReturnsExistingRow`/`TestCreateChatMessageIdempotencyKeyConcurrent` tests assert must NOT happen after the fix. I could not run `go test` against a real MySQL DB from this isolated worktree (no `.env`/bound containers here); `go build ./...`, `go vet ./...` and `gofmt -l` (on the added lines - see Test) are clean, and CI is the verification gate for these, per this repo's established practice for this constraint.

## Fix

**`iznik-batch/database/migrations/2026_07_14_000001_add_idempotency_key_to_chat_messages.php`** (+ companion prod `.sql`): adds a nullable `chat_messages.idempotency_key VARCHAR(64)` and a `UNIQUE (chatid, userid, idempotency_key)` index. Nullable so older/cached clients that send no key are unaffected (MySQL never treats two NULLs as equal in a unique index) - no backfill needed on the existing table.

**`iznik-server-go/chat/chatmessage.go` (`CreateChatMessage`):** replaced `db.Create(&payload)` with an atomic `INSERT ... ON DUPLICATE KEY UPDATE id = LAST_INSERT_ID(id)` via the raw `sql.DB` (mirrors `GetOrCreateUser2UserChat`'s existing pattern for `chat_rooms`) when a key is supplied, falling back to a plain insert otherwise. This is atomic (no SELECT-then-INSERT race - fixes blocker B2), the `sqlResult.LastInsertId()` error is checked (fixes B4a), and `db.DB()` always returns the write connection even with a read replica configured, so there's no separate SELECT to race a lagging replica at all (fixes B4b by construction). A deliberate resend (client sends a fresh key) always creates a new row (fixes B1) - proven by `TestCreateChatMessageDeliberateResendCreatesNewRow`.

**`composables/useFetchRetry.js`:** renamed `retryable` → `isSafeMethod`, extended to `GET`/`HEAD`/`OPTIONS`; only these auto-retry on an ambiguous outcome. `waitForOnline()` is only awaited when the method is safe, so a mutating request fails fast instead of hanging until reconnect.

**`stores/chat.js` (`send`/`report`) + `components/ChatFooter.vue`:** every `chat.send()` call now carries an idempotency key (`generateUUID()` from `composables/useTrace.js`, reused rather than duplicated). `ChatFooter`'s composer keeps the SAME key across a manual "tap Send to retry" of the same not-yet-sent text (`getSendIdempotencyKey`, new pure helper in `composables/useChat.js`) - editing the text first gets a fresh key. On failure the typed text is retained (never cleared in the catch path) and a visible error ("...Tap Send to try again.") is shown - this is the caller-side handling blocker B3 required.

## Other Call Sites

- `ourFetch = fetchRetry(fetch)` in `api/BaseAPI.js` is the single shared instance behind `$getv2`/`$postv2`/`$patchv2`/`$putv2`/`$postFormv2`/`$delv2`, used by all 37 API classes that `extends BaseAPI` - the safe-method restriction protects every mutating endpoint app-wide. No separate service-worker or native-app fetch/retry layer exists (ModTools + Freegle share one Capacitor wrapper of the same web bundle).
- Other `chat_messages` INSERT sites, deliberately NOT touched (server-generated messages, not user chat sends, no idempotency risk from this bug's mechanism): `chat/chatroom.go:1402` (nudge), `message/message.go:4663`, `message/bulkItem.go:277,407`, `message/helper.go:584`, `amp/amp.go:468`, and `CreateChatMessageLoveJunk` (separate handler/external integration).

## Test

- `iznik-nuxt3/composables/useFetchRetry.spec.js` - 30 tests (9 new, verified to fail against pre-fix code - see Evidence); `tests/unit/stores/chat.spec.js` - 44 tests (2 new + 4 updated for the new payload field); new `tests/unit/composables/useChatSendIdempotency.spec.js` (4 tests, the retry-key-reuse pure function) and `tests/unit/components/ChatFooterSendRetry.spec.js` (4 tests, source-tied since ChatFooter.vue can't be mounted in this suite - same constraint as `ChatFooterDraftPersistence.spec.js`). Full local suite: `npx vitest run` → 671 files / 14296 passed, 4 pre-existing skips.
- `iznik-server-go/test/chat_test.go` - 6 new tests: same-key retry returns existing row + single row; different-key resend creates a second row; two concurrent same-key requests collapse to one row (mirrors `TestPutChatRoomConcurrent`); no-key backward compatibility; key-too-long validation. `go build ./...`/`go vet ./...`/`gofmt -l` (added lines) clean; CI runs these against the real test DB.

## Discourse

### DIFF
--- a/iznik-batch/database/migrations/2026_07_14_000001_add_idempotency_key_to_chat_messages.php
+++ b/iznik-batch/database/migrations/2026_07_14_000001_add_idempotency_key_to_chat_messages.php
@@ -1,0 +1,54 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+/**
+ * Client-generated idempotency key for chat message sends (Discourse #9913: two
+ * copies of a just-sent chat message shown transiently - a retried/duplicated
+ * POST created a second, genuinely-persisted row).
+ *
+ * Replaces the rejected 60s content-hash dedupe approach (which silently swallowed
+ * legitimate identical resends like a deliberate second "ok" or "?") with a
+ * client-supplied key: the same logical send always carries the same key, a
+ * genuinely new message always gets a new one. The unique index makes at-most-once
+ * DB-enforced and atomic (INSERT ... ON DUPLICATE KEY UPDATE - see
+ * iznik-server-go/chat/chatmessage.go CreateChatMessage), closing the
+ * select-then-insert race the earlier attempt left open. Nullable: older/cached
+ * clients that don't send a key are unaffected (MySQL treats each NULL in a
+ * unique index as distinct, so they never collide with each other or with keyed
+ * rows).
+ *
+ * chat_messages is large/high-traffic, but a nullable ADD COLUMN with no computed
+ * default and a plain ADD INDEX both default to ALGORITHM=INPLACE, LOCK=NONE on
+ * Percona/InnoDB 8.0 (unlike the ENUM-widen migrations on this table, which force
+ * INPLACE explicitly because MODIFY COLUMN can otherwise pick COPY) - see
+ * 2026_06_25_000001_add_session_to_browse_scroll_depth for the same shape.
+ */
+return new class extends Migration
+{
+    public function up(): void
+    {
+        if (!Schema::hasTable('chat_messages')) {
+            return;
+        }
+
+        if (!Schema::hasColumn('chat_messages', 'idempotency_key')) {
+            Schema::table('chat_messages', function (Blueprint $table) {
+                $table->string('idempotency_key', 64)->nullable()->after('userid');
+                $table->unique(['chatid', 'userid', 'idempotency_key'], 'chat_messages_idempotency_key_unique');
+            });
+        }
+    }
+
+    public function down(): void
+    {
+        if (Schema::hasColumn('chat_messages', 'idempotency_key')) {
+            Schema::table('chat_messages', function (Blueprint $table) {
+                $table->dropUnique('chat_messages_idempotency_key_unique');
+                $table->dropColumn('idempotency_key');
+            });
+        }
+    }
+};
--- a/iznik-batch/database/migrations/2026_07_14_000001_add_idempotency_key_to_chat_messages_migration.sql
+++ b/iznik-batch/database/migrations/2026_07_14_000001_add_idempotency_key_to_chat_messages_migration.sql
@@ -1,0 +1,19 @@
+-- Production SQL: add a client-generated idempotency key to chat_messages
+-- (Discourse #9913). A retried/duplicated send of the same logical message
+-- carries the same key; the unique index makes at-most-once DB-enforced via
+-- INSERT ... ON DUPLICATE KEY UPDATE in iznik-server-go/chat/chatmessage.go.
+-- Nullable, so older/cached clients that don't send a key are unaffected.
+-- MySQL has no ADD COLUMN IF NOT EXISTS, so guard via information_schema.
+
+SET @col := (
+    SELECT COUNT(*) FROM information_schema.COLUMNS
+    WHERE TABLE_SCHEMA = DATABASE()
+      AND TABLE_NAME = 'chat_messages'
+      AND COLUMN_NAME = 'idempotency_key'
+);
+SET @ddl := IF(@col = 0,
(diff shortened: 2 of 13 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #1069) — this fix did not pass and is not ready to merge.

- **Incomplete diff / load-bearing change missing**: Every artefact in the visible diff (migration comment, SQL comment, stores/chat.js comment, ChatFooter comment) asserts that the server dedupes via 'INSERT ... ON DUPLICATE KEY UPDATE - see iznik-server-go/chat/chatmessage.go CreateChatMessage', but the diff never touches iznik-server-go. If that ha
- **Regression - scope far exceeds the bug**: useFetchRetry.js now disables retry for ALL non-GET/HEAD/OPTIONS requests app-wide, not just chat send. Two concrete regressions: (1) `await miscStore.waitForOnline()` is now skipped for mutating requests, so a POST issued while momentarily offline (mobile, lift, tunnel) rejects immediately instead 
- **Test proves nothing**: tests/unit/components/ChatFooterSendRetry.spec.js does not exercise any behaviour: it `readFileSync`s ChatFooter.vue and regex-matches the source text. It asserts that the string `getSendIdempotencyKey(pendingSend.value, msg)` appears in the file and that a particular error-message literal is presen
- **Unvalidated user input reaching a write path**: `idempotencykey` is now a client-supplied string written straight into a VARCHAR(64) column that participates in a UNIQUE index, and no validation appears anywhere in the diff. Two failure modes: (1) a key longer than 64 chars either errors under strict mode (500 on every send from such a client) or

PR #1072, opened 2026-07-14
fix(modtools): don't show cross-group held-by wording on an already-approved copy

edwh:
## Root Cause
ModTools read the legacy, cross-group `messages.heldby` field to decide whether the message being viewed was held on the group being moderated - instead of the per-group `messages_groups.heldby` row for that exact group. The legacy field is set whenever **any** group holds **any** copy of a rippled post. So a mod who approved a post on their own group, which then rippled to another group and was held there, saw their own (Approved, unheld) copy display "Held by X. Please check with them before releasing it" - even though their copy was never held. The same unscoped field also drove the Hold/Release button toggle and the confirm-before-acting check, so in principle a mod could be offered Release for a copy never held on their own group, or have the confirm dialog silently skipped when a row couldn't be identified.

This is round 4. Rounds 2 and 3 (PR #1065, PR #1068) were auto-closed by adversarial review. This round fixes every blocker raised on #1068, listed below with citations.

## Evidence
AssertFlip - reverted the four production files (composable + 3 components) via `git stash push` (kept only the new/updated tests), ran the affected specs, then restored and re-ran:
```
Test Files  4 failed | 1 passed (5)
     Tests  26 failed | 251 passed (277)

 FAIL  tests/unit/composables/rippleStatus.spec.js > normaliseHeldById (5 tests)
 FAIL  tests/unit/composables/rippleStatus.spec.js > messageGroupRow (Discourse 9904) (5 tests)
 FAIL  tests/unit/composables/rippleStatus.spec.js > messageGroupHeldById (Discourse 9904) (6 tests)
 FAIL  tests/unit/components/modtools/ModMessage.spec.js > Held message > does not show a held banner when this group is unheld even though the legacy cross-group message.heldby is set by a rippled-to group holding its own copy (Discourse 9904)
 FAIL  tests/unit/components/modtools/ModMessage.spec.js > Held message > fails safe with generic wording (no named mod) when this group cannot be identified at all but the legacy field says it is held somewhere
 FAIL  tests/unit/components/modtools/ModMessageButton.spec.js > confirmButton computed (6 tests, incl. "fails SAFE (confirms) when this group cannot be identified at all - the undefined branch")
 FAIL  tests/unit/components/modtools/ModMessageButtons.spec.js > offers Hold (not Release) when a DIFFERENT group holds its own copy - the exact Discourse 9904 scenario
 FAIL  tests/unit/components/modtools/ModMessageButtons.spec.js > does not fall back to the legacy field when this group's own row is unheld but no other group holds it either
```
All 26 fail on the pre-fix (reverted) code; all pass once the fix is restored. Full local run after restoring: `npx vitest run tests/unit/` -> `Test Files 669 passed (669)` / `Tests 14301 passed | 4 skipped (14305)` - no regressions (this includes 2 new tests already on master from an unrelated commit merged after this branch's base).

Go: `go build ./...` and `go vet ./message/... ./test/...` are clean; `go test ./test/... -run TestGetMessageGroupsHeldbyIsPerGroupScoped -c` compiles the new test successfully. I could not execute it against a live MySQL instance in this sandboxed worktree session - it has no bound Docker network route to the shared test DB from WSL (`iznik-server-go/CLAUDE.md`: "You can't connect to the database from WSL"). CI will run the full Go suite on this PR before merge.

## API Contract
`groups[].heldby` is **already emitted** by the existing, unmodified Go serialiser - no backend change is needed, only citations:
- `iznik-server-go/message/messageGroup.go:27` - `Heldby *uint64 \`json:"heldby,omitempty"\`` on `MessageGroup` (table `messages_groups`).
- `iznik-server-go/message/message.go:487` - the per-group SQL select explicitly selects `heldby`: `SELECT groupid, msgid, arrival, collection, autoreposts, approvedby, heldby, spamtype, ... FROM messages_groups WHERE msgid = ? AND deleted = 0`, scanned into `messageGroups`.
- `iznik-server-go/message/message.go:212` - `MessageGroups []MessageGroup \`gorm:"-" json:"groups"\`` on `Message` - so the JSON `groups` array is exactly those per-group rows, `heldby` included.
- Call path: `GetMessages` (`GET /api/message/:ids`) -> `GetMessagesByIds` runs the select above. `messageStore.fetch`/`fetchMT` (what `ModMessage.vue`/`ModMessageButtons.vue` use) call this exact endpoint via `api/MessageAPI.js`.

To make this an executable proof rather than just a citation, I added `iznik-server-go/test/message_test.go::TestGetMessageGroupsHeldbyIsPerGroupScoped`, which hits `GET /api/message/:id` for a message that is Approved+unheld on group A and Pending+held (by a different user) on group B, and asserts the JSON `groups[]` shows A's row as unheld and B's row as held by that user - reproducing the reporter's exact scenario end-to-end against the real handler (compiled, not run locally - see Evidence).

## Fix
- **`composables/rippleStatus.js`**: `messageGroupRow(message, groupid)` finds the matching `messages_groups` row for the exact group, returning the row, `null` (groups is a loaded array with no matching row - conclusively not held), or `undefined` (groups isn't an array yet / groupid unresolvable - genuinely unknown). `messageGroupHeldById` derives a plain numeric holder id from that via the new `normaliseHeldById` (handles numeric, numeric-string, and the legacy `{id, displayname}` object shape). This distinguishes a loaded-but-empty `groups: []` from a message still loading, so callers only fail safe when truly unresolvable.
- **`ModMessage.vue`**: the held banner, the footer's button-hiding gate, and `heldbyId`/`heldbyName` all derive from `messageGroupRow`/`messageGroupHeldById` scoped to `currentGroupid`, never `message.heldby`. `heldbyName` reads `displayname` straight off the matched row's raw `heldby` object when present, falling back to a `userStore` lookup only for numeric ids (previously discarded the embedded displayname). A narrow, explicit fallback (`heldByUnknownGroup`) covers the case where this group's row genuinely can't be identified (message/groups still loading): it shows generic "Held on another group" wording, naming no one - **except** when the legacy field's holder matches the current user, in which case they see "You held this" and keep the Release button, so the actual holder is never locked out even in this fallback (`legacyHeldbyId`, `effectiveHeldbyId`, `heldAndNotByMe`). The holder-name fetch is now a `watch(heldbyId, ...)` (not a one-shot `onMounted`), so it re-fetches whenever the held id changes - e.g. after a `hold()`/`release()` refetch.
- **`ModMessageButtons.vue`**: the Hold/Release toggle (`heldByThisGroup`) uses `messageGroupHeldById` scoped to `props.groupid`, rendered via the existing `v-if`/`v-else` pair so Hold and Release stay structurally mutually exclusive.
- **`ModMessageButton.vue`**: `confirmButton` also uses `messageGroupHeldById`, and now fails **safe**: `held === undefined || typeof held === 'number'` - so an unidentifiable row (message/groups still loading) triggers the confirm dialog instead of silently skipping it. Spam/delete (which have their own confirm flow) short-circuit to `false` first.
- No backend change: the field was already correctly emitted per-group; the bug was purely which field the frontend read. The legacy `messages.heldby` top-level field is left serialised for backward compatibility with any other consumer - not read for hold-state decisions anywhere in this diff.

## Reviewer Blockers Addressed
1. **Unverified API contract** - see "API Contract" above: cited the exact Go struct/select/route, plus a new Go test (`TestGetMessageGroupsHeldbyIsPerGroupScoped`) that exercises the real handler end-to-end.
2. **Holder locked out of their own held message** - the unknown-group fallback now compares `normaliseHeldById(message.heldby)` against `myid` (`legacyHeldbyId` vs `effectiveHeldbyId` in `heldAndNotByMe`), covered by "keeps the holder able to Release even via the unknown-group fail-safe fallback, when the legacy holder is me".
3. **Confirm guard fails open** - `ModMessageButton.vue`'s `confirmButton` now confirms on `undefined` as well as a numeric holder id, covered by "fails SAFE (confirms) when this group cannot be identified at all - the undefined branch".
4. **displayname discarded** - `heldbyName` reads the raw matched row's `heldby.displayname` directly when it's object-shaped, only falling back to a `userStore` lookup for numeric ids; watched (not one-shot) so the name renders once the user is fetched. Asserted by "shows release button when held, warning when held by someone else" (checks the rendered name text).
5. **Other `.heldby` call sites unaudited** - see "Other Call Sites" below; hold/release audited in stores/message.js.
6. **Root cause / groups:[] vs missing untouched** - `messageGroupRow` now explicitly distinguishes a loaded-but-empty `groups: []` (conclusively not held - falls through to `null`) from a genuinely missing/not-yet-loaded `groups` (returns `undefined`, the only case that triggers the fail-safe fallback). The legacy `messages.heldby` write path itself is unchanged (out of scope for this frontend-read bug, and changing it risks other consumers); noted here for visibility.

## Other Call Sites
Grepped `\.heldby\b` across all of `iznik-nuxt3` (not just `modtools/components`):
- `ModMessage.vue` (banner, footer gate, `heldbyId`/`heldbyName`, holder-name fetch) - fixed, see Fix.
- `ModMessageButtons.vue` (`heldByThisGroup`, gates Hold vs Release) - fixed.
- `ModMessageButton.vue` (`confirmButton`) - fixed.
- `stores/message.js` `hold()`/`release()` - **audited, no change needed**: both already call the API then fully re-fetch the message (`api.message.fetchMT`) and replace `this.list[message.id]` with the fresh copy. Since every consuming component reads the message via the reactive `messageStore.byId` getter, that replacement updates the correct `groups[].heldby` row (and everything derived from it) automatically - no separate optimistic patch is needed. Proved by two new store-level tests asserting `store.byId(id).groups.find(...).heldby` reflects the post-refetch value after `hold()`/`release()`.
- `modtools/stores/member.js`, `modtools/stores/spammer.js`, `ModMember.vue`, `ModMemberButtons.vue`, `ModMemberReviewActions.vue`, `ModAdmin.vue` (`member.heldby`/`membership.heldby`/`admin.heldby`) - unrelated: those rows are inherently per-object with no rippling/multi-group concept, so they don't share this bug. Left untouched.

## Test
- Files: `iznik-nuxt3/tests/unit/composables/rippleStatus.spec.js`, `tests/unit/components/modtools/ModMessage.spec.js`, `tests/unit/components/modtools/ModMessageButtons.spec.js`, `tests/unit/components/modtools/ModMessageButton.spec.js`, `tests/unit/stores/message.spec.js`, `iznik-server-go/test/message_test.go` (`TestGetMessageGroupsHeldbyIsPerGroupScoped`).
- Command: `npx vitest run tests/unit/` (whole unit suite) - `669 passed (669)` files, `14301 passed | 4 skipped` tests.
- Proves fail-before/pass-after via AssertFlip (see Evidence above): reverted just the 4 production files with `git stash`, confirmed 26 new/updated tests fail, restored, confirmed all pass with no other regressions.

## Discourse

### DIFF
--- a/iznik-nuxt3/composables/rippleStatus.js
+++ b/iznik-nuxt3/composables/rippleStatus.js
@@ -148,3 +148,71 @@ export function isHomeGroup(group, groups) {
   const gid = group.groupid ?? group.id
   return gid != null && parseInt(gid) === homeId
 }
+
+/**
+ * Coerce a heldby value - numeric, numeric-string, or object ({id, displayname}, the
+ * legacy PHP-API shape) - to a plain numeric user id. Returns null for anything that
+ * isn't held (null/undefined/unparseable).
+ *
+ * @param {number|string|{id:number}|null|undefined} h
+ * @returns {number|null}
+ */
+export function normaliseHeldById(h) {
+  if (h === null || h === undefined) return null
+  if (typeof h === 'object') return h.id ?? null
+  const n = parseInt(h)
+  return Number.isNaN(n) ? null : n
+}
+
+/**
+ * The messages_groups row for the given group on this message - i.e. the group's own
+ * copy, as opposed to the legacy cross-group messages.heldby column, which is set
+ * whenever ANY group holds ANY copy of a rippled post. Reading that legacy field leaked
+ * a hold placed on one group's copy into every other group's view of the same message: a
+ * mod saw their own already-approved copy display "Held by X, check with them before
+ * releasing it" even though their copy was never held (Discourse 9904).
+ *
+ * Distinguishes "loaded, but no row" from "not loaded yet", so callers can tell a
+ * genuinely-resolved absence apart from a still-loading message:
+ *  - the matching row object if found
+ *  - null if `message.groups` is a loaded array but has no row for this group
+ *  - undefined if `message.groups` isn't an array yet (not loaded), or `groupid` can't
+ *    be resolved to a number at all
+ *
+ * @param {{groups?: Array<{groupid:number|string, heldby?:number|string|{id:number}|null}>}} message
+ * @param {number|string} groupid
+ * @returns {object|null|undefined}
+ */
+export function messageGroupRow(message, groupid) {
+  const groups = message?.groups
+  if (!Array.isArray(groups)) return undefined
+  const gid = parseInt(groupid)
+  if (Number.isNaN(gid)) return undefined
+  return groups.find((g) => parseInt(g.groupid) === gid) ?? null
+}
+
+/**
+ * The id of the moderator holding THIS message on THIS group - see messageGroupRow for
+ * the loaded/not-loaded distinction. This is the ONLY place that should turn a group row
+ * into a held-by id; every consumer (the held-by banner, the Hold/Release button toggle,
+ * the confirm-before-acting check) must call this rather than inspecting
+ * message.heldby or a group row directly, so they can't drift out of sync with each
+ * other.
+ *
+ * Returns:
+ *  - a numeric user id if the group's own copy is held
+ *  - null if the row is loaded and definitely NOT held (including a loaded-but-empty
+ *    groups array, which conclusively has no row for this group)
+ *  - undefined if the row can't be identified at all (message still loading) - callers
+ *    must not guess who holds it in this case, since that mod may be on a group the
+ *    reader can't act on.
+ *
+ * @param {object} message
+ * @param {number|string} groupid
+ * @returns {number|null|undefined}
+ */
+export function messageGroupHeldById(message, groupid) {
+  const row = messageGroupRow(message, groupid)
+  if (row === undefined || row === null) return row
+  return normaliseHeldById(row.heldby)
+}
--- a/iznik-nuxt3/modtools/components/ModMessage.vue
+++ b/iznik-nuxt3/modtools/components/ModMessage.vue
@@ -323,16 +323,19 @@
                 at
                 {{ datetimeshort(message.outcomes[0].timestamp) }}
               </NoticeMessage>
-              <div v-if="message.heldby">
+              <div v-if="heldNoticeVisible">
                 <NoticeMessage variant="warning" class="mb-2">
-                  <p v-if="me.id === heldbyId">
+                  <p v-if="me.id === effectiveHeldbyId">
                     You held this. Other people will see a warning to check with
                     you before releasing it. If you release it, it will stay in
                     Pending.
                   </p>
-                  <p v-else>
+                  <p v-else-if="heldByThisGroup">
                     Held by <strong>{{ heldbyName }}</strong
                     >. Please check with them before releasing it.
+                  </p>
+                  <p v-else>
+                    Held on another group. Please check before releasing it.
                   </p>
                   <ModMessageButton
                     :messageid="message.id"
@@ -721,7 +724,7 @@
         </b-row>
       </b-card-body>
       <b-card-footer v-if="!noactions && expanded">
-        <div v-if="message.heldby && heldbyId !== myid">
+        <div v-if="heldAndNotByMe">
           This message is held by someone else. The buttons are hidden so you
           don't click them by accident. Please check with them before releasing
           the message.
@@ -734,11 +737,7 @@
           This message needs editing so that we know where it is.
         </NoticeMessage>
         <div
-          v-if="
-            pending &&
-            (!message.heldby || (message.heldby && heldbyId === myid)) &&
-            !editing
-          "
+          v-if="pending && !heldAndNotByMe && !editing"
           class="text-end mb-1"
         >
           <b-button variant="danger" @click="spamReport">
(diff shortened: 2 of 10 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #1072) — this fix did not pass and is not ready to merge.

- **Duplicate implementation**: ModMessage.vue reimplements the helper this PR just added. `messageGroupHeldById()` in rippleStatus.js is exactly `messageGroupRow()` + tri-state passthrough + `normaliseHeldById(row.heldby)`, and its own docblock says: 'This is the ONLY place that should turn a group row into a held-by id; every co
- **Known regression / inconsistent fail-safe**: ModMessageButtons.vue fails OPEN on the unresolved-row case while the other two consumers fail SAFE, producing a self-contradictory UI in the very scenario this PR added a test for. `heldByThisGroup = typeof messageGroupHeldById(message.value, props.groupid) === 'number'` collapses the helper's `und
- **Partial implementation / root cause not addressed**: The fix is display-only and never touches the write path or the API that produces the bad data. The diff's own comments state the root cause: 'the legacy cross-group messages.heldby column ... is set whenever ANY group holds ANY copy of a rippled post'. That column is still written by hold(), still 

PR #1073, opened 2026-07-14
fix(spatial): skip recent-message upsert when the parent was deleted mid-run

edwh:
## The exception

```
PDOException 1452: Cannot add or update a child row: a foreign key constraint fails
(`iznik`.`messages_spatial`, CONSTRAINT `messages_spatial_ibfk_1`
 FOREIGN KEY (`msgid`) REFERENCES `messages` (`id`))
INSERT INTO messages_spatial (msgid, ...) VALUES (119128577, ...)
  at MessageSpatialService::upsertRecentMessages (line 87)
```

## Root cause (investigated, not assumed)

I initially guessed a same-node purge race; the evidence says something more specific.

**Audit trail for msgid 119128577** (`logs` table): rippled to 7 groups → origin removed **07-01** (`messages_groups.deleted=1` on the rippled copies, message row untouched) → autoreposted on its origin group **07-08** (arrival set to the value in the failing insert) → hard-deleted **07-14**.

**Timing**: the FK error fired **once**, at `2026-07-14 02:30:31`. `purge:messages` runs `dailyAt('02:30')`. The reconciler runs every 5 minutes; the one run that errored is the one that overlapped the purge.

**Mechanism** (confirmed against live config, not inferred):
- batch-prod splits reads → db2 (`DB_HOST_READ`=10.220.0.150), writes → db3 (`DB_HOST`=10.220.0.47); `config/database.php` has `'sticky' => false`.
- Both nodes report `wsrep_sync_wait = 0`.
- Laravel routes `->get()` (the candidate SELECT) to the read PDO (db2) and `DB::statement()` (the INSERT) to the write PDO (db3).

Galera certifies writes synchronously but **applies** them to other nodes asynchronously. With sync-wait off and no stickiness, the reconciler's SELECT on db2 read message 119128577 as still-present (its delete was committed on db3 but not yet applied on db2), then the INSERT on db3 failed the FK because the parent was already gone there. `config/database.php:62-64` explicitly assumes "a committed write is visible on the read nodes without meaningful delay" — true for read-after-*write*-of-your-own-row-in-a-txn, **false for read-after-delete across the split**.

The message's deletion itself was normal purge behaviour — nothing corrupt about it.

## Impact

The exception propagated out of the per-row loop, so a **single** vanished message aborted the entire run: every later `messages_spatial` upsert in that batch was dropped, and all four cleanup passes in `updateSpatialIndex()` (outcomes, removed_deleted, removed_old, removed_non_approved) were skipped. Self-clearing the next run (the SELECT no longer returns a purged message), but it recurs whenever the 02:30 purge overlaps a reconciler pass.

## Fix

Catch errno **1452** per row and skip — a candidate that passed the SELECT's `whereNull('messages.deleted')` filter but fails the FK can only mean the parent was deleted after the snapshot, which is always a benign "don't index it" case. Anything other than 1452 is re-thrown so real errors still surface. A count of skips is logged.

Regression test reproduces the snapshot→insert deletion deterministically with a `DB::listen` hook (fires after the candidate SELECT, before the inserts) and asserts (a) no exception and (b) a valid sibling message is still indexed.

## Deliberately not done here

- **Not** setting `wsrep_sync_wait=1` — that forces causal reads cluster-wide and would add latency to every read across the site (db3 is already CPU-bound). Too broad for this.
- **Not** forcing this reconciler's SELECT onto the write PDO — it would pile a heavy candidate scan onto the hottest node and only fixes this one job. The per-row skip is the correct resilience posture for an eventually-consistent reconciler and defends against any cause of the parent vanishing.

### DIFF
--- a/iznik-batch/app/Services/MessageSpatialService.php
+++ b/iznik-batch/app/Services/MessageSpatialService.php
@@ -4,6 +4,7 @@ namespace App\Services;
 
 use App\Models\Message;
 use App\Models\MessageGroup;
+use Illuminate\Database\QueryException;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Log;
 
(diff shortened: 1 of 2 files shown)


edwh:
Closing in favour of #1074, which fixes the root cause: autorepost was resurrecting rippling-removed memberships, which is what fed this reconciler the row that then raced the purge. With that fixed the reconciler is no longer handed resurrected rows, and a per-row catch that swallows FK 1452 would mask rather than fix. If the reconciler-vs-purge race resurfaces from an unrelated delete path, we can revisit a targeted guard then.

PR #1080, opened 2026-07-15
fix(chat): stop a stale message-fetch race from clobbering a just-sent message

edwh:
## Root Cause
`stores/chat.js` `send()` fires an **un-awaited** `this.fetchMessages(chatid)` as a fire-and-forget refresh after posting. Every current caller of `send()` (`ChatFooter.vue`'s `_updateAfterSend()`, `ChatMessageAddress.vue`, `ModChatFooter.vue`, etc.) then *also* makes its own explicit `fetchMessages()` call right afterwards. That means every single chat send dispatches **two unsynchronised, concurrent** `GET /chat/:id/message` requests for the same chat, with no guarantee they resolve in the order they were sent (read-replica lag, network jitter, etc. can easily reorder them — see the existing DB read/write-split issues in this codebase).

`fetchMessages()`'s update logic only compared array *lengths* to decide whether to apply a response:

```js
if (this.messages[id]?.length !== messages.length) {
  update()
}
```

There was no check for whether the response was still the *latest* outstanding request. So if the earlier-dispatched request happened to resolve *after* the later one (e.g. it hit a lagging read replica while the second request hit a fresher one), its stale, pre-send snapshot would silently overwrite the store — the just-sent message would flash into view and then vanish again, only reappearing once a later poll caught up. This is the race behind the transient duplicate/inconsistent rendering reported in direct member Chat.

I confirmed the mechanism directly in this code path rather than assuming the prior PR's ChitChat-style diagnosis applied: direct Chat's `ChatFooter.vue` binds only `@keyup.enter.exact="sendOnEnter"` (single handler, guarded synchronously by `sending.value`), so the ChitChat keydown+keyup double-submit bug does not exist here. There is also no optimistic local message append anywhere in `stores/chat.js` — every update is a full-array replace from the API response. The actual race is the redundant unguarded double-fetch above.

## Evidence
Failing test before the fix (`tests/unit/stores/chat.spec.js`):
```
❯ tests/unit/stores/chat.spec.js:397:34
AssertionError: expected [ { id: 1, message: 'hi' } ] to have a length of 2 but got 1
    395|       // The just-sent message must still be visible — a late, stale r…
    396|       // not be allowed to overwrite the fresher state.
    397|       expect(store.messages[10]).toHaveLength(2)
       |                                  ^
    398|       expect(store.messages[10].map((m) => m.id)).toEqual([1, 2])

 Test Files  1 failed (1)
      Tests  1 failed | 42 passed (43)
```
This reproduces the bug directly: a slow, stale `fetchMessages` response (dispatched first, resolves last) wipes out a freshly-rendered just-sent message.

## Fix
Added a per-chat monotonic sequence counter (`messagesFetchSeq`) in the chat store. Each `fetchMessages(id)` call stamps itself with the current sequence number for that chat on entry; when its response arrives, it only applies the update if it is still the most recently-dispatched call for that chat id. A response from a superseded (older) request is dropped instead of overwriting fresher state — regardless of network arrival order. This fixes the race at its root (in the shared store method), rather than papering over it in one caller, so it also covers ModTools' `ModChatPane`/`ModChatFooter`, which share the same `chatStore.fetchMessages`.

## Other Call Sites
Grepped all callers of `chatStore.send(` and `chatStore.fetchMessages(` — every one goes through the same shared store method, so this fix covers direct member chat (`ChatFooter.vue`), address sending (`ChatMessageAddress.vue`), message reporting (`MessageReportModal.vue`), and the ModTools chat panes uniformly. No separate optimistic-append pattern exists elsewhere in the chat code (verified via grep for `messages[...].push/unshift` — none found; all updates are full-array replaces).

## Test
- File: `iznik-nuxt3/tests/unit/stores/chat.spec.js`
- Command: `npx vitest run tests/unit/stores/chat.spec.js`
- Fail before / pass after: the new test `fetchMessages > keeps the freshest data when an overlapping fetch resolves late (9913)` fails on unfixed code (`expected length 2 but got 1`) and passes once the sequence-guard fix is applied.
- Full unit suite: `npx vitest run` — 669 files, 14289 tests passed, 4 pre-existing skips, no regressions.

## Discourse

### DIFF
--- a/iznik-nuxt3/stores/chat.js
+++ b/iznik-nuxt3/stores/chat.js
@@ -14,6 +14,7 @@ export const useChatStore = defineStore({
     listByChatId: {},
     listByChatMessageId: {},
     messages: {},
+    messagesFetchSeq: {},
     searchSince: null,
     showContactDetailsAskModal: false,
     showClosed: false,
@@ -27,6 +28,7 @@ export const useChatStore = defineStore({
       this.listByChatId = {}
       this.listByChatMessageId = {}
       this.messages = {}
+      this.messagesFetchSeq = {}
       this.searchSince = null
       this.showContactDetailsAskModal = false
       this.currentChatMT = null
@@ -258,10 +260,26 @@ export const useChatStore = defineStore({
       }
     },
     async fetchMessages(id, force) {
+      // Callers routinely fire off more than one fetchMessages(id) for the same chat in
+      // quick succession and don't wait for each other (e.g. send() triggers its own
+      // refresh and the caller of send() triggers another straight after). Network
+      // timing gives no guarantee the responses land in the order they were requested,
+      // so track which call is the most recent for this chat and ignore a response if a
+      // newer request has since been made - an older, slower response must never
+      // clobber fresher data (#9913).
+      const seq = (this.messagesFetchSeq[id] || 0) + 1
+      this.messagesFetchSeq[id] = seq
+
       let messages = []
       const miscStore = useMiscStore() // MT
       const params = miscStore.modtools ? { modtools: true } : {}
       messages = await api(this.config).chat.fetchMessages(id, params)
+
+      if (this.messagesFetchSeq[id] !== seq) {
+        // A newer fetchMessages(id) call has been made since this one started - this
+        // response is stale, so drop it rather than overwrite more current data.
+        return messages
+      }
 
       const update = () => {
         this.messages[id] = messages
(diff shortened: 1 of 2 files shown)


edwh:
Superseded and closing.

9913 (a just-sent chat message briefly showing twice) is fixed on master in 7f2f59f1e via the correct diagnosis: `useFetchRetry` retries a POST on a network error / 5xx, so a retried send inserts a genuine SECOND `chat_messages` row (different id, identical content); the `cleanup:chat-duplicates` cron later removes one - which matches the reporter's "the repeated message has now disappeared". The fix dedupes retry-duplicates on read in `chatStore.fetchMessages`.

That is a different mechanism from the stale-fetch-reorder race this PR targeted (a full-array replace of unique-id server data can't produce a duplicate; the reorder would make a message vanish/reappear, not show twice). This PR also now conflicts with the master change.

PR #1084, opened 2026-07-16
refactor(notified): generalise reach ledger to messages_notified with channel

edwh:
## What

Stage 1 of reducing repeat views in the digest (design:
`plans/2026-07-16-digest-reduce-repeat-views.md`). **Behaviour-neutral** groundwork:
generalise the per-(member, post) "we notified this member about this post" ledger so
the upcoming digest work can record its own sends in the *same* table instead of a
parallel one.

`rippling_reach_notified (msgid, userid, notified_at)` -> `messages_notified
(msgid, userid, channel, notified_at)`, PK `(msgid, userid, channel)`. Existing rows
become `channel='reach'`; nothing writes another channel yet.

## Why the `channel` guard matters

Several consumers treat "a row exists for (msgid, userid)" as "the **reach** mail was
sent" — reply attribution keys off `notified_at <= replied_at`. The moment another
channel shares the table, those checks must filter `channel='reach'` or a digest send
would be miscounted as reach. This PR adds that filter everywhere first, while all
rows are still `reach`, so it's provably a no-op now and correct later.

## Changes

- **Migration** `2026_07_16_000001_generalise_notified_ledger`: add `channel`
  (default `'reach'`), widen PK in a single `ALTER` (msgid stays leftmost so the
  msgid FK index is never dropped), `RENAME TABLE`. `down()` drops non-reach rows
  before narrowing the PK. + idempotent production `*_migration.sql`.
- **`channel='reach'`** on every reach read/write: `UnifiedDigestService`,
  `ReplyAttributionBackfillService`, Go `rippling/metrics.go`, Go
  `chat/chatmessage.go`.
- Table renamed across `test-fixtures.sql`, Go test table-creates, and doc comments.

## Test Plan

- `setup-test-database` runs the rename migration **and** reloads `test-fixtures.sql`
  cleanly (verified locally — no ERROR 1146 from the fixture's `LOCK TABLES`).
- Go suite green (reach/metrics attribution tests exercise the renamed table +
  `channel` filter end-to-end).
- Laravel reach/attribution/digest tests green, including a new assertion in
  `ReplyAttributionBackfillServiceTest`: a member notified only on `channel='digest'`
  is **not** read as reach-notified (`was_notified=0`, attribution != `ripple_notified`).
  (The 20 spatial `JobModel`/`Location`/`Newsfeed` failures seen on the local box are a
  local test-DB data gap — `iznik_batch_test` has no postcode/location rows — and are
  untouched by this branch.)

## Follow-ups (not in this PR)

Stage 2: digest writes `channel='digest'` on spool success. Stage 3: per-member
scorer penalty / immediate suppression using `messages_notified` + `messages_likes`
views. Both behind a config flag.

https://claude.ai/code/session_01Rm8xqAcBsnaYWZkrp5KAMA

### DIFF
--- a/iznik-batch/app/Services/Ripple/ExpandService.php
+++ b/iznik-batch/app/Services/Ripple/ExpandService.php
@@ -107,7 +107,7 @@ class ExpandService
      * post and overwrites the stored polygon + schedule at the post's CURRENT
      * tick. It deliberately does NOT ripple into/out of groups, touch
      * memberships, or bump updated_at — so it generates no mail (sendReachDigests
-     * only scans recently-updated rows, and the rippling_reach_notified ledger
+     * only scans recently-updated rows, and the messages_notified ledger
      * blocks re-notification regardless) and never retracts copies already
      * delivered to far groups (we shrink future reach, we don't claw back).
      *
--- a/iznik-batch/app/Services/Ripple/ReplyAttributionBackfillService.php
+++ b/iznik-batch/app/Services/Ripple/ReplyAttributionBackfillService.php
@@ -10,7 +10,7 @@ use Illuminate\Support\Facades\DB;
  *
  * Durable bits are re-derivable from timestamped tables, so freezing them now - before the
  * evidence decays further - is the whole point:
- *   was_notified            <- rippling_reach_notified (notified_at <= replied_at)
+ *   was_notified            <- messages_notified (notified_at <= replied_at)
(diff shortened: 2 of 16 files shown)


edwh:
Closing in favour of a single PR taking the read/click approach instead of send-tracking: we only treat a post as 'seen' for a member when they opened/clicked a digest that contained it (or viewed it in-app), and that unified signal drives both the digest ordering and the browse-page ordering. The channel/send-ledger groundwork here isn't the right primitive for that.

PR #1104, opened 2026-07-18
fix(email/admin-mail): tracking pixel breaks Outlook rendering (blank on reopen)

edwh:
## Root Cause
`admin.blade.php`'s tracking pixel was appended as a bare `<mj-image>` directly under `<mj-body>` instead of being wrapped in `<mj-section><mj-column>`. This violates MJML's schema (an `mj-image` must live inside `mj-column` > `mj-section`). The in-process `mrml` compiler renders that malformed nesting without the Outlook-conditional (`<!--[if mso | IE]>`) wrapper table that every other section in the email gets, leaving the compiled HTML without the well-formed structure Outlook's Word rendering engine needs — matching the reporter's admin email going blank on reopen in Outlook.

Confirmed by rendering the real Blade view: every properly-nested section produces a `<!--[if mso | IE]><table ... bgcolor=... width="600">...` wrapper plus a `max-width:600px` centred `<div>`; the pixel's `<table>` at the very end of `<body>` has neither.

## Evidence
Failing test before the fix (`tests/Unit/Mail/AdminMailTest.php::test_admin_mail_tracking_pixel_is_wrapped_in_section_and_column`), reverted temporarily to prove it catches the bug:

```
1) Tests\Unit\Mail\AdminMailTest::test_admin_mail_tracking_pixel_is_wrapped_in_section_and_column
Failed asserting that '...<mj-image src="https://example.com/pixel.png" .../>\n            </mj-body>\n</mjml>\n' matches PCRE pattern
"#<mj-section[^>]*>\s*<mj-column>\s*<mj\-image src\="https\://example\.com/pixel\.png" .../>\s*</mj-column>\s*</mj-section>#".

  at tests/Unit/Mail/AdminMailTest.php:287
  Tests: 1 failed (1 assertions)
```

## Fix
Wrapped the tracking pixel include in both affected templates in `<mj-section padding="0"><mj-column>...</mj-column></mj-section>` — the exact same pattern already used correctly by 13 other mailables (welcome, admin/chase, digest siblings, donation, session).

## Other Call Sites
Grepped every `{!! $trackingPixelMjml !!}` usage. Two templates had the bug:
- `resources/views/emails/mjml/admin/admin.blade.php` (this report — topic 9925)
- `resources/views/emails/mjml/digest/unified.blade.php` (Daily Unified Digest — same literal defect, fixed in this PR since it's the identical root cause)

13 other templates (`welcome/welcome.blade.php`, `admin/chase.blade.php`, `message/{chaseup,autorepost-warning,deadline-reached,chaseup-promised,mod-std-message}.blade.php`, `donation/{giftaid-chaseup,thank-you,ask}.blade.php`, `session/{merge-offer,forgot-password,verify-email}.blade.php`) already wrap it correctly — no changes needed there.

## Test
- File: `iznik-batch/tests/Unit/Mail/AdminMailTest.php::test_admin_mail_tracking_pixel_is_wrapped_in_section_and_column`
- Command: `php artisan test --filter=test_admin_mail_tracking_pixel_is_wrapped_in_section_and_column`
- Proves fail-before/pass-after: confirmed by reverting the blade fix, re-running (fails, output above), restoring the fix, re-running (passes).
- Regression check: ran `AdminMailTest`, `SendAdminCommandTest`, `UnifiedDigestTest`, `UnifiedDigestSummaryTest`, `UnifiedDigestRetryTest`, `EmailDeliveryIntegrationTest` together — 107 passed, 0 failed (6 integration tests gracefully skipped, no Mailpit in this isolated test environment).

## Discourse
https://discourse.ilovefreegle.org/t/9925/1

### DIFF
--- a/iznik-batch/resources/views/emails/mjml/admin/admin.blade.php
+++ b/iznik-batch/resources/views/emails/mjml/admin/admin.blade.php
@@ -82,7 +82,11 @@
         @include('emails.mjml.partials.footer', ['email' => $user->email_preferred, 'settingsUrl' => $settingsUrl])
 
         @if(isset($trackingPixelMjml))
-        {!! $trackingPixelMjml !!}
+        <mj-section padding="0">
+            <mj-column>
+                {!! $trackingPixelMjml !!}
+            </mj-column>
+        </mj-section>
         @endif
     </mj-body>
 </mjml>
--- a/iznik-batch/resources/views/emails/mjml/digest/unified.blade.php
+++ b/iznik-batch/resources/views/emails/mjml/digest/unified.blade.php
@@ -426,7 +426,11 @@
         @include('emails.mjml.partials.footer', ['email' => $user->email_preferred, 'settingsUrl' => $settingsUrl])
 
         @if(isset($trackingPixelMjml))
-        {!! $trackingPixelMjml !!}
+        <mj-section padding="0">
+            <mj-column>
+                {!! $trackingPixelMjml !!}
+            </mj-column>
+        </mj-section>
         @endif
 
         @endif
--- a/iznik-batch/tests/Unit/Mail/AdminMailTest.php
+++ b/iznik-batch/tests/Unit/Mail/AdminMailTest.php
@@ -254,4 +254,40 @@ class AdminMailTest extends TestCase
         $this->assertStringContainsString('3 days', $html,
             'Preheader must state how long the admin has been pending');
     }
+
+    public function test_admin_mail_tracking_pixel_is_wrapped_in_section_and_column(): void
+    {
+        // A bare <mj-image> as a direct child of <mj-body> violates the MJML
+        // schema (mj-image must live inside mj-column > mj-section). mrml
+        // renders that malformed nesting without the Outlook-conditional
+        // wrapper table every other section gets, which is what left the
+        // reported admin email blank on reopen in Outlook (topic 9925). Every
+        // other mailable (welcome, chase, digest siblings, etc.) wraps its
+        // tracking pixel in its own <mj-section padding="0"><mj-column> - this
+        // pins admin.blade.php to the same pattern.
+        $pixel = '<mj-image src="https://example.com/pixel.png" width="1px" height="1px" alt="" padding="0" />';
+
+        $html = view('emails.mjml.admin.admin', [
+            'user' => (object) ['email_preferred' => 'test@example.com', 'displayname' => 'Test User'],
+            'userSite' => config('freegle.sites.user'),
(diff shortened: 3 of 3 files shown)


edwh:
This misdiagnoses 9925 and I'd recommend closing it. The real cause is already fixed.

**What actually happened (confirmed against the live production DB via the V2 tunnel):**

The admin email Emma received was Freegle-wide alert **11194** (2026-07-13 12:27:23, `groupid NULL`, to Mods). Its stored body:

| col | length | content |
|---|---|---|
| `html` | **11** | `<p><br></p>` (Quill's empty-editor sentinel) |
| `text` | 750 | the real message |

The mailer renders the HTML column, so the body arrived empty — the 750-char message was stranded in the `text` column. That is the empty-editor-sentinel bug, **already fixed** by `5ed574101` (merge `7d5cafd6c`, `fix/alert-empty-html-editor-sentinel`), which replaced `req.Html == ""` with `htmlIsBlank()` in `alert.go CreateAlert`. Alert 11194 went out ~30 minutes before that fix landed; it was one of the ~288 empty mod emails from that incident. Emma is a mod who received the all-groups alert.

**Why the tracking-pixel / Outlook theory doesn't hold:**
- The evidence for 9925 is `html_len = 11` (`<p><br></p>`) — the email was content-empty *at send time*, not blanked by Outlook's renderer on reopen.
- Every alert sent after the fix (11197, 11200, …) carries the **same** tracking pixel and a full HTML body, with no blank-body reports. If the pixel nesting broke Outlook on reopen, those would blank too.
- "Blank on reopen" is the reporter's framing of noticing the empty body later, not two distinct render states.

**Concern with the change itself:** beyond fixing the wrong cause for 9925, this also rewrites the tracking-pixel nesting in `unified.blade.php` — the live Daily Unified Digest — on that same wrong basis. That's an unnecessary structural change to a high-volume live template.

If we want the `mj-image`-inside-`mj-column` wrapping purely as MJML-schema hygiene, that can stand on its own merits in a separate change with a clear "matches the other 13 templates" rationale — but it should not be attributed to 9925, which is already resolved.

edwh:
Closing per the investigation above. 9925's blank admin email was the empty-HTML-editor-sentinel bug (alert 11194, html=`<p><br></p>`, real text stranded in the text column), already fixed by 5ed574101 (merged, live). This PR targets a different, unconfirmed Outlook tracking-pixel theory and edits the live unified digest template on that basis. If we want the mj-image-in-mj-column wrapping purely as MJML hygiene, it can come back as a standalone change not attributed to 9925.

PR #1110, opened 2026-07-18
fix(apiv2): collision-safe chat room consolidation on email-verify account merge

edwh:
## Summary

- Fixes chat history loss during email-verify account merges. When a user verifies an email owned by another account, `PatchSession` merges that account into theirs. The chat_rooms reassignment used two bare `UPDATE`s, and `chat_rooms` has a unique key on `(user1, user2, chattype)` - so when both accounts already had a room with the same counterparty, the `UPDATE` failed with a duplicate-key error, the rooms stayed on the merged account, and the merge then marked that account deleted. Those rooms and their `chat_messages` were stranded on a deleted user. **Correction to the figures originally quoted here:** db3's log covers 26 May-17 Jul, and in that window there were 121 merges with 68 collision aborts - roughly one history-losing merge every day or two, not "68 in 2h". The damage per event is large, though: a collision aborts the whole multi-row `UPDATE`, stranding every room on that side for that user. Of the 119 merged users in the log, 42 still have stranded rooms: 2,410 rooms holding 17,030 chat messages.
- The merge now runs in a single transaction (`MergeUserAccounts` in `session/merge.go`), so a failure rolls back cleanly instead of half-applying across unsynchronized goroutines.
- Chat rooms are consolidated instead of blindly reassigned (`mergeChatRooms`):
  - Rooms directly between the two accounts (which would become self-chats) are deleted; FK cascade removes their messages and roster rows.
  - Rooms whose counterparty+chattype the survivor already holds get their `chat_messages` (and `refchatid` references), roster presence, and `latestmessage` recency migrated into the surviving room before the duplicate is deleted. Messages must move before the delete because `chat_messages.chatid` cascades on room deletion. Counterparty matching handles either `user1`/`user2` ordering, so opposite-ordering duplicates are consolidated too instead of leaving two rooms for the same pair.
  - Only collision-free rooms are reassigned.
- Roster rows for the merged user move to the survivor (`UPDATE IGNORE` + delete of leftovers), so unread pointers carry over where possible.
- `User2Mod`-style rooms (NULL `user2`) can't collide on the unique key (MySQL permits repeated NULLs), so they flow through the plain reassignment path unchanged.
- If the merge transaction fails, the email verification still succeeds and the failure is logged loudly; the duplicate account survives intact for a later merge instead of being half-dismembered.

## Code Quality Review

- The `Merged user %d into %d` log line format is unchanged - existing log greps keep working.
- Idempotent for accounts already half-merged by the old code: consolidation recomputes collisions from current state.
- Pre-existing gofmt drift elsewhere in `session.go` / `session_test.go` on master was left untouched; the new/edited regions are gofmt-clean.
- No docs page `covers:` these paths, so no docs updates are required.

## Test Plan

- New Go test `TestPatchSessionConfirmEmailMergeConsolidatesChatRooms` exercises, in one merge: a same-ordering collision (the case that aborted with error 1062 before), an opposite-ordering duplicate, a direct room between the two accounts, a non-colliding room, roster migration, message re-authoring, and `latestmessage` recency. Verified red against the old handler (fails with the exact production signature - error 1062 on `chat_rooms.user1_2`, rooms stranded on the merged user) and green with the fix.
- New Go test `TestMergeUserAccountsRollsBackOnFailure` forces a mid-merge foreign-key failure and proves the transaction rolls back atomically - nothing half-applied.
- Existing `TestPatchSessionConfirmEmailMergesUser` still passes.
- Full local Go suite green via the status API (3502 tests).

## Future Improvements

- **Backfill**: precisely identifiable from the node logs: 42 merged users with 2,410 stranded rooms / 17,030 messages (db3 log back to 26 May; db1/db2 logs only reach 7 Jul and add one more user). Merges before log coverage (the Go path shipped ~end of March) are an unknown extra tail without reliable DB-only identification. Because the new consolidation is idempotent, the natural backfill is to re-run it for each logged (survivor, loser) pair. Scope/approval pending.
- Merges run ~2/day, dominated by old accounts merging into freshly created ones - returning users signing up anew and then verifying their old email. Low volume, but the pattern may still interest product.
- After a merge the survivor can hold two `User2Mod` rooms for the same group (the unique key doesn't constrain NULL `user2`); pre-existing behaviour, not addressed here.
- V1's merge handled more reference tables (e.g. `users_chatlists_index` is left to its cache-refresh cycle); unchanged in scope here.

### DIFF
--- a/iznik-server-go/session/merge.go
+++ b/iznik-server-go/session/merge.go
@@ -1,0 +1,129 @@
+package session
+
+import (
+	"gorm.io/gorm"
+)
+
+// MergeUserAccounts merges the account `loser` into `survivor` after the
+// survivor proves ownership of one of the loser's email addresses via email
+// verification. All references move to the surviving user and the loser is
+// marked deleted. Everything runs in one transaction so a failure rolls back
+// cleanly rather than leaving the merge half-applied.
+func MergeUserAccounts(db *gorm.DB, survivor uint64, loser uint64) error {
+	return db.Transaction(func(tx *gorm.DB) error {
+		if err := tx.Exec("UPDATE messages SET fromuser = ? WHERE fromuser = ?", survivor, loser).Error; err != nil {
+			return err
+		}
+
+		if err := mergeChatRooms(tx, survivor, loser); err != nil {
+			return err
+		}
+
+		if err := tx.Exec("UPDATE chat_messages SET userid = ? WHERE userid = ?", survivor, loser).Error; err != nil {
+			return err
+		}
+
+		if err := tx.Exec("UPDATE users_emails SET userid = ? WHERE userid = ?", survivor, loser).Error; err != nil {
+			return err
+		}
+
+		// The survivor may already belong to some of the loser's groups —
+		// IGNORE skips those, then the leftovers are removed.
+		if err := tx.Exec("UPDATE IGNORE memberships SET userid = ? WHERE userid = ?", survivor, loser).Error; err != nil {
+			return err
+		}
+		if err := tx.Exec("DELETE FROM memberships WHERE userid = ?", loser).Error; err != nil {
+			return err
+		}
+
+		return tx.Exec("UPDATE users SET deleted = NOW() WHERE id = ?", loser).Error
+	})
+}
+
+// mergeChatRooms reassigns the loser's chat rooms to the survivor without
+// tripping the unique key on (user1, user2, chattype). A blind UPDATE fails
+// with a duplicate-key error when both users already have a room with the
+// same counterparty, aborting the whole statement and stranding those rooms
+// (and their history) on the about-to-be-deleted loser. Instead:
+//
+//  1. Rooms directly between the pair would become self-chats after the
+//     merge — delete them (their chat_messages and chat_roster rows go via
+//     FK cascade).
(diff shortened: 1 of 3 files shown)


edwh:
Live in production: this merge rode along in the master deploy of 8b28a0985 (08:31-08:34 UTC today, all of db1/db2/db3). Verified on all three nodes: running binary contains the consolidation path, monit OK, zero panics, zero `user1_2` collisions since restart. No email-verify merges have occurred yet today (they cluster in the evening on db3), so first real-world exercise of the new path is pending - the log line to watch is `Merged user X into Y` with no adjacent `Error 1062`.

edwh:
Correction to my deploy comment above: the "merges cluster in the evening" framing was wrong - it came from misreading the log window. db3's log spans 26 May-17 Jul: 121 merges (~2/day) with 68 collision aborts in that period, not 121 in 2 hours. Zero merges so far today is therefore normal, not a quiet morning. The PR description figures have been corrected too. Sizing for the backfill (read-only queries, shared DB): 42 logged merge-losers still hold 2,410 stranded rooms with 17,030 chat messages; loser→survivor pairs are recoverable from the node logs, so an idempotent re-run of the new consolidation per pair would repair them. Awaiting scope approval before writing that.

PR #1115, opened 2026-07-18
fix(ChitChat distance filter): stop Nearby falling back to unfiltered Anywhere

edwh:
## Root Cause
`newsfeed.GetNearbyDistance` (iznik-server-go/newsfeed/newsfeed.go) auto-sizes the radius for the ChitChat "Nearby" feed filter. It fetches up to 100 nearest posts from the spatial KNN index, then narrows to the 10th-nearest post that is both non-alert and posted in the last 31 days. Whenever it couldn't find 10 such posts — either because the spatial call errored/returned too few raw candidates, or (the common production case) because too few of the raw candidates passed the recency/alert filter — it returned distance `0`. `getFeed()` treats `0` identically to no-location-available: it applies **no geographic restriction at all**, producing exactly the same feed as "Anywhere". Nationwide only ~1,300 of ~150k newsfeed rows are recent+non-alert (0.9%), so this fallback triggers routinely outside busy areas, not just as an edge case.

Traced end-to-end: the Nuxt frontend (`pages/chitchat/[[id]].vue` → `stores/newsfeed.js` → `api/NewsAPI.js`) correctly sends `?distance=nearby`, and the Go handler (`newsfeed.Feed`) correctly parses it as "auto-calculate" (`gotDistance=false`) — both are working as designed. The bug is entirely inside the radius calculation itself.

Confirmed as a regression (not working-as-designed): the pre-#459 (`242d8a15c`) implementation always bounded its fallback to a maximum search radius (doubling 1km→128km before giving up), so it never returned "no restriction". The #459 spatial-server rewrite dropped that ceiling and returns 0 outright, which is Discourse #9937's exact symptom (a 169-mile-away post topping "Nearby").

## Evidence
Failing test output (before fix), `TestGetNearbyDistanceFallsBackToRawKNNWhenTooFewRecentPosts`:
```
newsfeed_nearby_test.go:115:
    Error Trace: .../test/newsfeed_nearby_test.go:115
    Error:       "0" is not greater than "0"
    Test:        TestGetNearbyDistanceFallsBackToRawKNNWhenTooFewRecentPosts
    Messages:    Nearby should still apply a geographic radius when there aren't enough recent posts to size one from alone, rather than silently becoming Anywhere
--- FAIL: TestGetNearbyDistanceFallsBackToRawKNNWhenTooFewRecentPosts (0.40s)
```
And the HTTP-level regression test, `TestFeedNearbyDistanceExcludesFarAwayPost` (Edinburgh user, London post ~332 miles away):
```
newsfeed_nearby_filter_test.go:57:
    Error:    Should be false
    Test:     TestFeedNearbyDistanceExcludesFarAwayPost
    Messages: far-away post should NOT appear under Nearby
--- FAIL: TestFeedNearbyDistanceExcludesFarAwayPost (0.33s)
```
Both pass after the fix; confirmed by running with the change stashed out (identical failures reproduce on unmodified master), proving the tests genuinely reproduce the bug.

## Fix
`GetNearbyDistance` now defaults to a bounded `maxNearbyKm = 128` ceiling — the largest radius the pre-#459 doubling-box algorithm ever actually searched before giving up — instead of 0/unrestricted, whenever there's no local data to size a tighter radius from (spatial error, empty result, or too few raw KNN candidates). When there are enough raw candidates (≥10) but too few pass the recency/alert filter, it now falls back to the raw KNN 10th-nearest distance (still real local density data) rather than the flat ceiling, so quieter areas get a radius that reflects their actual local activity instead of an arbitrary constant. The existing "enough recent, non-alert posts" happy path is unchanged.

## Other Call Sites
```
$ grep -rn "spatial.KNN" --include=*.go . | grep -v _test.go
./location/location.go:59: spatial.KNN("postcodes", ...)   # ClosestPostcode - single lookup, not a user-facing distance filter
./job/job.go:56:           spatial.KNN("jobs", ...)         # GetJobs - fails CLOSED (returns empty list) on error/empty, not open
./newsfeed/newsfeed.go:129: spatial.KNN("newsfeed", ...)    # fixed here
```
Neither other call site exhibits the "silently fall back to fully unrestricted" pattern that was unique to `GetNearbyDistance`'s overloaded 0-means-both-"no filter" reading — no changes needed there.

## Test
- File: `iznik-server-go/test/newsfeed_nearby_test.go` (`TestGetNearbyDistanceFallsBackToRawKNNWhenTooFewRecentPosts`) and `iznik-server-go/test/newsfeed_nearby_filter_test.go` (`TestFeedNearbyDistanceExcludesFarAwayPost`, new file)
- Command: `go test ./test/... -run 'TestGetNearbyDistanceFallsBackToRawKNNWhenTooFewRecentPosts|TestFeedNearbyDistanceExcludesFarAwayPost' -v`
- Proves fail-before/pass-after: both tests fail on unmodified master (confirmed via `git stash`) and pass after the fix. Full `newsfeed`/`Feed`-related suite (36 tests) passes with the change; `go build ./...` and `go vet` are clean.

## Discourse
https://discourse.ilovefreegle.org/t/9937/1

### DIFF
--- a/iznik-server-go/newsfeed/newsfeed.go
+++ b/iznik-server-go/newsfeed/newsfeed.go
@@ -102,6 +102,15 @@ type Newsfeed struct {
 	Previews       []NewsfeedPreview `json:"previews" gorm:"-"`
 }
 
+// maxNearbyKm bounds the "Nearby" radius whenever there's no local data to
+// size a tighter one from (spatial service error/empty result, or too few
+// raw KNN candidates to even try). The pre-#459 GetNearbyDistance doubled its
+// search box (1km, 2km, 4km, ...) and always used the last box it tried
+// before giving up, which never exceeded 128km — restoring that ceiling
+// keeps "Nearby" from ever falling through to a fully unfiltered feed like
+// "Anywhere" (Discourse #9937: a 169-mile-away post topped a "Nearby" feed).
+const maxNearbyKm = 128.0
+
 func GetNearbyDistance(uid uint64) (float64, utils.LatLng, float64, float64, float64, float64) {
 	const nearbyLimit = 10
 	// Over-fetch from the spatial index so that, after dropping alerts and posts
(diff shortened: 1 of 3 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #1115) — this fix did not pass and is not ready to merge.

- **Partial implementation**: maxNearbyKm is only a *default*, never a *cap*. Both data-driven paths write distKm unconditionally: `distKm = distDeg * 111.0` (line ~155) and `distKm = results[nearbyLimit-1].Distance * 111.0` (line ~163). Neither is clamped, so in a sparse area where the 10th raw KNN candidate is 2.5 degrees away
- **Test proves nothing / untested branch**: The new raw-KNN fallback branch (`results[nearbyLimit-1].Distance * 111.0`) — the only branch that can produce an unbounded radius — is not actually pinned by any test. TestGetNearbyDistanceFallsBackToRawKNNWhenTooFewRecentPosts asserts only `assert.Greater(t, dist, 0.0)`. With the mocked distances 

PR #1116, opened 2026-07-18
fix(jobs page on iOS - scroll/layout): exclude /jobs from the sticky footer ad

edwh:
## Root Cause

Two distinct bugs on iOS, per the reporter (Neville_Reid, iPhone SE3, iOS app 100.0.750):

1. The `/jobs` page's own job list relied entirely on document/body-level scroll (`.pageContent` in `LayoutCommon.vue` sets `max-height: 100vh`/`100dvh` but no `overflow-y`, so the page depends on the browser scrolling the whole document, enabled for list pages via a `class: 'overflow-y-scroll'` body class from `useHead`). On iOS the app is a Capacitor/WKWebView build (`capacitor.config.ts`: `ios.contentInset: 'automatic'`), and the reporter's repro path is specifically returning from a backgrounded external browser tab (they tapped a job, which opened the default browser, then returned to the app) - a known trigger for WKWebView's document-level scroll view becoming unresponsive after the page's dynamically-loaded (`client-only`, async-fetched) content changes height post-hydration.
2. `LayoutCommon.vue`'s sticky footer ad (`allowAd`) was shown on `/jobs` too, duplicating the same job ads in a small `position: fixed` panel with its own `overflow-y: auto` div (`JobsDaSlot.vue` `.jobs-slot`) - giving the page a second, independently-scrolling 2-row panel.

## Why the previous attempt was rejected

PR #1116 only fixed bug (2) - it excluded `/jobs` from `allowAd`, removing the duplicate panel - but left bug (1) (the page's own list not scrolling) completely unaddressed. Its test also didn't exercise any real logic: it read `LayoutCommon.vue` off disk as a string and regex-matched the `allowAd` computed block's source text, which would pass even if the logic did nothing at runtime.

What's different in this re-attempt:
- Fixes bug (1): `pages/jobs.vue`'s `.jobs-page` now has its own bounded, `overflow-y: auto` scroll region instead of depending on document scroll (see Fix below).
- Replaces the source-regex test with a real `@vue/test-utils` mount of `LayoutCommon.vue` that stubs `useRoute()` and asserts on what actually renders.

## Evidence

Failing-test output (`tests/unit/components/LayoutCommon.spec.js`) with the `/jobs` exclusion in `allowAd` temporarily reverted, proving the new test exercises real behaviour rather than passing vacuously:

```
 FAIL  tests/unit/components/LayoutCommon.spec.js > LayoutCommon > sticky ad > hides the sticky jobs ad footer on the dedicated /jobs page
AssertionError: expected true to be false // Object.is equality

- Expected
+ Received

- false
+ true

 ❯ tests/unit/components/LayoutCommon.spec.js:166:48
    164|       // that the logic behaves correctly).
    165|       const wrapper = mountLayoutAtRoute('/jobs')
    166|       expect(wrapper.find('.sticky').exists()).toBe(false)
       |                                                ^
```

With the fix restored, the same test (plus a contrasting "shows the sticky ad footer on an ordinary route" case, so the assertion can't pass by the ad simply never rendering) both pass: `tests/unit/components/LayoutCommon.spec.js (58 tests) passed`.

## Fix

- `iznik-nuxt3/pages/jobs.vue`: `.jobs-page` now gets `height: calc(100vh - <navbar height>)` (with `100dvh` fallback, at the same `md`/`xl` breakpoints `LayoutCommon.vue` uses for its navbar-height margin) plus `overflow-y: auto; -webkit-overflow-scrolling: touch;`, so the job list scrolls within its own CSS-computed region instead of depending on WKWebView's document-level scroll view. This mirrors the same already-working, iOS-tested pattern used by `ChatPane.vue` / `pages/chats/[[id]].vue` for full-height scrollable content below the fixed navbar. Also removed the now-unneeded `class: 'overflow-y-scroll'` body attribute, since the page no longer relies on body scroll.
- `iznik-nuxt3/components/LayoutCommon.vue`: unchanged in this re-attempt - the `/jobs` exclusion from `allowAd` (added in the original PR #1116 commit) is still correct and is kept.

This is a CSS/layout fix; the actual iOS WKWebView scroll-view bug this works around can't be reproduced in a unit-test environment (jsdom doesn't run real layout/scroll), so I couldn't write an automated test that fails specifically on that browser behaviour. I did verify the changed template compiles and renders correctly by copying it into a running dev container and confirming no Vite/SCSS compilation errors and a 200 response from `/jobs`; I did not have interactive browser-tool access in this session to visually confirm the scroll behaviour on an emulated iOS viewport.

## Other Call Sites

`grep -rn "routePath.value ===" --include=*.vue` finds only the one site in `LayoutCommon.vue` (unchanged). `grep -rn "jobs-page\|jobs-grid"` finds only `pages/jobs.vue` itself - no other component or e2e test depends on these classes.

## Test

`iznik-nuxt3/tests/unit/components/LayoutCommon.spec.js`
Command: `npx vitest run tests/unit/components/LayoutCommon.spec.js`
Proves fail-before (test fails when the `/jobs` exclusion is reverted, see Evidence) / pass-after (passes with the fix in place). Full unit suite (`npx vitest run`) also passes: 679 files, 14386 tests, 4 pre-existing skips - no regressions.

The `pages/jobs.vue` height/overflow CSS change itself is pure layout and not meaningfully testable at the unit level (see justification under Fix above).

## Discourse
https://discourse.ilovefreegle.org/t/9363/38

### DIFF
--- a/iznik-nuxt3/components/LayoutCommon.vue
+++ b/iznik-nuxt3/components/LayoutCommon.vue
@@ -202,8 +202,14 @@ const allowAd = computed(() => {
     routePath.value === '/partnerships' ||
     routePath.value === '/partnerships/' ||
     routePath.value === '/together' ||
-    routePath.value === '/together/'
+    routePath.value === '/together/' ||
+    routePath.value === '/jobs' ||
+    routePath.value === '/jobs/'
   ) {
+    // The /jobs page already shows a full scrollable list of job ads, so the
+    // sticky footer ad (which duplicates the same jobs in its own small,
+    // independently-scrollable panel) would give the page two scroll regions
+    // instead of one - see Discourse topic 9363 post #38.
     return false
   }
   return routePath.value !== '/' || loggedIn.value
--- a/iznik-nuxt3/pages/jobs.vue
+++ b/iznik-nuxt3/pages/jobs.vue
@@ -79,11 +79,7 @@ useHead(
     route,
     runtimeConfig,
     'Jobs',
-    'Freegle gets a little bit to help keep us going if you click on them.',
-    null,
-    {
-      class: 'overflow-y-scroll',
-    }
+    'Freegle gets a little bit to help keep us going if you click on them.'
   )
 )
 
@@ -176,9 +172,40 @@ onMounted(async () => {
 @import 'bootstrap/scss/variables';
 @import 'bootstrap/scss/mixins/_breakpoints';
 
+/* The page's job list must be its own scrollable region, sized to the space
+   below the fixed navbar (same convention as ChatPane.vue / pages/chats), and
+   must NOT rely on document/body scroll. On iOS the app (WKWebView, see
+   capacitor.config.ts contentInset: 'automatic') can leave the document-level
+   scroll view unresponsive after returning from a backgrounded external
+   browser tab, while an element with its own `overflow-y: auto` keeps working
+   - which is exactly what the reporter saw (Discourse topic 9363 post #38):
+   the page's own list was stuck, but the ad panel's own scrollable div (which
+   always used `overflow-y: auto`, see JobsDaSlot.vue) kept scrolling fine. */
 .jobs-page {
-  min-height: 100vh;
+  height: calc(100vh - 60px);
+  overflow-y: auto;
+  -webkit-overflow-scrolling: touch;
   background: $gray-100;
+
+  @supports (height: 100dvh) {
+    height: calc(100dvh - 60px);
+  }
+
+  @include media-breakpoint-up(md) {
+    height: calc(100vh - 66px);
+
+    @supports (height: 100dvh) {
+      height: calc(100dvh - 66px);
+    }
+  }
+
+  @include media-breakpoint-up(xl) {
+    height: calc(100vh - 76px);
+
+    @supports (height: 100dvh) {
+      height: calc(100dvh - 76px);
+    }
+  }
 }
 
 .jobs-page-content {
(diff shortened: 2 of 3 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #1116) — this fix did not pass and is not ready to merge.

- **Test proves nothing**: The new test in tests/unit/components/LayoutCommon.spec.js does not exercise `allowAd` at all. It reads LayoutCommon.vue off disk as a string, regex-extracts the `const allowAd = computed(...)` block, and asserts `expect(allowAdSrc).toMatch(/routePath\.value\s*===\s*['"]\/jobs['"]/)`. That asserts t
- **Partial implementation**: The bug as the author themselves restates it (test comment lines: 'a non-scrollable job list at the top plus a separate ~2-row independently scrollable panel below it') has two components: (a) the page's own job list did not scroll on iOS, and (b) a second independently-scrollable sticky panel exist

edwh:
Auto-closed by the monitor's adversarial review (PR #1116) — this fix did not pass and is not ready to merge.

- **partial-implementation**: The diff's own comment in pages/jobs.vue names a document-level, app-wide root cause ('on iOS the app (WKWebView...) can leave the document-level scroll view unresponsive after returning from a backgrounded external browser tab') but the fix is scoped to a single page. Every other page that relies o
- **coverage-gap-on-primary-fix**: The reported symptom is 'the job list would not scroll'. The fix for that symptom is the .jobs-page height/overflow rewrite in pages/jobs.vue - and it has zero test coverage of any kind. Both new tests (LayoutCommon.spec.js) assert only on the presence/absence of div.sticky, i.e. the ad-exclusion ha

PR #1117, opened 2026-07-18
fix(ModTools message search): surface aborted all-communities searches instead of reporting empty

edwh:
## Root Cause
`ListMessagesMT`'s all-communities member/message search runs a UNION-per-group query (`buildMTUnionAllMsgIDQuery`) capped with a `MAX_EXECUTION_TIME(20000)` hint, added in a previous fix (Discourse 9518/366) so a slow cross-group search returns rather than hanging forever. When that cap fires, MySQL aborts the statement with error 3024 (`ER_QUERY_TIMEOUT`) — but every `db.Raw(sql, args...).Pluck("msgid", &msgIDs)` call in `ListMessagesMT` discarded the returned error. An aborted search therefore left `msgIDs` empty, which the handler then reported as `{"messages": []}` — indistinguishable from a search that genuinely completed and found nothing. This exactly matches the report: a mod pastes a member name, the all-communities search silently returns nothing after ~20s (the cap), and searching again after picking one group (fast, well under the cap) finds it.

Live-DB check confirmed the mechanics: several moderators cover 20-79 groups, and `messages_groups` has ~7.4M Approved rows — a leading-wildcard `fullname`/`email` LIKE joined per group across that many groups is a realistic multi-second-to-timeout query, consistent with the "each search takes about 20 seconds" follow-up from the same reporter.

## Evidence
Failing-test output before the fix (from a temporarily-reverted copy of `message_list.go`):
```
$ go vet ./message/...
vet: message/message_list_timeout_test.go:68:17: undefined: isQueryTimeoutErr
```
The new detection logic doesn't exist pre-fix, so the inverted assertion can't even compile against the buggy code. The companion test in the same file (`TestSearchTimeout_BuggyPluckPatternCannotDistinguishAbortFromEmpty`), which uses only pre-existing APIs, compiles and **passes** both before and after the fix — it documents the defect directly: a real MySQL abort (verified against a live MySQL connection, error 3024) leaves `ids` empty with nothing to distinguish it from a genuine empty result.

## Fix
- Added `isQueryTimeoutErr` (`message/message_list.go`) to recognise MySQL error 3024 (`ER_QUERY_TIMEOUT`) via `errors.As`.
- Every `buildMTUnionAllMsgIDQuery`-backed query in `ListMessagesMT` (searchall by ID/subject, searchmemb by user ID/name, and the default listing) now checks the returned error; if any branch was aborted and the final result is empty, the response includes `"timedOut": true` instead of a bare empty list.
- `iznik-nuxt3`: `messageStore.fetchMessagesMT` now records `messageSearchTimedOut` from the response, and the ModTools approved-messages page shows a distinct warning ("that search took too long... try picking a specific community, or search again") instead of "Nothing found" when a search was aborted rather than completed.
- The 20s cap itself is unchanged — it still guards against the original hang (9518/366); this only stops the abort being reported as a definitive "not found".

## Other Call Sites
All 5 `db.Raw(...).Pluck("msgid", &msgIDs)` call sites in `ListMessagesMT` used the same discard-the-error pattern and are now fixed identically (searchall-by-id, searchall-by-subject, searchmemb-by-id, searchmemb-by-name, default listing). The `Edit`-collection branch uses a separate, unbounded query with no `MAX_EXECUTION_TIME` cap, so it isn't affected by this bug. The older `ListMessages` handler (`/api/messages`, used by the member-facing frontend, not ModTools) has no `MAX_EXECUTION_TIME` cap at all — out of scope for this fix.

## Test
- `iznik-server-go/message/message_list_pure_test.go`: unit tests for `isQueryTimeoutErr` (real 3024 error, wrapped error, unrelated errors, nil).
- `iznik-server-go/message/message_list_timeout_test.go`: live-DB AssertFlip pair reproducing a genuine MAX_EXECUTION_TIME abort (scaled down to ~100ms so it's deterministic and fast) and proving `isQueryTimeoutErr` correctly classifies it.
- `iznik-server-go/test/message_test.go`: `TestListMessagesMT_SearchMembAllCommunities` — regression test for the reported scenario (member-name search with no groupid across a mod's several communities finds the message, without `timedOut`).
- Full `message` package suite and the `ListMessagesMT`/search-related `test` package suite pass. The broader `test` package suite was run against an ad hoc throwaway DB (no dedicated status-API stack was provisioned for this isolated worktree); several unrelated pre-existing failures were found (session timezone drift, email-tracking config, rippling `outer_bound` schema quirk, missing routing/spatial fixtures) and each was individually confirmed to reproduce identically with this fix fully reverted, so they're environment gaps in that throwaway setup, not regressions from this change.

## Discourse
https://discourse.ilovefreegle.org/t/9938

### DIFF
--- a/iznik-nuxt3/modtools/pages/messages/approved/[[id]]/[[term]].vue
+++ b/iznik-nuxt3/modtools/pages/messages/approved/[[id]]/[[term]].vue
@@ -21,7 +21,19 @@
       <ModtoolsViewControl misckey="modtoolsMessagesApprovedSummary" />
     </div>
     <div>
-      <NoticeMessage v-if="loaded && !messages.length && !busy" class="mt-2">
+      <NoticeMessage
+        v-if="loaded && !messages.length && !busy && messageStore.messageSearchTimedOut"
+        variant="warning"
+        class="mt-2"
+      >
+        That search took too long across all your communities and was
+        stopped before it finished, so this isn't necessarily "nothing
+        found". Try picking a specific community, or search again.
+      </NoticeMessage>
+      <NoticeMessage
+        v-else-if="loaded && !messages.length && !busy"
+        class="mt-2"
+      >
         Nothing found. Almost always this is because the member or message
         doesn't exist (or has been very deleted).
       </NoticeMessage>
--- a/iznik-nuxt3/stores/message.js
+++ b/iznik-nuxt3/stores/message.js
@@ -32,6 +32,12 @@ export const useMessageStore = defineStore({
     // The context from the last fetch, used for fetchMore (ModTools)
     context: null,
 
+    // Whether the last fetchMessagesMT search was aborted by the server's
+    // query time cap rather than genuinely completing with no matches
+    // (Discourse 9938 - an all-communities member search that times out
+    // must not look identical to "no such member" in the UI).
+    messageSearchTimedOut: false,
+
     // Freegle Helper (AI concierge) state per bulk-offer msgid:
     // { batch, repliers, proposals, sent }. Loaded by the clearance management page.
     helper: {},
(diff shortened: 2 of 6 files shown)


edwh:

PR #1121, opened 2026-07-19
fix(login): don't conflate a DB lookup error with unknown-email/wrong-password across the login dispatcher

edwh:
## Root Cause
`PostSession`'s login-related handlers (`handleEmailPasswordLogin`, `handleLostPassword`, `handleUnsubscribe`, `handleLinkLogin`) each looked up a user with a bare `db.Raw(...).Scan(&x)`, discarding the query's error. When a lookup query itself errored (e.g. a replica connection dropped mid-query), the destination variable stayed at its Go zero value and was reported identically to "genuinely not found" — e.g. `ret:2 "We don't know that email address."` for `handleEmailPasswordLogin`. A backend failure was indistinguishable from "this account doesn't exist."

This matches the report exactly: a moderator's ModTools session crashed mid-navigation; on reopening they were shown the login page, entered their correct credentials, got "We don't know that email address.", then the same credentials worked moments later with no other change (Discourse #9941 post 1).

## Previous attempt and why it was rejected
PR #1119 fixed only `handleEmailPasswordLogin`'s own lookup and was auto-closed on review for two reasons, both addressed here:
1. **partial-implementation** — the identical swallowed-error pattern existed elsewhere in the *same* login dispatcher: `handleLostPassword` (literally the same message string), `handleUnsubscribe`, and `handleLinkLogin` (which has *two* such lookups — the user-exists check and the stored-key check), plus one call further into the chain in `auth.VerifyPassword`.
2. **test-does-not-isolate-the-fix** — the regression test swapped the process-global `database.DBConn` to a bad connection for the whole HTTP request, breaking every query in the pipeline rather than isolating the one lookup under test.

## Fix
Extracted each lookup into a small, exported, dependency-injected function that takes `db *gorm.DB` explicitly and uses `database.RetryQuery` (which already existed for exactly this purpose but had no callers) instead of a bare `db.Raw().Scan()`:
- `LookupUserIDByEmail`, `LookupUserAndEmailByEmail`, `LookupUserExists`, `LookupLinkCredentials` (session/session.go)
- `auth.VerifyPassword` now takes `db` explicitly and returns `(bool, error)`

Every one of `handleEmailPasswordLogin`, `handleLostPassword`, `handleUnsubscribe`, `handleLinkLogin`, the shared `getOrCreateLoginKey` helper, and `auth.VerifyPassword` now propagates a lookup error as a `500` backend failure instead of misreporting it as "unknown email" / "unknown user" / "wrong password". Recoverable connection errors are retried by `RetryQuery`; anything left after that is a real backend failure.

Taking `db` as an explicit parameter (rather than reading the `database.DBConn` global) means each function can be called directly in a test with a real-but-broken `*gorm.DB`, without touching global state or exercising the rest of the HTTP pipeline.

No new account-existence disclosure: the change strictly *removes* a case where a backend failure leaked as "unknown"/"wrong password" — it does not add any new distinguishing signal for real not-found vs. found.

## Evidence
AssertFlip: with the bug pattern temporarily reintroduced (each helper swallowing its query error again, matching the pre-fix code), all 5 new isolation tests fail:
```
--- FAIL: TestLookupUserIDByEmail_DBErrorNotConfusedWithUnknownEmail
    Error: An error is expected but got nil.
--- FAIL: TestLookupUserAndEmailByEmail_DBErrorNotConfusedWithUnknownEmail
    Error: An error is expected but got nil.
--- FAIL: TestLookupUserExists_DBErrorNotConfusedWithUnknownUser
    Error: An error is expected but got nil.
--- FAIL: TestLookupLinkCredentials_DBErrorNotConfusedWithNoKey
    Error: An error is expected but got nil.
--- FAIL: TestVerifyPassword_DBErrorNotConfusedWithWrongPassword
    Error: An error is expected but got nil.
```
With the fix restored, all 5 pass, and the full `go test ./...` suite (all packages) and the full `iznik-server-go/test` package (3500+ tests) pass clean.

## Other Call Sites
Same "bare `db.Raw(...).Scan()`, error discarded, zero value treated as not-found" pattern exists broadly elsewhere in the codebase (profile-display queries, membership/role checks, signup's own duplicate-email check in `user.PutUser`, `session/social_auth.go`, etc.) but is out of scope here — a repo-wide sweep is a separate, much larger change. This PR fixes every instance specifically in the credential/account lookup chain reachable from `PostSession`'s login/lost-password/unsubscribe/link-login dispatch (the exact scope the previous review flagged as incomplete), plus `auth.VerifyPassword` one call further into that same chain. `user.PutUser`'s call site of `VerifyPassword` was updated only for the new signature (its own separate `existingUID` lookup is unchanged and untouched).

## Test
- File: `iznik-server-go/test/login_dberror_test.go` (new)
- Tests: `TestLookupUserIDByEmail_DBErrorNotConfusedWithUnknownEmail`, `TestLookupUserAndEmailByEmail_DBErrorNotConfusedWithUnknownEmail`, `TestLookupUserExists_DBErrorNotConfusedWithUnknownUser`, `TestLookupLinkCredentials_DBErrorNotConfusedWithNoKey`, `TestVerifyPassword_DBErrorNotConfusedWithWrongPassword`
- Command: `go test ./test/... -run TestLookup\|TestVerifyPassword_DBError -v`
- Each opens an independent `*gorm.DB` connection pointed at `information_schema` (which has no `users`/`users_logins` tables — the connection succeeds, only the query errors) and passes it directly to the function under test. This does not touch `database.DBConn` or exercise any other part of the request pipeline — addressing the previous review's isolation complaint directly.
- Proves fail-before/pass-after: see Evidence above.
- Full regression: `go test ./...` (all packages) and `go test ./test/...` (3500+ tests) both pass clean against a freshly-migrated test database.

## Discourse

### DIFF
--- a/iznik-server-go/auth/auth.go
+++ b/iznik-server-go/auth/auth.go
@@ -20,6 +20,7 @@ import (
 	"github.com/freegle/iznik-server-go/utils"
 	"github.com/gofiber/fiber/v2"
 	"github.com/golang-jwt/jwt/v4"
+	"gorm.io/gorm"
 )
 
 // PersistentToken represents the old-style session token from the Authorization2 header.
@@ -241,14 +242,24 @@ func GetPasswordSalt() string {
 
 // VerifyPassword checks a plaintext password against a user's stored Native login.
 // We filter by userid only (not uid) because some legacy Native logins have NULL uid.
-func VerifyPassword(userID uint64, password string) bool {
-	db := database.DBConn
-
+//
+// Returns a non-nil error when the credentials lookup itself failed (a backend
+// problem) - retried via database.RetryQuery for transient connection errors.
+// Callers must not treat that the same as "password did not match": a DB error
+// here is a different failure mode from a wrong password and must be reported
+// as a backend failure, not a false claim about the supplied credentials
+// (Discourse #9941 - the same swallowed-error pattern that misreported the
+// email lookup as "unknown email" existed here too, one call further into the
+// same login path).
+func VerifyPassword(db *gorm.DB, userID uint64, password string) (bool, error) {
 	var logins []struct {
 		Credentials string
 		Salt        string
 	}
-	db.Raw("SELECT credentials, salt FROM users_logins WHERE userid = ? AND type = ? ORDER BY lastaccess DESC", userID, utils.LOGIN_TYPE_NATIVE).Scan(&logins)
+	err := database.RetryQuery(db, &logins, "SELECT credentials, salt FROM users_logins WHERE userid = ? AND type = ? ORDER BY lastaccess DESC", userID, utils.LOGIN_TYPE_NATIVE)
+	if err != nil {
+		return false, err
+	}
 
 	for _, login := range logins {
 		if login.Credentials == "" {
@@ -260,10 +271,10 @@ func VerifyPassword(userID uint64, password string) bool {
 		}
 		hashed := HashPassword(password, salt)
 		if strings.EqualFold(hashed, login.Credentials) {
-			return true
+			return true, nil
 		}
 	}
-	return false
+	return false, nil
 }
 
 // CreateSessionAndJWT creates a sessions row and returns the persistent token data and a JWT.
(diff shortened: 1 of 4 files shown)


edwh:

PR #1139, opened 2026-07-22
fix(mobile app notification badge): sync native icon badge from live unread state

edwh:
## Root Cause
The native app-icon badge (`mobileStore.setBadgeCount`) was only ever updated from an incoming push payload, or force-reset to 0 at startup. There is a client-side reactive sync too — `useNavbar.js`'s `chatCount` computed calls `setBadgeCount()` as a side effect — but it only fires when `NavbarMobile`'s bottom-nav badge is actually being rendered. On a phone, while viewing a specific chat (`/chats/:id` at xs/sm breakpoints), `NavbarMobile` hides that badge in favour of `ChatMobileNavbar` (`isSpecificChatPage`), which is exactly when a chat gets marked as read. `ChatMobileNavbar` calls `useNavbar()` too, but only destructures `{ online, backButtonCount, backButton }` — its own copy of `chatCount` is never read, so its `setBadgeCount` side effect never fires there.

If a member reads a chat and backgrounds/closes the app before visiting a screen where the bottom-nav badge re-renders, the OS icon badge is left showing whatever stale non-zero count a past push last set — exactly the reported symptom: a permanent notification blob with no actual unread replies.

## Evidence
`tests/unit/stores/mobile.spec.js` — `startBadgeSync` (AssertFlip, confirmed fails on pre-fix code with `store.startBadgeSync is not a function`, 4/60 tests failing; passes after fix, 60/60):
```
✓ mobile store > startBadgeSync > does nothing when not on the app
✓ mobile store > startBadgeSync > immediately applies the current live chat+notification total to the native badge
✓ mobile store > startBadgeSync > clears a stale non-zero badge once everything has been read in-app, with no new push
✓ mobile store > startBadgeSync > reacts live as the underlying chat/notification counts change

 Test Files  1 passed (1)
      Tests  60 passed (60)
```

## Fix
Added `mobileStore.startBadgeSync()` — a store-level `watch()` on `chatStore.unreadCount + notificationStore.count` that calls `setBadgeCount()` on every change, independent of which navbar component happens to be mounted. Called once from `initApp()` alongside the other app-startup wiring, so it fires immediately on load and then on every subsequent count change for the lifetime of the app session — including while the user is on the specific-chat page where the old component-coupled sync was blind.

## Other Call Sites
`grep -rn "setBadgeCount" iznik-nuxt3` — other call sites (`useNavbar.js`, `modtools/layouts/default.vue`, `modtools/composables/useModMe.js`) are the pre-existing component-coupled syncs; left as-is since they're harmless (idempotent, guarded by `lastBadgeCount` in `setBadgeCount`) and this fix supersedes them with an always-on path. No other code computes/sends the native badge count.

## Test
`iznik-nuxt3/tests/unit/stores/mobile.spec.js` (`startBadgeSync` describe block)
Command: `npx vitest run tests/unit/stores/mobile.spec.js`
Proves fail-before (`startBadgeSync is not a function`) / pass-after (60/60, badge synced from live counts including the stale-clear case).

## Discourse

### DIFF
--- a/iznik-nuxt3/stores/mobile.js
+++ b/iznik-nuxt3/stores/mobile.js
@@ -122,6 +122,34 @@ export const useMobileStore = defineStore({
       await this.initPushNotifications(PushNotifications, Badge)
       await this.checkForAppUpdate()
       this.initWakeUpActions(App)
+      this.startBadgeSync()
+    },
+
+    // Keep the native app-icon badge in sync with the member's own live
+    // unread state (chats + notifications), independent of which component
+    // happens to be on screen. useNavbar()'s chatCount computed also nudges
+    // the badge, but only as a side effect of NavbarMobile's bottom-nav badge
+    // being rendered - and that badge is hidden in favour of ChatMobileNavbar
+    // while viewing a specific chat on a phone (NavbarMobile.vue
+    // isSpecificChatPage), which is exactly when a chat gets marked read.
+    // ChatMobileNavbar calls useNavbar() too but never reads its chatCount,
+    // so that recompute (and its setBadgeCount call) never fires there. If
+    // the member reads the chat and backgrounds the app before visiting a
+    // screen where the bottom-nav badge re-renders, the icon badge is left
+    // showing a stale non-zero count forever (Discourse 9953). This watch
+    // lives in the store instead, so it fires on every count change
+    // regardless of what's mounted.
+    startBadgeSync() {
+      if (!this.isApp) return
+      const chatStore = useChatStore()
+      const notificationStore = useNotificationStore()
+      return watch(
+        () => (chatStore.unreadCount || 0) + (notificationStore.count || 0),
+        (total) => {
+          this.setBadgeCount(total)
+        },
+        { immediate: true }
+      )
     },
 
     async getDeviceInfo(Device) {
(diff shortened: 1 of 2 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #1139) — this fix did not pass and is not ready to merge.

- **Partial implementation / duplicate implementation**: The diff adds a second, independent writer of the native badge but leaves the root cause it identifies untouched. The new comment in stores/mobile.js states that `useNavbar()`'s `chatCount` computed 'also nudges the badge... as a side effect' of NavbarMobile rendering. That side-effect-inside-a-comp
- **Test proves nothing / reproduction test is not a reproduction**: All four new tests exercise only the newly-added `startBadgeSync()` method. Run against the pre-fix tree, `store.startBadgeSync` is undefined, so `stopWatch = store.startBadgeSync()` throws TypeError — the tests 'fail before the fix' for the trivial reason that the API does not exist, not because th
- **Coverage: reported trigger path untested**: The bug report path spelled out in the code comment (Discourse 9953) is: phone + viewing a specific chat, so NavbarMobile's bottom-nav badge is replaced by ChatMobileNavbar, so useNavbar()'s chatCount is never read, so the badge is never re-set when the chat is marked read. No test in the diff touch
- **Incomplete diff: fix not wired-in under test**: `this.startBadgeSync()` is added to the init path in stores/mobile.js, but no test asserts that init calls it. The entire suite calls startBadgeSync() directly with a hand-set `store.isApp = true`. If a later refactor drops the call from init, every test in this PR still passes and the bug silently 

PR #1155, opened 2026-07-22
fix(chat): refresh chat-list entry after sending, not just listByChatId

edwh:
## Root Cause
`chatStore.fetchChat(id)` in `iznik-nuxt3/stores/chat.js` wrote a freshly-merged object into `listByChatId[id]` but pushed the *raw* API response into the `list` array - two separate object references for the same chat. Any later in-place update via `listByChatId` (e.g. `send()`'s snippet patch) was therefore invisible in `list`, which is what the chat list sidebar actually renders (`pages/chats/[[id]].vue`), and whose row-visibility gate is `c.lastmsg > 0`.

A chat opened via the reply-to-post flow (`ChatButton.openChat`, used by `ChatReplyPane`'s state machine) starts with **no messages at all**, so `lastmsg` is 0 until the first message is sent. `chatStore.send()` never refreshed `lastmsg` after posting - it only patched `snippet` on `listByChatId`, which (per the above) never reached `list` anyway. Net effect: a chat created by replying to a Wanted/Offer post could never appear in the sidebar list, no matter how long the member waited.

Discourse 9955: Derek replied to a Wanted via the reply overlay and the resulting chat never showed up in his chat list, blocking him from attaching a photo to his reply.

## Evidence
Failing test (before the fix), run via the targeted vitest suite:
```
× tests/unit/stores/chat.spec.js > chat store > send > refreshes the list entry lastmsg after sending so a brand-new chat becomes visible (#9955) 6ms
  → expected +0 to be 555 // Object.is equality
FAIL  tests/unit/stores/chat.spec.js > chat store > send > refreshes the list entry lastmsg after sending so a brand-new chat becomes visible (#9955)
AssertionError: expected +0 to be 555 // Object.is equality
```
All other 50 existing tests in the file passed unchanged, confirming this was an isolated gap.

## Fix
- `fetchChat(id)`: merge into a single object and keep `listByChatId[id]` and the matching `list` array entry pointing at the *same* object (replacing it in place when already present, instead of only pushing when absent).
- `send()`: replaced the incomplete local-only `listByChatId[chatid].snippet = message` patch with `await this.fetchChat(chatid)`, so `lastmsg`/`lastdate`/`snippet`/`unseen` all refresh from the server immediately after any send - for every caller, not just the ones that separately remembered to call `fetchChat` afterwards (e.g. `ChatFooter.vue` already did this via `_updateAfterSend`; the reply-to-post flow via `ChatButton.vue` did not).

## Other Call Sites
`chatStore.send()` is called from `ChatFooter.vue`, `ModChatFooter.vue`, `ChatButton.vue` (reply-to-post), `ChatMessageAddress.vue`, `MessageReportModal.vue`, and `ModChatNoteModal.vue`. Since the fix lives inside the shared `send()`/`fetchChat()` store actions, all of these now consistently refresh the list entry - not just the reply-to-post path that exposed the bug. `unhide()` (which also calls `fetchChat`) benefits from the same in-place `list` refresh.

## Test
`iznik-nuxt3/tests/unit/stores/chat.spec.js` (new test in the `send` describe block, plus a `mockFetchChat` reset in `beforeEach` to avoid cross-test leakage now that `send()` depends on it).
Command: `POST /api/tests/vitest {"filter":"tests/unit/stores/chat.spec.js"}` via the status API.
Proves fail-before (`expected +0 to be 555`) / pass-after (51/51 passed) against the exact production code path.

## Discourse
https://discourse.ilovefreegle.org/t/9955/1

### DIFF
--- a/iznik-nuxt3/stores/chat.js
+++ b/iznik-nuxt3/stores/chat.js
@@ -294,15 +294,29 @@ export const useChatStore = defineStore({
         try {
           const chat = await api(this.config).chat.fetchChat(id, false)
           if (chat) {
-            // Merge to preserve any existing data (e.g. from listing).
-            this.listByChatId[id] = {
+            // Merge to preserve any existing data (e.g. from listing), and keep
+            // listByChatId and the list array pointing at the SAME merged object.
+            // Previously these were two separate copies (a spread into
+            // listByChatId, the raw fetch result into list), so an in-place
+            // mutation of one (e.g. send() patching the snippet) was invisible
+            // in the other. The chat list panel renders from `list`, so a chat
+            // whose only update went through listByChatId never appeared to
+            // have moved/changed - including its lastmsg, which gates whether
+            // the row renders at all (Discourse 9955).
+            const merged = {
               ...this.listByChatId[id],
               ...chat,
             }
-            // If this chat isn't in the list (e.g. too old to appear in fetchChats
-            // results), add it so it can be rendered in the chat list panel.
-            if (!this.list.find((c) => c.id === id)) {
-              this.list.push(chat)
+            this.listByChatId[id] = merged
+
+            const idx = this.list.findIndex((c) => c.id === id)
+            if (idx === -1) {
+              // Not in the list (e.g. too old to appear in fetchChats results,
+              // or brand new) - add it so it can be rendered in the chat list panel.
+              this.list.push(merged)
+            } else {
+              // Already in the list - refresh it in place with the fresh data.
+              this.list[idx] = merged
             }
           }
         } catch (e) {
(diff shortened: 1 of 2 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #1155) — this fix did not pass and is not ready to merge.

- **Known regression / error handling**: send() now does `await this.fetchChat(chatid)` AFTER the message has already been accepted by the server, and fetchChat rethrows every non-404 error (stores/chat.js:307-316: `if (e?.response?.status === 404) { delete this.listByChatId[id] } else { throw e }`). So a flaky follow-up GET turns a SUCCES
- **Duplicate implementation**: The refresh added to send() already exists at the main call site: ChatFooter.vue:643-652 `_updateAfterSend()` does `await fetchMessages()` then `await chatStore.fetchChat(props.id)` after every footer send. With this diff the normal in-chat send path performs two chat GETs and two fetchMessages per 
- **Partial implementation / root cause not addressed**: The diff makes fetchChat write into `this.list`, but `list` is replaced wholesale on every listing refresh: fetchChats does `this.list = chats` (stores/chat.js:~281) and listChatsMT does `this.list = chatrooms` (~line 116), and pollForChatUpdates re-runs fetchChats every 30 s (setTimeout(this.pollFo

PR #1161, opened 2026-07-23
feat(post): warn to check contact lens solution is sealed and in date

edwh:
## Sealed-and-in-date posting warning
Discourse [8643/18](https://discourse.ilovefreegle.org/t/all-in-one-contact-lens-solution/8643/18): a compose-time advisory (like the existing fire-label / car-seat / knife guidance) for items that need to be sterile and in date. Advisory only, not a block.

`components/PostItem.vue` — new `warnings` entry: type "Sealed and in date", message "Please make sure this is sealed and in date.", keywords covering contact lens solution/fluid/cleaner, cosmetics, make-up, mascara, and sunscreen/sun cream/sun lotion/aftersun.

## Duplicate-detection fix
`PostItem`'s `duplicate` computed reads `messageStore.all` (the `list` dict) and matches on `item.name`, but nothing loaded the member's own posts there on a cold `/give`/`/find` visit. `useCompose` fired `fetchByUser(myid, false)`, which only fills `byUserList` with lean summaries that omit `item.name` — a store slot the check never reads — so the "you already have an open post … use Repost" warning never appeared unless the member had already visited `/myposts` that session. `useCompose` now fetches the member's active posts and pulls each full record into `list` (mirrors `pages/myposts.vue`), so the check works on a direct load.

## Tests
- `tests/unit/components/PostItem.spec.js` — keyword coverage for the new warning; existing duplicate-detection unit tests.
- `tests/e2e/test-post-item-warnings.spec.js` — warning renders, item still posts (OFFER/WANTED), the edit-modal path, and a duplicate-detection test that now loads `/give` with a fresh page (`page.goto`) so it exercises the compose-time load rather than the previous client-side-nav workaround.
- Coverage for previously-untested code: `iznik-server-go/rippling/analytics_extra_test.go`, `iznik-server-go/town/select_near_extra_test.go`, `iznik-batch/tests/Unit/Services/{EeeSqliteService,MicrovolunteeringNotifyService}Test.php`, and e2e coverage for several 0%-covered informational pages.
- Fixes a pre-existing flaky assertion in `test-modtools-move-message.spec.js`: its group-detection loop matched the first group with any visible `.card` instead of the target message's `#msg-<id>`.

## Discourse
https://discourse.ilovefreegle.org/t/all-in-one-contact-lens-solution/8643/18

### DIFF
--- a/iznik-batch/tests/Unit/Services/EeeSqliteServiceTest.php
+++ b/iznik-batch/tests/Unit/Services/EeeSqliteServiceTest.php
@@ -1,0 +1,950 @@
+<?php
+
+namespace Tests\Unit\Services;
+
+use App\Services\EeeSqliteService;
+use PDO;
+use Tests\TestCase;
+
+/**
+ * Unit tests for EeeSqliteService.
+ *
+ * The service owns its own SQLite file (separate from the MySQL `iznik` DB used
+ * elsewhere in the app), so every test points `freegle.eee.sqlite_path` at a
+ * fresh temp file and constructs a brand-new service instance — no shared
+ * state, no MySQL fixtures required.
+ */
+class EeeSqliteServiceTest extends TestCase
+{
+    /** @var string[] */
+    private array $tempPaths = [];
+
+    protected function tearDown(): void
+    {
+        foreach ($this->tempPaths as $path) {
+            foreach ([$path, $path . '-wal', $path . '-shm', $path . '-journal'] as $file) {
+                if (is_file($file)) {
+                    @unlink($file);
+                }
+            }
+            $dir = dirname($path);
+            if (is_dir($dir) && str_contains($dir, 'eee_sqlite_test_')) {
+                @rmdir($dir);
+                $parent = dirname($dir);
+                if (is_dir($parent) && str_contains($parent, 'eee_sqlite_test_')) {
+                    @rmdir($parent);
+                }
+            }
+        }
+        $this->tempPaths = [];
+
+        parent::tearDown();
+    }
+
+    private function newService(?string $path = null): EeeSqliteService
+    {
+        $path ??= sys_get_temp_dir() . '/eee_sqlite_test_' . uniqid() . '.sqlite';
+        $this->tempPaths[] = $path;
+        config(['freegle.eee.sqlite_path' => $path]);
+        return new EeeSqliteService();
+    }
+
+    // ─────────────────────────────────────────────────────────────────────────
+    // getPdo() / migrate() bootstrap
+    // ─────────────────────────────────────────────────────────────────────────
+
+    public function test_get_pdo_creates_missing_nested_directory_and_file(): void
+    {
+        $path = sys_get_temp_dir() . '/eee_sqlite_test_' . uniqid() . '/nested/classifications.sqlite';
+        $this->tempPaths[] = $path;
+        config(['freegle.eee.sqlite_path' => $path]);
+        $service = new EeeSqliteService();
+
+        $this->assertFileDoesNotExist($path);
+
+        $pdo = $service->getPdo();
+
+        $this->assertInstanceOf(PDO::class, $pdo);
+        $this->assertFileExists($path);
+    }
+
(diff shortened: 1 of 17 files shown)


edwh:
## Coverage added (playwright)
Modules: `iznik-nuxt3/components/ProxyImage.vue`, `iznik-nuxt3/components/SupportLink.vue`

Root-caused the red check rather than guessing: diffed the Coveralls per-file job report for this PR's playwright job (185097921) against master's baseline job (185072049) via the public `source_files.json` API — no test run needed. Exactly one file differed: `ProxyImage.vue`, `covered_line_count` 9→7 out of 12 between the two runs. Same disease as the previously-fixed `SpinButton.vue`/`LoginModal.vue` jitter (#910/#1007): nothing waited for the pages exercising its less-common prop branches (`preload`+`fetchpriority`, non-http src + query-string) to actually finish rendering, so a bare title-check test only got branch coverage credit some of the time.

Fix: two deterministic specs with explicit web-first waits on the rendered `` elements/attributes (not just page titles) — the logged-out homepage's `FreeglerPhotoGrid` (image 1 gets `preload`+`fetchpriority="high"`, image 3 gets neither) and `NationalReuseDay`'s banner (`src="/NRD/Banner.png?a=1"`, non-http + query-string). Also picked up `SupportLink.vue` (0% coverage, `<client-only>`) on the same page.

Coverage before (master baseline, job 185072049): 22.123%
Coverage before (this PR's last failing run, job 185097921 / b34e82800): 22.075% (-0.05%)
Coverage after: Coveralls will report this on the CI run for this push (pipeline #9790) — not polled to completion here to stay within session budget; the fix targets the exact single file confirmed to be the entire diff between the two prior runs.

## Tests added
| Area | Scenario | File:Line |
|---|---|---|
| ProxyImage preloadValue/imgAttrsComputed | Logged-out homepage grid: image 1 preload+fetchpriority="high", image 3 neither | test-proxy-image-branches.spec.js:27 |
| ProxyImage fullSrc | National Reuse Day banner: non-http prefix + query-string encode branches; logged-out SupportLink renders | test-proxy-image-branches.spec.js:52 |
| SupportLink href (logged-in) | Skipped — see latent bug | test-proxy-image-branches.spec.js:79 |

Ran locally twice against the running dev stack (`npx playwright test tests/e2e/test-proxy-image-branches.spec.js`), both green (2 passed, 1 skipped).

## Latent bugs found
| Where | Description |
|---|---|
| `SupportLink.vue:29` | `const myid = authStore.user?.id` is captured once, non-reactively, at this `<client-only>` component's setup time. Confirmed with a genuinely authenticated session (login + `/qr` showing real mod-only content in the same test) that the mailto footer still renders the logged-out text — SupportLink's own mount raced ahead of the auth store restoring from localStorage, and nothing re-evaluates `myid` afterwards. Not fixed here; test skipped with `// TODO: latent bug` at test-proxy-image-branches.spec.js:79.

## No production code changed
All changes are test files only (`iznik-nuxt3/tests/e2e/test-proxy-image-branches.spec.js`).

edwh:
## Correction to my previous comment

The previous comment claimed a latent reactivity bug in `SupportLink.vue`. That was wrong — investigated further rather than leaving it as a skipped test (per repo policy against `test.skip()`).

**Actual root cause:** the failing assertion was on National Reuse Day, whose page uses `layout: 'no-navbar'`. Only `layouts/default.vue`/`layouts/login.vue` call `useBootSession()`/`fetchMe()`, so `authStore.user` is never populated on that page at all, regardless of login — nothing to do with how `SupportLink.vue` reads it. Confirmed directly against the actual running `freegle-prod-local` container's unmodified `SupportLink.vue` (it's a fully-baked image, not bind-mounted, so this was a clean test of the real component): the original code renders the logged-in mailto text correctly once tested on a page whose layout actually boots the session.

**Fix:** moved the logged-in assertion to `/mobile` (default layout, plain unconditional `<SupportLink/>`). No production code was touched in either commit on this branch. All 3 tests in `test-proxy-image-branches.spec.js` now pass for real, verified 3x locally against the unmodified production component.

PR description updated to match.

PR #1163, opened 2026-07-23
fix(browse): index cross-posted/rippled posts per group in messages_spatial (9954)

edwh:
## The bug (Discourse 9954, "Post disappeared after I approved it")
A moderator approved a Wanted, then couldn't find it **"with my member hat on"** (i.e. browse). The post was Approved and fully intact — not withdrawn, not deleted. It was **invisible in browse on her own group** because its `messages_spatial` row was on a *different* group.

## Root cause
`messages_spatial` backs the public browse/map and needs **one row per group** a message is approved on (a cross-posted or rippled-out post shows in browse on each of its groups). Both writers — the Go inserter `addApprovedMessageToSpatialIndex` and the every-5-minute `MessageSpatialService` reconciler — do `INSERT ... ON DUPLICATE KEY UPDATE` **assuming the unique key is `(msgid, groupid)`**.

But the table was created with a **single-column `UNIQUE (msgid)`** (`2025_12_10_..._create_messages_spatial_table`, `->unique('msgid')`). So the second and later groups collapsed onto the first-inserted row:

- a rippled/cross-posted post ends up indexed on **only one, non-deterministic group** → invisible in browse on all the others, including its origin group;
- the reconciler's `ON DUPLICATE KEY UPDATE` also updates `groupid`, so under `UNIQUE(msgid)` it kept **flipping the single row between groups** every 5 minutes (observed live: the one row's groupid changed between two queries), never settling.

A scope check found **thousands** of approved posts in the last 2 days missing their group's spatial row.

## Fix
- **Migration** `2026_07_23_000001_messages_spatial_per_group_unique_key`: swap the unique key `msgid` → `(msgid, groupid)`. The composite still serves `WHERE msgid = ?` (msgid leads); the writers' inserts now create one row per group instead of overwriting, and the churn stops. Idempotent, defensive (checks `information_schema`), with a reversible `down()`.
- **Backfill** in the same migration (bounded to messages already in the index, `INSERT IGNORE`) so the browse gap clears immediately instead of waiting for each post to be re-touched.
- **Regression** in `MessageSpatialServiceTest`: a two-group message must be indexed once per group.

## Validation
Migration PHP syntax-checked (`php -l`) via the batch container. Migration + Laravel suite run in CI (the local WSL toolchain is gated). Note: after deploy, consider a wider one-time backfill for older un-touched posts (the reconciler self-heals recent ones).

https://claude.ai/code/session_01RjavhAHn6siYfunR5bacK6

### DIFF
--- a/iznik-batch/database/migrations/2026_07_23_000001_messages_spatial_per_group_unique_key.php
+++ b/iznik-batch/database/migrations/2026_07_23_000001_messages_spatial_per_group_unique_key.php
@@ -1,0 +1,102 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * messages_spatial backs the public browse/map, with one row PER GROUP a message
+     * is approved on - a cross-posted or rippled-out post must appear in browse on
+     * each of its groups. The Go inserter (addApprovedMessageToSpatialIndex) and the
+     * every-5-minute reconciler both do `INSERT ... ON DUPLICATE KEY UPDATE` assuming
+     * the unique key is (msgid, groupid), and deliberately never update groupid on
+     * conflict.
+     *
+     * But the table was created with a single-column `UNIQUE (msgid)`. So the second
+     * and subsequent groups for a message collapsed onto the first-inserted row: a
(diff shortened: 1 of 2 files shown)


edwh:
Closing - based on a wrong model. messages_spatial.groupid is historical, and browse visibility is driven by message reach (rippling_reach), not per-group spatial rows. Re-diagnosing #9954 via reach.

PR #1164, opened 2026-07-23
fix(search): re-index and re-embed a message when its subject/body is edited (9954)

edwh:
## Root Cause

Support Tools / ModTools message search reads `messages_index` (keyword search, joined to `words`/`messages_spatial` in `iznik-server-go/message/search.go` `GetWordsExact`/`GetWordsTypo`/`GetWordsStarts`/`GetWordsSounds`) and `messages_embeddings` (vector search). Both are populated **once**, by background jobs that only pick up messages *missing* from the table (`MessageSearchService::indexUnindexedMessages`, `whereNotIn (SELECT msgid FROM messages_index)`; `GenerateEmbeddingsCommand`, `LEFT JOIN messages_embeddings ... WHERE me.msgid IS NULL`). Nothing in the Go V2 message-edit path wrote to either table on edit, so a word introduced by an edit (e.g. adding "Moulinex" to a Wanted's subject) was never indexed and never searchable, while the original wording kept matching.

## Evidence

Failing test run against the pre-fix code (`TestPatchMessageSubjectEditInvalidatesSearchIndexes`, effectively unchanged from #1165's repro):

```
--- FAIL: TestPatchMessageInvalidatesSearchIndexes (0.13s)
    Error: searching for a word added in an edit should find the message - the search index must be refreshed on edit
```

## Fix

Added `invalidateMessageSearchIndexes(db, msgid, subjectChanged, textChanged)` in `iznik-server-go/message/message.go`: on an edit, it drops the message's stale `messages_index`/`messages_embeddings` rows so the existing background indexer/embedder rebuild them from the new text. Wired into `applyPatchMessageCore` (any editor, owner or mod), `handleApproveEdits` (a pending edit is approved), and `handleRevertEdits` (a pending edit is rejected and the previous subject/body restored - same gap, that path writes `messages.subject`/`textbody` directly).

**Re-attempt refinement:** `messages_index` is derived from the message **subject only** (`MessageSearchService::indexString` is only ever called with subject text — see `indexUnindexedMessages`); `messages_embeddings` is derived from **subject + textbody** (`GenerateEmbeddingsCommand`'s `SELECT ms.msgid, m.subject, LEFT(m.textbody, 500) as body`). The original combined-fix version invalidated `messages_index` on `subjectChanged || textChanged`, which meant a **body-only** edit dropped perfectly valid, unrelated keyword-index rows — making the post briefly unsearchable by keyword (until the next `messages:update-index` run, every 30 min) for no reason, since the words that drive keyword search hadn't changed. `invalidateMessageSearchIndexes` now takes independent flags: `messages_index` only drops on `subjectChanged`; `messages_embeddings` drops on `subjectChanged || textChanged`. Confirmed `messages:update-index` (30 min) and `embeddings:generate` (5 min) are both live-scheduled in `iznik-batch/routes/console.php` (the "NOT YET ENABLED" comment above `messages:update-index` is stale — it was enabled in commit `132a6bad7`), so this is a bounded re-index window, not permanent staleness.

## Why the previous attempt (#1165) was rejected

#1165 only refreshed `messages_index` (keyword search) and didn't cover `messages_embeddings` (vector search) or the revert-edit path; it was closed as a duplicate/subset of this PR, which combines both. This re-attempt fixes a correctness gap introduced when combining them: the combined version over-invalidated the keyword index on body-only edits.

## Other Call Sites

Grepped every `UPDATE messages SET subject|textbody` in `iznik-server-go/message/message.go`:

- `handleApproveEdits`, `handleRevertEdits` — covered (see Fix).
- Draft→live promotion subject reconstruction (two sites, e.g. around submitting a draft to a group) — run before a message is ever approved/indexed for the first time, covered by the normal initial-indexing job, not this edit-refresh gap.
- `applyPatchMessageCore`'s own subject rebuild from item/type/location — already covered, it runs before the `current`/`subjectChanged` snapshot that triggers the invalidation call.
- **Known follow-up (not fixed here, out of scope for 9954):** the Bulk Offer catalogue rebuild in `applyPatchMessageCore` (`req.Bulkitems != nil` branch) regenerates `textbody` from a computed summary *after* the `subjectChanged`/`textChanged` snapshot is taken, when the caller didn't supply `req.Textbody` directly. That regenerated text isn't covered by the invalidation check, so `messages_embeddings` could go stale on a bulk-item edit that doesn't also touch subject/textbody directly. Distinct, narrower bug in a different feature area — flagging for a separate fix rather than expanding this PR's scope.

## Test

- File: `iznik-server-go/test/message_reindex_on_edit_test.go`
  - `TestPatchMessageSubjectEditInvalidatesSearchIndexes` — reproduces the reported bug (subject edit) and proves both indexes clear.
  - `TestPatchMessageBodyOnlyEditPreservesKeywordIndex` — new; proves a body-only edit clears the embedding but leaves the still-valid keyword index alone. Fails without the refinement above (`expected: 4, actual: 0` — messages_index wrongly wiped), passes with it.
- Command: `go test ./test/... -run 'TestPatchMessageSubjectEditInvalidatesSearchIndexes|TestPatchMessageBodyOnlyEditPreservesKeywordIndex' -v` (also ran `TestPostMessageApproveEdits`, `TestPostMessageRevertEdits`, and the `TestPatchMessageEdit*` review-flow tests - all pass, no regressions from the signature change).
- Command run against a stashed (pre-fix) working tree confirmed the pre-existing, unrelated `users.tnuserid` duplicate-key failures elsewhere in `message_test.go` are caused by concurrent parallel test runs sharing this DB, not this change.

## Discourse

https://discourse.ilovefreegle.org/t/9954/3

### DIFF
--- a/iznik-server-go/embedding/store.go
+++ b/iznik-server-go/embedding/store.go
@@ -258,6 +258,26 @@ func (s *Store) FindByMsgid(msgid uint64) (Entry, bool) {
 	return Entry{}, false
 }
 
+// Evict removes the entry for msgid from the in-memory store, if present, and
+// reports whether one was removed. Used when a message's embedding is invalidated
+// on edit: the DB row is deleted so the batch re-embeds the new content, but
+// Refresh() is presence-keyed (see its "Known limitation") and would otherwise keep
+// the STALE blob for a msgid it already holds if the delete+re-embed both land
+// between two refresh ticks. Evicting forces the next Refresh to treat the msgid as
+// new and reload the regenerated embedding, so vector search stops matching the old
+// wording (Discourse 9954).
+func (s *Store) Evict(msgid uint64) bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	for i := range s.entries {
+		if s.entries[i].Msgid == msgid {
+			s.entries = append(s.entries[:i], s.entries[i+1:]...)
+			return true
+		}
+	}
+	return false
+}
+
 // VectorSearchResult from vector search. SubjectCos and BodyCos are the pure
 // per-field cosines; HasBody distinguishes "body exists but cosine is 0" from
 // "no body embedding" (BodyCos is 0 in both cases). The caller decides how to
--- a/iznik-server-go/message/message.go
+++ b/iznik-server-go/message/message.go
@@ -2110,12 +2110,20 @@ func addApprovedMessageToSpatialIndex(db *gorm.DB, msgid uint64) {
 // body-only edit must not drop the keyword index - those rows still accurately reflect the
 // unchanged subject, and dropping them would make the message unsearchable by keyword for
 // no reason until the next background run.
+//
+// Deleting the messages_embeddings row is necessary but not sufficient for vector search:
+// apiv2 serves vector search entirely from an in-process store (embedding.Global) that
+// Refresh()es every ~2 min and is presence-keyed, so a delete+re-embed landing between two
+// ticks would leave the STALE embedding in memory (see Store.Refresh's "Known limitation").
+// We therefore also Evict the msgid from that store so the next Refresh reloads the
+// regenerated blob.
 func invalidateMessageSearchIndexes(db *gorm.DB, msgid uint64, subjectChanged bool, textChanged bool) {
 	if subjectChanged {
 		db.Exec("DELETE FROM messages_index WHERE msgid = ?", msgid)
 	}
 	if subjectChanged || textChanged {
 		db.Exec("DELETE FROM messages_embeddings WHERE msgid = ?", msgid)
+		embedding.Global.Evict(msgid)
 	}
 }
 
--- a/iznik-server-go/message/message.go
+++ b/iznik-server-go/message/message.go
@@ -2096,16 +2096,27 @@ func addApprovedMessageToSpatialIndex(db *gorm.DB, msgid uint64) {
 	}
 }
 
-// invalidateMessageSearchIndexes drops the keyword-index (messages_index) and vector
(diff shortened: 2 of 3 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #1164) — this fix did not pass and is not ready to merge.

- **partial-implementation**: The vector-search half of the fix does not reliably take effect. apiv2 keeps embeddings in an in-process store (iznik-server-go/embedding/store.go) refreshed every 2 minutes (main.go:92). Store.Refresh() is presence-keyed, not content-keyed: it computes openIds from messages_embeddings, keeps every 

edwh:
Reopened — the adversarial review's finding was **valid**, and this push fixes it.

**The finding (correct):** deleting the `messages_embeddings` row triggers a batch re-embed with the new content, but apiv2 serves vector search entirely from an in-process store (`embedding.Global`). `Store.Refresh()` is presence-keyed (its own doc-comment "Known limitation"): a msgid it already holds is `kept` with its old blob and never re-fetched. So a delete + re-embed landing between two 2-min refresh ticks leaves the **stale in-memory embedding** matching the *old* wording — exactly the 9954 symptom, just moved from the DB to memory.

**The fix (`757d43…` → new commit):** `invalidateMessageSearchIndexes` now also calls `embedding.Global.Evict(msgid)` whenever it drops the embedding row. Eviction forces the next `Refresh` to treat the msgid as new and reload the regenerated blob. Added `Store.Evict()` plus a test that seeds the store, edits the message, and asserts the entry is evicted.

Keyword-index handling is unchanged from the review's earlier refinement (drops only on subject change, since `messages_index` is subject-derived). gofmt-clean; CI re-running.

PR #1165, opened 2026-07-23
fix(search): refresh full-text index when a message subject/body is edited

edwh:
## Root Cause

Full-text search (Support Tools message search, and member search - `messages_index` joined to `messages_spatial`/`words` in `iznik-server-go/message/search.go`) is populated by the nightly iznik-batch catch-up job `MessageSearchService::indexUnindexedMessages`. That job only indexes messages it has **never** indexed before (`msgid NOT IN (SELECT msgid FROM messages_index)`), so once a message is indexed it is never revisited.

Nothing in the Go V2 message-edit path (`applyPatchMessageCore` in `iznik-server-go/message/message.go`) wrote to `messages_index` either. So editing a message's subject to add new wording left the OLD wording searchable and the NEW wording invisible - exactly what Angelika reported: message 121085393 ("WANTED: Spindle/stem for Moulinex food processor" on Braintree) was edited on 22 July to add "Moulinex food processor", but a Support Tools search for "Moulinex" found nothing, while "spindle"/"stem" (the original wording) still matched.

## Evidence

Failing test run against the pre-fix code (`TestPatchMessageSubjectEditRefreshesSearchIndex`):

```
=== RUN   TestPatchMessageSubjectEditRefreshesSearchIndex
    message_test.go:7700:
        Error Trace: /app/test/message_test.go:7700
        Error:       Should be true
        Test:        TestPatchMessageSubjectEditRefreshesSearchIndex
        Messages:    searching for a word added in an edit should find the message - the search index must be refreshed on edit
--- FAIL: TestPatchMessageSubjectEditRefreshesSearchIndex (0.13s)
FAIL
```

## Fix

Added `reindexMessageSearchWords(db, msgid)` in `iznik-server-go/message/search.go`: it re-derives the message's indexed words from its **current** subject (reusing the existing `GetWords` common-word filter, truncated to `words.word`'s `varchar(10)`), deletes the message's old `messages_index` rows, and re-inserts the current set - mirroring `MessageSearchService::indexString`'s upsert shape.

Wired it into `applyPatchMessageCore` (the single core function behind both `PATCH /message` and `PATCH /message/tn/:tnpostid`) whenever the subject changes, and into `handleRevertEdits` when a moderator rejects a pending edit and the subject is restored to its pre-edit wording (same gap: that path writes `messages.subject` directly, bypassing the edit hook).

Only messages currently in the `Approved` collection are reindexed - `messages_index` is always joined to `messages_spatial`, which (like the batch job) only has rows for Approved messages, so indexing a Pending/Rejected edit would never surface in a search result anyway.

## Other Call Sites

Grepped every place `messages.subject` is written in `iznik-server-go`:

- `handleApproveEdits` (message.go) - re-applies the *same* subject that `applyPatchMessageCore` already wrote live and already reindexed; no-op for search, left as-is.
- Draft→live promotion / new-post subject reconstruction (message.go, two sites) - these run before a message is ever approved/indexed for the first time, so they're covered by the normal initial-indexing job, not this edit-refresh gap.
- `applyPatchMessageCore`'s own subject rebuild from item/type/location - already covered, since it runs before the `subjectChanged` check that triggers the new reindex call.

## Test

- File: `iznik-server-go/test/message_test.go` - `TestPatchMessageSubjectEditRefreshesSearchIndex`
- Command: `go test ./test/... -run TestPatchMessageSubjectEditRefreshesSearchIndex -v` (also ran the full set of adjacent PatchMessage/search/edit-review tests - all pass, no regressions)
- Proves fail-before (assertion fails against pre-fix code, see Evidence above) / pass-after (passes with the fix applied).

## Discourse

https://discourse.ilovefreegle.org/t/9954/3

### DIFF
--- a/iznik-server-go/message/message.go
+++ b/iznik-server-go/message/message.go
@@ -2690,6 +2690,14 @@ func handleRevertEdits(c *fiber.Ctx, myid uint64, req PostMessageRequest) error 
 		}
 		args = append(args, req.ID)
 		db.Exec("UPDATE messages SET "+strings.Join(clauses, ", ")+" WHERE id = ?", args...)
+
+		// The rejected edit's subject already reached messages_index (applyPatchMessageCore
+		// refreshes it live, before a mod ever reviews it) - restoring the old subject here
+		// must refresh it again, or the index is left showing the reverted wording instead
+		// of the original (same gap as Discourse 9954/3, on the revert path).
+		if old.Oldsubject != nil {
+			reindexMessageSearchWords(db, req.ID)
+		}
 	} else {
 		// No recorded old values — just clear the editedby flag.
 		db.Exec("UPDATE messages SET editedby = NULL WHERE id = ?", req.ID)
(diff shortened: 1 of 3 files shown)


edwh:
Closing as a duplicate of #1164, which is the combined fix: it invalidates BOTH the keyword index (messages_index) and the vector embedding (messages_embeddings) on edit — #1165 only refreshed messages_index — and now also covers the revert-edit path that #1165 added. Folded that path into #1164.

PR #1166, opened 2026-07-23
fix(modtools): re-index message when subject is edited

edwh:
## Root Cause
`MessageSearchService::indexUnindexedMessages()` (iznik-batch) only indexes a message's subject **once**: it selects recently-approved messages `whereNotIn` `messages_index` msgid, indexes them, and never revisits a msgid that already has any `messages_index` row. When a message's subject is later edited, `iznik-server-go`'s `applyPatchMessageCore` commits the new subject to `messages.subject` immediately but never touches `messages_index` — so the stale, pre-edit words stay searchable forever and any term only present in the new subject can never be found by ModTools keyword search.

Confirmed against the reported message (id 121085393, live/Approved on Braintree and rippled into 6 other groups): `messages_edits` shows its subject was edited from `"WANTED: Spindle/stem/drive shaft (Bocking CM7)"` to `"WANTED: Spindle/stem for Moulinex food processor (Bocking CM7)"`. `messages_index` still only contains the words from the **original** subject (`drive`, `stem`, `spindle`, `shaft`) — `moulinex`, `food` and `processor` were never indexed, so searching "Moulinex" in ModTools finds nothing, even though the message is live and approved.

## Evidence
Failing test output (before fix):
```
=== RUN   TestPatchMessageSubjectEditInvalidatesSearchIndex
    message_test.go:7738:
        Error Trace: /src/test/message_test.go:7738
        Error:       Not equal:
                     expected: int(0)
                     actual  : int64(1)
        Test:        TestPatchMessageSubjectEditInvalidatesSearchIndex
        Messages:    editing the subject should clear the stale search index rows so the batch job rebuilds them from the new subject
--- FAIL: TestPatchMessageSubjectEditInvalidatesSearchIndex (0.17s)
FAIL
```

## Fix
In `applyPatchMessageCore` (`iznik-server-go/message/message.go`), when `subjectChanged` is true, delete the message's existing `messages_index` rows. This makes the message look "unindexed" again from `indexUnindexedMessages()`'s point of view, so the next batch run rebuilds the index from the current subject. Placed alongside the existing `contentcheck_checked_at` reset for the same `subjectChanged` condition (same pattern, same call site), so it fires for both self-edits and moderator edits, review-required or not.

## Other Call Sites
`grep -n "UPDATE messages SET subject" iznik-server-go` finds 3 other sites, all already covered:
- Two in `applyPatchMessageCore` itself (subject reconstructed from type/item/location) — these run *before* the `subjectChanged` diff, so they're already caught.
- `handleApproveEdits` — re-applies a subject that was already committed (and already invalidated) at edit-submission time; no separate change needed.
- One in the draft-creation path (`newMsgID`) — a brand new message has no existing index to invalidate.

## Test
- File: `iznik-server-go/test/message_test.go`, `TestPatchMessageSubjectEditInvalidatesSearchIndex`
- Command: `go test ./test/... -run TestPatchMessageSubjectEditInvalidatesSearchIndex -v`
- Proves fail-before (stale index row survives a subject-changing PATCH) / pass-after (row is cleared) the fix in `message.go`. Also re-ran 24 adjacent PATCH/edit tests (content-check reset, edit review flow, edit records) — all pass.

## Discourse
https://discourse.ilovefreegle.org/t/9954/5

### DIFF
--- a/iznik-server-go/message/message.go
+++ b/iznik-server-go/message/message.go
@@ -3457,6 +3457,18 @@ func applyPatchMessageCore(c *fiber.Ctx, myid uint64, req patchMessageRequest) e
 		db.Exec("UPDATE messages_groups SET contentcheck_checked_at = NULL, contentcheck_reasons = NULL WHERE msgid = ?", req.ID)
 	}
 
+	// The keyword search index (messages_index) is built once from the message's
+	// subject by the batch reindexer (MessageSearchService::indexUnindexedMessages(),
+	// iznik-batch), which skips a msgid entirely once it has ANY messages_index rows
+	// - so a subject edit otherwise leaves the OLD subject's words searchable forever
+	// while a term only present in the NEW subject (e.g. a brand name added after the
+	// original post) can never be found by search (Discourse 9954/5). Deleting the
+	// stale rows here makes the message look "unindexed" again so the next batch run
+	// rebuilds it from the current subject.
+	if subjectChanged {
+		db.Exec("DELETE FROM messages_index WHERE msgid = ?", req.ID)
+	}
+
 	if (subjectChanged || textChanged || typeChanged || locationChanged || itemsChanged || imagesChanged) && !isMod {
 		// Store oldtype/newtype only when type actually changed.
 		var oldType, newType interface{}
--- a/iznik-server-go/test/message_test.go
+++ b/iznik-server-go/test/message_test.go
@@ -7689,6 +7689,55 @@ func TestPatchMessageNonContentEditKeepsContentCheck(t *testing.T) {
 	assert.True(t, checkedAtStillSet, "a non-content edit should not discard the existing content-check stamp")
 }
 
+// TestPatchMessageSubjectEditInvalidatesSearchIndex (Discourse 9954/5): the keyword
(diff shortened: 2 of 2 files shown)


edwh:
Closing as a duplicate of #1164.

This PR does only the `messages_index` delete on subject change. #1164 already does that, as part of the complete fix for 9954:
- drops the `messages_embeddings` row (vector search) on a subject **or** body change;
- evicts the stale entry from apiv2's in-process embedding store (`embedding.Global.Evict`), so vector search stops matching the old wording even between the 2-min store refreshes (the gap the adversarial review caught);
- scopes keyword-index invalidation to subject changes, since `messages_index` is subject-derived;
- covers the approve-edits and revert-edits paths too.

#1164 is green. Post 5's keyword-miss ("Moulinex") is the same root cause as post 3 and is fixed there.

On post 5's other symptom — a mod scrolling her groups and not finding 121085393 — the earlier diagnosis found that message is live and appears in the ModTools listing of its own group (ListMessagesMT is not reach-gated for a group's mod). A scroll-miss is consistent with the post being on a group the mod wasn't scrolling, or buried by arrival order, rather than a listing bug. If that proves wrong it's a separate listing issue, not this reindex fix.

PR #1169, opened 2026-07-24
fix(TrashNothing join request sync): don't swallow the membership INSERT error in the partner join path

edwh:
## Root Cause
`putMembershipsPartner` (`iznik-server-go/membership/membership.go`), the handler behind `PUT /apiv2/memberships?partner=...` used by the TrashNothing (TN) integration, discarded the error from the `INSERT INTO memberships` call and unconditionally returned `{"ret":0,"status":"Success"}` to the partner. If that INSERT failed for any reason, no membership row was created, so the join never appeared in any moderator-facing queue and TN was told it had succeeded regardless - the request vanished with no trace and no way for either side to know something went wrong.

Production data confirms this actually happens: several TN-created accounts (`tnuserid` set, only ever set by this same handler's `CreatePartnerUser`) have **zero** rows in `memberships` despite repeated join attempts recorded in `memberships_history` for multiple groups each. This matches the report exactly - the requester's join request never reached a moderator.

## Evidence
Test run against the **unfixed** handler (kept the new `InsertPartnerMembership` seam but restored the original "ignore the error" behaviour) - `go test ./test/... -run TestPutMembershipsPartnerInsertFailureNotSwallowed -v`:
```
=== RUN   TestPutMembershipsPartnerInsertFailureNotSwallowed
    membership_test.go:3879:
        Error Trace: /app/test/membership_test.go:3879
        Error:       Should not be: 200
        Test:        TestPutMembershipsPartnerInsertFailureNotSwallowed
        Messages:    a failed membership insert must not be reported as Success to the partner
--- FAIL: TestPutMembershipsPartnerInsertFailureNotSwallowed (0.48s)
FAIL
```

## Fix
- Extracted the `memberships` INSERT into an exported package-level var `InsertPartnerMembership` (a seam so it can be stubbed in tests without breaking a real DB connection).
- `putMembershipsPartner` now checks its error and returns HTTP 500 (instead of a false "Success") when the insert fails, so the partner can see the failure and retry rather than the request disappearing silently.
- Also checks (and logs) the error on the companion `memberships_history` insert, matching the error-handling style already used elsewhere in this file (e.g. the `Hold`/`Approve`/`Reject` actions a few lines up).

## Other Call Sites
Same "insert result is captured/ignored but the handler still unconditionally returns Success" pattern exists at two other spots in the same file - left alone here to keep this fix scoped to the reported partner-join bug:
- `addMemberToGroup` (JWT self-join path, `membership.go`): checks `result.RowsAffected` to decide whether to write history/log, but returns `"Success"` regardless of whether the initial INSERT itself errored.
- `deleteMembershipsPartner` (partner leave path, `membership.go`): same shape on the `DELETE FROM memberships` call - lower impact since it only affects leaving a group, not the reported joining bug.

## Test
- File: `iznik-server-go/test/membership_test.go`, `TestPutMembershipsPartnerInsertFailureNotSwallowed`.
- Command: `go test ./test/... -run TestPutMembershipsPartnerInsertFailureNotSwallowed -v`.
- Proves fail-before (buggy code returns 200 "Success" even though the insert failed) / pass-after (fixed code returns a non-200 error and leaves no membership row) using the new `InsertPartnerMembership` seam to simulate the DB failure.
- Also re-ran the full `TestPutMembershipsPartner*`/`TestDeleteMembershipsPartner*` set to check for regressions in the surrounding code; all pass except a pre-existing, unrelated flake (`TestDeleteMembershipsPartnerUnsubscribe`, a hardcoded-`tnuserid` collision in the shared test DB) reproduced in complete isolation on unmodified `master` - unrelated to this change.

## Discourse
https://discourse.ilovefreegle.org/t/9961

### DIFF
--- a/iznik-server-go/membership/membership.go
+++ b/iznik-server-go/membership/membership.go
@@ -1215,20 +1215,36 @@ func putMembershipsPartner(c *fiber.Ctx, db *gorm.DB, partnerKey string) error {
 		return c.JSON(fiber.Map{"ret": 0, "status": "Success", "fduserid": userid, "addedto": utils.COLLECTION_APPROVED})
 	}
 
-	// Insert membership.
-	db.Exec("INSERT INTO memberships (userid, groupid, role, collection) VALUES (?, ?, ?, ?)",
-		userid, groupid, utils.ROLE_MEMBER, utils.COLLECTION_APPROVED)
+	// Insert membership. This must not be swallowed: if it fails, the partner
+	// (e.g. Trash Nothing) gets told the join succeeded while no membership
+	// row - and nothing else downstream - was ever created, so the request
+	// silently never reaches a moderator (Discourse #9961).
+	if err := InsertPartnerMembership(db, userid, groupid); err != nil {
+		stdlog.Printf("Failed to insert partner membership user %d group %d: %v", userid, groupid, err)
+		return fiber.NewError(fiber.StatusInternalServerError, "Failed to join group")
+	}
 
 	// Record in memberships_history with processingrequired=1 so the
 	// Laravel batch (memberships:process) sends the group welcome email,
 	// runs spam checks, and applies review flags. Without this row the
 	// cron has nothing to do and welcomes are silently dropped.
-	db.Exec("INSERT INTO memberships_history (userid, groupid, collection, processingrequired) VALUES (?, ?, ?, 1)",
-		userid, groupid, utils.COLLECTION_APPROVED)
+	if result := db.Exec("INSERT INTO memberships_history (userid, groupid, collection, processingrequired) VALUES (?, ?, ?, 1)",
+		userid, groupid, utils.COLLECTION_APPROVED); result.Error != nil {
+		stdlog.Printf("Failed to insert partner memberships_history user %d group %d: %v", userid, groupid, result.Error)
+	}
 
 	logMembershipAction(log.LOG_TYPE_GROUP, log.LOG_SUBTYPE_JOINED, groupid, userid, userid, "via partner")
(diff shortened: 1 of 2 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #1169) — this fix did not pass and is not ready to merge.

- **Partial implementation / root cause not addressed**: The second INSERT (memberships_history) is still effectively swallowed: on error it only calls stdlog.Printf and then falls through to `return c.JSON(... "status": "Success" ...)`. The code's own comment two lines above says "Without this row the cron has nothing to do and welcomes are silently drop
- **Partial implementation / non-atomic write leaves the exact broken state the PR cites as evidence**: The handler auto-creates the user (tnuserid set) earlier in the function, before this hunk. When InsertPartnerMembership now returns an error the handler returns 500 and returns immediately — with no transaction, no rollback, and no cleanup of the just-created user. The result is a user row with tnu
- **Test proves nothing (cannot fail before the fix)**: TestPutMembershipsPartnerInsertFailureNotSwallowed depends on `membership.InsertPartnerMembership`, a package-level var introduced by this same diff. Against the pre-fix code the test does not compile, let alone fail, so it is not a reproduction test — it is a post-hoc test of newly added error-hand
- **Assertions are not meaningful**: Two problems in the same test. (1) `assert.NotEqual(t, 200, resp.StatusCode)` passes for ANY non-200 — 400, 403 for a bad partner key, 404 for a bad groupid, a routing 405. If insertTestPartnerKey/domain matching or the query-string setup ever drifts and the request 403s before reaching the INSERT a

PR #1171, opened 2026-07-24
fix(membership): banned join is a real failure in both PUT paths, not fake Success

edwh:
## Root cause
`PUT /memberships` forks into `putMembershipsPartner` (TrashNothing and other partners) and `addMemberToGroup` (website self-join + moderator "Add member"). Both reported a banned member's join as `ret=0 "Success"` while writing nothing — no membership, no history, no log — so the join silently vanished with nothing for a moderator to find, while the caller was told it worked (Discourse #9961: TrashNothing join requests never arrived).

## Fix
**Server** — both branches now return `403 "Failed - banned"` for a banned member (V1 parity: `User::addMembership` returns FALSE for `isBanned()`), instead of a fake Success.

**Client** — the 403 is surfaced appropriately rather than thrown uncaught + logged to Sentry:
- `notABannedFailure` / `isBannedFailure` helpers (mirroring `heldConflict`) keep the expected 403 out of Sentry and let callers react.
- **Self-join** (`auth.joinGroup`, the group Join button): swallowed silently — we don't reveal the ban to the member, we just don't join.
- **Moderator "Add member"** (`ModAddMemberModal`): surfaced — the modal shows "That person is banned… unban them first" so the mod knows why nothing happened.

## Tests
- Go: `TestPutMembershipsPartnerBannedNotFakeSuccess` (partner path); the two website-path banned tests now assert the 403 failure (no Approved membership, `users_banned` row intact).
- Vitest: `bannedFailure.spec.js` covers the Sentry-suppression predicate and the caught-error detector.

## Discourse
https://discourse.ilovefreegle.org/t/9961

### DIFF
--- a/iznik-nuxt3/api/MembershipsAPI.js
+++ b/iznik-nuxt3/api/MembershipsAPI.js
@@ -1,5 +1,6 @@
 import BaseAPI from '@/api/BaseAPI'
 import { notAHeldConflict } from '~/api/heldConflict'
+import { notABannedFailure } from '~/api/bannedFailure'
 
 export default class MembershipsAPI extends BaseAPI {
   update(data) {
@@ -7,7 +8,9 @@ export default class MembershipsAPI extends BaseAPI {
   }
 
   joinGroup(data) {
-    return this.$putv2('/memberships', data)
+    // A banned member's join is refused with 403 "Failed - banned" - expected, so
+    // don't log it to Sentry (the caller handles it).
+    return this.$putv2('/memberships', data, notABannedFailure)
   }
 
   leaveGroup(data) {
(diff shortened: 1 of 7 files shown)


edwh:
Auto-closed by the monitor's adversarial review (PR #1171) — this fix did not pass and is not ready to merge.

- **Partial implementation**: The identical fake-success is left in place in the sibling branch of the SAME endpoint. `PutMemberships` (membership.go:1118) forks into `putMembershipsPartner` (fixed here) and `addMemberToGroup` (membership.go:1249-1253), and the latter still does `if bannedCount > 0 { return c.JSON(fiber.Map{"ret
- **Contract break / claim does not match diff**: The comment claims V1 parity with `ret=4 "Failed - likely ban"`, but the code returns `fiber.NewError(fiber.StatusForbidden, ...)`, which main.go:47-66 renders as HTTP 403 `{"error":403,"message":"Failed - banned"}` - there is no `ret` field at all, and V1 returned HTTP 200 with a ret code. The repo

PR #1185, opened 2026-07-27
fix(mjml): pin mrml to =6.0.1 so mj-raw content survives compilation

edwh:
## Problem
`master` is red (pipeline #9853, commit `d6981931a`): `Tests\Unit\Mail\TrackingPixelOutlookTest` fails with "Tracking pixel `` must survive MJML compilation — 0 imgs". The prior master commit (`6edc3bb12`) was green.

## Root cause
The batch image builds the mrml PHP extension (wrapper pinned to `v1.2.1`) but sets the underlying **mrml Rust crate to `version = "6"` and `rm -f Cargo.lock`** (`iznik-batch/docker/Dockerfile`), so every build resolves whatever the latest 6.x is at that moment. `mj-raw` rendering changed across 6.x: **6.0.1 emits `<mj-raw>` content verbatim, a later 6.x drops it.** CI's cached build layer baked a version that drops it; a freshly built local image (mrml **6.0.1**) renders the pixel and passes the exact test. `d6981931a` (tracking pixel via `mj-raw`) surfaced this latent non-determinism.

Verified directly against the running batch image (mrml 6.0.1):
```
(new \Mjml\Mjml())->render('<mjml><mj-body><mj-section><mj-column><mj-raw><img .../></mj-raw></mj-column></mj-section></mj-body></mjml>')
```
emits the `` verbatim and matches the test's `#]*URL[^>]*>#` regex.

## Fix
Pin the crate to exactly **`=6.0.1`** (proven-working) so the build is reproducible and `mj-raw` keeps rendering. Editing the RUN line also invalidates the cached cargo layer, so CI rebuilds against 6.0.1 and re-runs the full email suite.

## Note
This unblocks the two open PRs (#1183 rspamd, #1184 spatial-remap), which branch off this red master and inherit the failure. Master's Playwright failure appears independent (this commit only touches the mrml pin) — this PR's CI run will show whether it re-passes.

https://claude.ai/code/session_012RoB9qVCxwKjFmQzB7qSF2

### DIFF
--- a/iznik-batch/docker/Dockerfile
+++ b/iznik-batch/docker/Dockerfile
@@ -9,6 +9,14 @@
 # inlining onto elements is required for Gmail, so we bump to mrml 6 and
 # enable css-inline. mrml 5->6 needs no source changes to the extension.
 #
+# Pin mrml to EXACTLY 6.0.1 (not "6" = ^6). With "6" + `rm Cargo.lock`, cargo
+# resolved whatever the latest 6.x was at build time, so the baked extension
+# drifted between builds. mj-raw rendering changed across 6.x: 6.0.1 emits
+# <mj-raw> content (e.g. the tracking pixel <img>) verbatim, but a later 6.x
+# dropped it, which turned CI's cached build red on TrackingPixelOutlookTest
+# while a freshly-built local image (6.0.1) passed. Exact-pinning makes the
+# build reproducible and keeps mj-raw working.
+#
 # Built here (per-branch) rather than in Dockerfile.base because the base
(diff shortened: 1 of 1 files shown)


edwh:
Closing: the revert of d6981931a (#1186) removed the mj-raw tracking pixel this pin was supporting, so pinning mrml to =6.0.1 is now orphaned. The proper mrml pin (plus a committed Cargo.lock for full determinism) should be folded into the future re-land of the mj-raw Outlook-scrollbar fix, once the ci/batch image cache is refreshed so the pin can actually build in CI. Rationale captured in #1186's description.

PR #1195, opened 2026-07-29
fix(rippling): stop /rippling/analytics 504ing on the sysadmin KPIs

edwh:
## Problem

The sysadmin rippling analytics tab was 504ing at the gateway on `GET /apiv2/rippling/analytics` (surfacing client-side as a CORS error, since the gateway's 504 carries no `Access-Control-Allow-Origin` header).

## Root cause

`rippling_reach` has no index on `created_at`, so every analytics query full-scanned the table. It is only ~52k rows but **~17GB**, because each row carries the reach polygons — so the scan costs ~15s (`EXPLAIN` showed `type: ALL`). The endpoint runs four such queries serially: ~61s total, just past the gateway timeout.

## Fix

1. **Covering index** `rippling_reach (created_at, total_freeglers)` (msgid rides along as the PK), so the window predicate is answered index-only without touching a polygon page. Measured on prod: base scan ~15s → ~1.1s, serial total ~61s → ~14s.
   - **Already applied to prod** (2026-07-29) node-by-node under `wsrep_OSU_method=RSU`; all three nodes back to Synced. The table's SPATIAL index makes MySQL refuse `LOCK=NONE`, so the build used `LOCK=SHARED` (2–14s per node). The Laravel migration is guarded on `information_schema` so it no-ops on prod and creates the index on dev/test DBs.
2. **Concurrent queries** in the Go handler: the four independent blocks in `Analytics` (counts, held replies, trend, section 3) now run in parallel, as do the two queries inside `rippledOutSection`, so wall time is the slowest single query (~7s, the rescue floor) rather than the sum.

The 504 is already resolved in production by the index alone; the Go change brings the response down to ~7s and needs the normal deploy after merge.

### DIFF
--- a/iznik-batch/database/migrations/2026_07_29_000001_add_rippling_reach_created_index.php
+++ b/iznik-batch/database/migrations/2026_07_29_000001_add_rippling_reach_created_index.php
@@ -1,0 +1,52 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\DB;
+
+/**
+ * rippling_reach(created_at, total_freeglers) index for the ModTools sysadmin rippling
+ * analytics tab (iznik-server-go/rippling/analytics.go). Every analytics query anchors
+ * on "posts rippled in the [start,end) window, split by density stratum", i.e. a range
+ * on created_at filtered by total_freeglers.
+ *
+ * Without this index each query full-scans rippling_reach. The table is only ~52k rows
+ * but ~17GB, because each row carries the reach polygons, so the scan costs ~15s — and
+ * the endpoint runs four such queries serially, tipping /rippling/analytics past the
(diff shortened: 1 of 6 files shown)


edwh:
Interim mitigation applied to prod while this awaits merge: the LB (applb HAProxy) global `timeout server` is 50s, and the serial endpoint still runs ~30-60s depending on node/window even with the new index (db3 — the active API backend — measured ~31s for a 90-day window). Added a per-path override in `api_server_backend`: `http-request set-timeout server 180s if { path_beg /apiv2/rippling/analytics }` (config backed up as haproxy.cfg.bak-rippling-analytics-20260729-153331). Verified end-to-end: the exact failing request now returns 200 in ~30s. The override is harmless to keep after this PR deploys (endpoint drops to ~20s), but can be removed.

PR #1213, opened 2026-07-31
fix(mail): don't geolocate private or reserved sender IPs

edwh:
## Problem

The incoming-mail spam check decided whether a sender IP was "internal" with a single prefix test:

```php
// Skip internal IPs
if (str_starts_with($ip, '10.')) {
```

That recognises only `10.0.0.0/8`. A mail server's own internal relay hop arriving in a `Received:` header — `172.20.0.1`, which is RFC1918 but sits in `172.16.0.0/12` — was therefore treated as the sender's real address and fed to the country blocklist and the GeoIP lookup. A legitimate email post could be judged by an address that was never the sender's.

The same address also reached `messages.fromcountry` through `createGroupPostMessage`, so ModTools' "posted from outside the UK" flag could be derived from a Docker gateway.

Reported as Discourse topic 9865.

## Change

`SpamCheckService::isPrivateOrReservedIp()` replaces the prefix test, using `filter_var` with `FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE`. That covers RFC1918 in full (`10/8`, `172.16/12`, `192.168/16`) plus loopback, link-local, and the other reserved blocks, rather than enumerating ranges by hand.

`createGroupPostMessage` applies the same check before calling `lookupIPCountryCode`, storing `NULL` when the only available address is private or reserved.

## Tests

`SpamCheckServiceTest` gains a data-provider regression test asserting such addresses skip the IP checks entirely, covering `172.20.0.1` (the reported address), `192.168.1.1` and `127.0.0.1`, with a country blocklist loaded so the assertion would fail if the address still reached the GeoIP path. `IncomingMailServiceTest` covers `fromcountry` being left NULL for a private sender IP.

The `Incoming` Laravel suite passes.

### DIFF
--- a/iznik-batch/app/Services/Mail/Incoming/IncomingMailService.php
+++ b/iznik-batch/app/Services/Mail/Incoming/IncomingMailService.php
@@ -2850,7 +2850,13 @@ class IncomingMailService
                 // Geolocate the sender IP so ModTools can flag posts from
                 // outside the UK (MessageHistory.vue). V1 (Message.php/Spam.php)
                 // stored the ISO code here; store NULL when it can't be resolved.
-                'fromcountry' => $email->senderIp ? $this->spamCheck->lookupIPCountryCode($email->senderIp) : null,
+                // Private/reserved IPs (e.g. an internal relay hop picked up
+                // from a Received: header) are never geolocated - see
+                // Discourse topic 9865, where 172.20.0.1 (RFC1918) was the
+                // only IP in the headers of a legitimate email post.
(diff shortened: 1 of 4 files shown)


edwh:
Closing: misdiagnosis. Topic 9865 was fixed on master by 91d15838 ("also denormalise lat/lng on the edit path"); the private/reserved-IP change here addressed a different, unreported thing and only cited 9865 in passing. Recovered from an abandoned worktree without checking what actually fixed 9865.

PR #1215, opened 2026-07-31
fix(rippling): distinguish never-reviewed posting status from a moderator's decision

edwh:
## Problem

Rippling auto-joins a poster to every group their post reaches (`ExpandService::addPosterMembershipToRippledGroups`) without setting `ourPostingStatus`, so the membership row keeps it NULL.

The Go API resolves a NULL `ourPostingStatus` to the literal `MODERATED` for ModTools display. That is indistinguishable from a moderator having explicitly chosen "Moderated", so every rippled-in arrival shows as an actively moderated member — even though the post that just arrived went live with no moderator involvement at all, having already been vetted on its origin group.

That produces the contradiction reported: a member "not approved by anyone but has come through as a moderated member".

## Change

The membership API response gains `OurPostingStatusReviewed`, true only when a moderator has actually stored a status. `ModModeration.vue` shows a short note when it is false, so the dropdown no longer reads as an active decision nobody made.

The NULL → `MODERATED` resolution itself is unchanged, so nothing that depends on the effective status shifts; this only distinguishes "never reviewed" from "reviewed and set to Moderated" in what moderators see.

## Tests

`user_test.go` covers the new field being false for a membership with a NULL status and true once a status is stored. `ModModeration.spec.js` covers the note appearing only in the never-reviewed case.

Full Go and vitest suites pass.

### DIFF
--- a/iznik-nuxt3/modtools/components/ModModeration.vue
+++ b/iznik-nuxt3/modtools/components/ModModeration.vue
@@ -1,11 +1,14 @@
 <template>
-  <div class="d-inline d-flex justify-content-start flex-wrap">
+  <div class="d-inline d-flex justify-content-start flex-wrap align-items-center">
     <b-form-select
       v-model="postingStatus"
       :options="options"
       class="sel"
       :size="size"
     />
+    <small v-if="!membership.ourpostingstatusreviewed" class="text-muted ms-1">
+      Not yet reviewed by a moderator - this is just the default
+    </small>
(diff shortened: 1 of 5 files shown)


edwh:
Closing: red herring. The NULL ourPostingStatus is not what makes a rippled-in arrival look moderated, so surfacing a never-reviewed flag would dress up a symptom rather than fix a cause.

PR #1238, opened 2026-08-02
fix(moderation): finish the per-group hold migration; compat field reports a hold only when the viewer's groups agree

edwh:
## What this PR now is

Originally a two-file change to `effectiveHeldby()`. Auditing the heldby migration plan
against master showed most of that plan already landed via #1199 and its follow-ups, so this
PR was extended to close everything that genuinely remained in the messages-hold domain.
`plans/messages-heldby-obsolescence.md` is rewritten in this PR to record the audited state.

## 1. The compat field reported another group's hold (Discourse 9481/642)

`Message.heldby` exists only for bundled ModTools app builds — removing it broke every
installed app once (9481/636), and `TestMessagePayloadKeepsMessageLevelHeldbyForBundledApps`
pins it. `effectiveHeldby()` resolved it as "held if ANY of the viewer's moderated groups
holds the message", so a mod covering two groups — held on one, not the other — saw the hold
reported against the copy they were actually reviewing. A message-wide yes/no field cannot
name which group it means.

Changed to AND across the viewer's own groups: report the holder only when every one of the
viewer's groups on the message agrees; nil on disagreement. nil is safe where "held" is not:
server-side enforcement (`dispatchPostMessageAction` → `heldByAnotherMod`) re-derives the
truth from `messages_groups.heldby` per-group and answers 409 to any blocked action, so a
compat client that acts on an invisible hold is refused with an explanation — while the old
false "held" hid every button on an unheld copy with no recovery. The single-group case (the
overwhelming majority) is behaviourally identical, which is why the bundled-app pinning test
passes unchanged.

## 2. The last mirror reader: microvolunteering invitations ignored holds

`MicrovolunteeringNotifyService` filtered candidates on `messages.heldby IS NULL`. Nothing
writes that column any more, so the predicate was unconditionally true: members were invited
to "please review" copies a moderator had held. Scoped to `messages_groups.heldby` — the
group id is already in the same query row.

The regression tests use separate reviewers per group deliberately: with one shared
reviewer, per-run dedup masks the bug (the reviewer gets exactly one notification whether or
not the held row enters the candidate set). With a reviewer whose only route to the message
is the held group, the buggy path visibly notifies them.

Both test files also stop seeding the dead `messages.heldby` column in their fixtures.

## 3. Missing regression test for ChaseUpService

Its per-group fix landed in `a6b52512f` with no test exercising the new filter (none of the
three chase-up test files mention holds). Added: a post cross-posted to two groups, held on
one — `notifyLanguishing()` must still count the unheld copy.

## 4. Missing store test for hold/release refresh

Nothing proved `hold()`/`release()` re-fetch the message and replace `groups[].heldby` — the
mechanism that keeps the Hold/Release toggle in sync without a reload. Recovered the spec
from closed PR #1072 and reshaped it to the current mock conventions.

## Deliberately not done here

- The dead `heldby` in `message.go`'s single-message raw SELECT (its value is discarded and
  recomputed) must be removed before the column can ever be dropped, but #1230 rewrites
  those exact queries into GORM builders — recorded as a Phase 5 precondition in the plan
  instead of creating a guaranteed conflict for housekeeping.
- Phase 3 (removing the payload field) stays blocked until the installed-app floor moves;
  the plan now says so explicitly, with the pinning test named.

## Test plan

- Go suite locally against a rebuilt apiv2: 3829 pass, 0 fail, 0 build failures. TestEffectiveHeldby_AmbiguousAcrossViewersOwnGroupsReturnsNil and TestMessagePayloadKeepsMessageLevelHeldbyForBundledApps both pass by name.
- Laravel suite locally: 5114 pass, 0 errors, 0 failures. The cross-group microvolunteering tests and the chase-up regression test all run in this suite.
- Store spec runs in CI (vitest has no local runner in this checkout).
- `scripts/check-docs-freshness.mjs`: OK, 44 pages — no covered path changed.


edwh:
Review: this trades a false positive for a false negative, and the code comment claims it does not.

The change makes `effectiveHeldby` return `nil` when the viewer's own groups disagree — held on one, unheld on another. The comment justifies that with:

> nil is exactly what an unheld post already reports, so this never invents a false negative - only avoids a false positive that a compat client cannot untangle.

That is not right. Consider the case the diff itself describes: a moderator covering group 10 (unheld) and group 20 (held). The post **is** genuinely held on group 20, and that moderator moderates group 20. Reporting `nil` tells them nothing is held when something is. For a mod working group 20, that is a false negative, not the absence of a false positive.

So the change does not remove an incorrect answer; it swaps which half of the population gets one. Before, mods of the unheld group saw a spurious hold. After, mods of the held group see no hold. Both are wrong, and the second is arguably worse: a spurious hold prompts someone to look, whereas a missing hold means a held post sits unreviewed with nothing to prompt anyone.

The underlying problem is stated accurately in the diff and is worth acting on instead: a message-wide yes/no field cannot express a per-group condition. No value of `effectiveHeldby` is correct for a multi-group moderator, because the question "is this held?" has different answers for different groups the same person moderates.

That suggests the compat field should be deprecated in favour of the per-group rows rather than patched again. `heldby.go` has now been changed three times in thirty days — every change it has ever had — each time to make a message-wide flag better approximate per-group truth:

- `c6899ef82` show a hold only to mods of the group it's on, not every rippled group
- `a6b52512f` make a hold belong to the group it was placed on
- `6a0290646` keep message-level heldby so bundled apps still show holds

This would be the fourth. The pattern suggests the shim, not the arithmetic, is what needs the decision.

Not closing this — the bug report behind it (Discourse 9481/642) is real and the analysis of the multi-group case is good. But it should not merge in this form.

PR #1266, opened 2026-08-05
Laravel ORM conversions: simple JSON predicates converted, keep-raw triage under active refutation

edwh:
Started as a fix for parity proofs the #1230 merge left stale, then took the ratchet out entirely now the Go migration has landed.

## 1. Re-anchor the stale parity proofs

PR #1257 (ripple_join) changed four statements #1230 had already converted. The merge updated **production** correctly and fixed one of the four declarations — but left the other three describing the pre-#1257 statements, and edited nothing in their parity tests.

That combination is silent: the test builds the OLD statement, the OLD statement still matches the OLD golden, so it passes. Nothing in `ci-ratchet.sh` could catch it — comparing a converted site's golden against what production renders today was precisely the parity tests' job, and that was the thing left behind.

| Site | What #1257 changed | Was |
|---|---|---|
| `848af7d73bfe` | `wasHome` gained `AND mem.rippled = 0` | golden + test pre-#1257 |
| `db03a274a8a2` | INSERT gained `was_ripple_join` | golden + test pre-#1257 |
| `f382f9bfe80b` | now calls `EstablishedOriginMemberExists()` | golden + all 4 shapes pre-#1257 |
| `9894f2a0d95d` | converted from raw in the merge | **no entry, no golden, no test at all** |

The last one never passed through the extractor while still raw, so it minted nothing and was invisible to every gate. Also retires a phantom Laravel site (`27496bd44361`) that renumbered and was left inflating that ceiling by one.

Verified by rendering all five assertions under `renderDryRun` (5/5 pass), plus a **negative control**: the pre-#1257 build functions now *fail* against the re-anchored goldens, which is what proves the goldens actually moved.

## 2. Retire the ratchet, add the hook

The Go conversions are done — **1,593 converted** — so the machinery that carried them out is overhead.

To be precise, because the manifest's wording invites the wrong reading: Go did **not** reach zero raw SQL. `raw: 0` was a *status* count meaning "nothing left untriaged". **48 raw call sites remain in Go production code** — index hints, runtime-varying identifiers, statements GORM will not render — each recorded as `keep-raw` with a written reason. `ci-ratchet.sh` regenerated a **5,974-entry** inventory on every build and provisioned Go and PHP toolchains on a host that needs neither, all to answer one question that does still matter: *did someone add raw SQL back?*

`.claude/check-raw-sql.sh` answers it at authoring time, and covers **both** stacks:

- **iznik-server-go** — converted bar the deliberate 48, so anything *new* is a regression. `gorm.Expr`/`clause.Expr` stay allowed (fragments in a real chain, not statements). It will fire if you edit one of the existing 48, whose justifications lived in the inventory rather than inline — accepted rather than pre-empted, since it reads diffs and so only interrupts someone actually editing one of those queries.
- **iznik-batch** — **569 raw sites, 0 converted**. Not asking anyone to fix those: it judges only the text being written, so editing around existing raw SQL is unaffected. It stops the pile growing silently.

Migrations, tests and the EeeSqlite store are exempt. Escape hatch in both languages is `// keep-raw: <reason>` — deliberately the same written-justification rule retired gate (e) enforced, so that comment is now the audit trail. Tested against 11 cases covering both languages, every exemption and both escape paths.

**Removed — Go side only.** `ormharness/`, `ormshadow/` (Layer 3, never imported by production), the Go extractor and ratchet, the orb step, the docs page.

**The Laravel half stays.** Those conversions have *not started* — 569 raw, 0 converted — and `tests/Support/OrmHarness` is the Layer 1/2 harness built specifically to enable them. `ProvenSitesTest` is explicit: *"every site named here stays status `raw`… the production call sites are untouched."* Deleting it would have thrown away scaffolding for work still to come. That work gets its own PR. Deleted in all — including 58 `test/orm_*` parity files, 17 harness self-tests and 6 `*_tier9_test.go` parity files in package dirs. Orb republished as **1.1.373** and pinned.

**Kept** — 3 files renamed out of the `orm_` prefix, because they guard live behaviour rather than prove a conversion:
- `insertid_gorm_writeback_test.go` — id comes back off the write result, not a replica-routable SELECT (the 9832 class)
- `insertid_ondupe_zero_test.go` — GORM reads `RowsAffected` before `LastInsertId`, so the ON DUPLICATE KEY sites need `gorm.WithResult()`. A real-DB test: the parity harness only compared rendered text and passed either way
- `retrygorm_test.go` — `RetryGorm` preserves retry-on-transient

`main_test.go`'s `TestMain` called two helpers from a deleted file, which would have broken the entire Go test package — caught by cross-referencing all 22 deleted helpers against surviving files.

## 3. Strip the dead site-id markers

**1,572** marker comments across **106** production files named ids that no longer resolve. The **1,297** that were *only* a reference are gone; the **275** with real reasoning attached keep the reasoning and lose the dead prefix.

Verified mechanically rather than eyeballed: `git diff --unified=0` over `iznik-server-go` contains **zero** changed lines that are not comments, gofmt reports the same 42 pre-existing unformatted files before and after, and the packages type-check.

## On the coverage check

`coverage/coveralls` reports **-0.1% to 73.675%**, and it is worth reading before treating it as a regression in test quality: **every individual language held or improved** - go +0.02%, playwright +0.3%, laravel and the JS unit suite unchanged.

The aggregate moved because this PR deletes ~142k lines of `ormharness`/`ormshadow`, a body of test-support code with near-total coverage. Removing well-covered lines from both numerator and denominator shifts the blend toward the lower-covered suites (playwright sits at 22%), so the combined figure drops while no surviving code lost a single covered line.

Nothing is being skipped or made optional: the code those tests covered was deleted alongside them. The stacked Laravel PR (#1266) adds parity tests on the Laravel side, which moves the blend back the other way.

## Correction: "JSON at zero raw sites" was wrong

I claimed, in this description, a commit message and repeatedly in review,
that JSON had reached zero raw sites. It has not. `app/` still contains 44
non-comment `JSON_EXTRACT` occurrences.

The error was in my *verification*, not the conversions. I checked with
`grep "Raw(['\"]JSON_EXTRACT"`, which only matches when `JSON_EXTRACT` is the
first thing inside the raw string. Every wrapped or compound form slipped
past it:

| shape | count | convertible? |
|---|---|---|
| `JSON_UNQUOTE(JSON_EXTRACT(...))` | 8 | no - a function over a column |
| `COALESCE(JSON_EXTRACT(...))` | 6 | no |
| `CAST(JSON_EXTRACT(...))` | 6 | no |
| `(JSON_EXTRACT(...) IS NULL OR ...)` compound | 4 | yes - closure + `whereJsonDoesntContainKey` |
| inside whole raw statements | 20 | blocked for other reasons |

What is true is narrower and worth stating exactly: every *simple* JSON
comparison predicate - `= true`, `= 0`, `= 1`, `IS NULL` - is now a builder
call, and those conversions are proven. The wrapped forms are blocked by the
same boundary as spatial and column arithmetic: no builder method expresses a
function applied to a column.

This is the same failure the whole exercise has been about - testing one
pattern and generalising - except this time it produced a false claim of
success rather than a false claim of impossibility.

## The raw-SQL guard is carried here, because it did not land with #1262

#1262 merged, but master's `.claude/check-raw-sql.sh` still opens with
`if [ -n "$CI" ]; then exit 0; fi` and no CircleCI job runs it - so the guard
that was supposed to replace the retired ratchet is currently inert. The fix
was written on that branch after the merge point and never reached master.

Cherry-picked here (`869197f38`, `21ae26b6c`):

- `--diff BASE` mode that scans a branch's added lines and fails the build,
  sharing one path/pattern classifier with the authoring-time hook
- a `check-raw-sql` CircleCI job wired into all three test workflows
- `.claude/check-raw-sql.test.sh`, 21 cases across both modes against real git
  history. It found a real bug on its first run: the exemption patterns were
  written `*/iznik-batch/tests/*` with a leading slash, which matches the
  absolute paths the editor mode receives but not the relative paths git emits,
  so in `--diff` mode every exemption silently failed

Verified on this branch: 21/21 self-tests pass, and
`check-raw-sql.sh --diff origin/master` exits 0 against these ~60 conversions.

This matters more than the conversions. The inventory grew 473 -> 481 from a
single master merge while this branch was draining it: conversion is a one-off,
the gate is what stops the pile refilling.

## Correction: "spatial cannot be converted" was too strong

I verified the spatial claim by reflecting over 584 methods across
`Query\Builder`, `Eloquent\Builder`, `Schema\Builder` and `Blueprint`, and
reported it as the one keep-raw category that survived review. The reflection
was sound; the conclusion drawn from it was not. It established that **core
Laravel** has no spatial builder API - not that these sites cannot be converted.

Maintained packages provide exactly the operations these 37 sites use:

- [`MatanYadaev/laravel-eloquent-spatial`](https://github.com/MatanYadaev/laravel-eloquent-spatial/blob/master/API.md) - `whereDistanceSphere()`, `orderByDistanceSphere()`, `whereWithin()`
- [`jackardios/laravel-eloquent-spatial`](https://packagist.org/packages/jackardios/laravel-eloquent-spatial) - `whereContains()`, `withDistanceSphere()`
- [`grimzy/laravel-mysql-spatial`](https://github.com/grimzy/laravel-mysql-spatial/blob/master/src/Eloquent/SpatialTrait.php), [Laravel Magellan](https://laravel-news.com/laravel-magellan)

I have NOT adopted one. These are Eloquent-model-oriented and want spatial
casts on the models, so it is a dependency and data-modelling decision rather
than a mechanical conversion - it should be made deliberately, not slipped into
a conversion PR.

The accurate statement of the remaining boundary is narrower than three
categories: **core Laravel's builder expresses no function or arithmetic
applied to a column.** That single limitation covers the spatial sites, the
sampling arithmetic (`% ?`, `MOD()`, `CRC32()`) and the `COALESCE`/`CAST`/
`JSON_UNQUOTE`-wrapped JSON sites alike.

## CI

Master merged in for two harness fixes this branch depends on, neither of them this branch's own work:

- `dc6cebd51` rolls the seeded posts forward at load time. They carried absolute dates and had aged past the browse feed's 31-day window, which failed five Playwright feed tests on master and on every branch alike.
- `544c3267a` drives `messages:update-spatial-index` during Playwright runs. Batch's scheduler is disabled in CI, so a post created by a spec never reached `messages_spatial`. `GetMessagesForUser` takes an INNER JOIN on that table for the other-user case and a LEFT JOIN for your own, so the poster saw their post and every other viewer saw an empty profile — which is what the two `test-profile-page.spec.js` failures on this branch were.

The `messages_spatial` conversions here were checked directly against the originals and are not implicated: `ST_SRID(POINT(lng, lat), 3857)` and `ST_GeomFromText('POINT(lng lat)', 3857)` return the same geometry and the same SRID, and `upsert()` compiles to the same `INSERT ... ON DUPLICATE KEY UPDATE col = VALUES(col)`.

---

## Addendum: CI was red on this branch for a reason unrelated to the ORM work

The branch's Playwright job had been failing on one spec across three consecutive pipelines. It turned out to be a product bug, plus three defects in the local test harness that were sitting behind it.

### The CI failure: a batched user fetch that never settled

`useUserStore.fetch()` debounces ids into a batch and hands the caller a promise that `processBatch()` settles when the request returns. It only settled it on success — when `fetchMultiple()` threw, the loop calling `resolve()` was skipped, so every waiter in that batch stayed pending for ever and the rejection escaped the `setTimeout` unhandled.

A single unknown id is enough: the API returns 404 for `/user/<id>`. So `/profile/<unknown-id>` awaited a promise that never settled, the page's own 404 handling could never run, and the route rendered nothing at all — `<body>` stayed empty, which is what the spec was timing out on.

Reproduced as a unit test before fixing (`Test timed out in 30000ms`, plus two unhandled rejections), then fixed by settling every waiter when the request fails. Waiters whose id was in the failed request, with nothing cached, get the error; the rest resolve as before, so an id being fetched elsewhere is not rejected on another batch's behalf. `stores/message.js` already had the equivalent guard — the user store was the outlier.

### Three harness defects found while verifying that fix

**Fixture dates aged out of the feed window.** `scripts/test-fixtures.sql` carries fixed `2026-07-09` dates and the feed only shows posts from the last `MessageSpatialService::RECENT_DAYS` (31) days. `scripts/setup-test-database.sh` already rolls them forward for exactly this reason; the Playwright status endpoint loads the fixtures itself and never got the same treatment. Measured with the fixtures 32 days old:

```
rows after fixture reload:  71
one reconciler run:         removed_old: 71
rows after reconciler:       0
```

Locally the batch scheduler runs the reconciler, so `messages_spatial` empties within minutes of a reset and browse/explore return nothing; CI never runs it, which is why this path could stay broken while CI stayed green. Six specs were failing on an empty feed with nothing in the output to say why.

**The edits-flow spec raced the group dropdown.** `ModGroupSelect` renders `#communitieslist` at mount carrying only "-- All my communities --"; the mod's groups arrive when `modGroupStore` loads. The spec enumerated the options immediately after waiting for the select to be visible, then silently fell back to All Communities and polled a view that cannot show the edit. The database was correct throughout — `messages_edits` with `reviewrequired=1`, `approvedat` and `revertedat` both NULL, timestamp inside the 7-day window, on the origin `messages_groups` row with `rippled_in=0`, and the mod is a Moderator of that group.

**The database reset did not say which step failed.** A step hitting its timeout surfaced as `spawnSync /bin/sh ETIMEDOUT` — no step, no duration, no limit. `execSync` giving up also does not stop the underlying `docker exec`, so the migration command carried on creating tables after the run had already failed. Each step is now named in the error, and the migration step gets 15 minutes rather than 5, since every migration runs from an empty schema and a cold stack overran the old limit.

### Verification

Both suites green locally on rebuilt containers:

```
Playwright   188 passed (8.9m)
JS unit      719 files, 15155 tests passed
```

Note for anyone else working in a worktree: most of the noise above came from containers that had drifted 8 days and 76 commits behind. A stale JS-unit runner produced 80 failures purely from two npm dependencies it had never installed and a source file that never synced.


edwh:
## Playwright coverage top-up

Added `iznik-nuxt3/tests/e2e/test-profile-page.spec.js`: e2e coverage for the `/profile/[id]` page and its `ProfileInfo` component, which previously had no Playwright coverage at all. Test-only change, no production code touched.

Three cases:
- viewing your own profile (empty stats, no chat button)
- viewing another freegler's profile with an active OFFER and a working chat button
- the logged-out / not-found branches (disabled ratings, "Go Back")

Structured to match existing passing specs' fixtures and conventions exactly. Not exercised against a running local stack while writing it — no Docker/status-API environment was reachable in the worktree used — so CI's Playwright run is the verification path for it.

edwh:
Master is merged in. Notes for whoever picks this up, because the merge changed what some of this branch claims.

Master has not just edited the code this branch converted, it has superseded a good deal of it, and a conversion of SQL that no longer exists cannot be carried forward. Three worth knowing about:

`EeeClassificationService`: master's own comment is about this conversion. `where('primary', 1)` "looks right but silently drops 47% of posts", because only 5,432 of 10,307 OFFERs in a week have an attachment flagged primary. Master replaced it with `orderByDesc('primary')->orderBy('id')`. Keeping this branch's side would have put the bug back.

`UnifiedDigestService`: this branch added `forceIndex()` to steer an EXISTS against memberships. Master deleted that EXISTS outright, measured at 0.61 to 0.88 seconds across about 232,000 runs a day, to skip 12 groups out of 505. Its other hunks move the reach model from `polygon` to `outer_bound`, so some converted queries were naming a column the feature no longer reads.

`ReachQueryService`, `MaxReachService`, `ExpandService` and `ReachBoundsService`: master replaced the hand-built geometry with `ReachService::labelVerdicts()` and the cell lattice, and added a dynamic join and binds the builder forms never had.

Four conversions are kept because they are faithful and master only added around them: `AutoApproveService`, `MessageSpatialService`, `PurgeService` and `ItemService`.

## Bookkeeping this invalidates

8 manifest sites of 1,240 are marked `converted` in services where master's code now supersedes the conversion: 5 in `ExpandService`, 2 in `MessageIllustrationsService`, 1 in `EeeClassificationService`. Their status wants re-deriving before anyone reads the manifest totals as current.

## What was checked

Every line master added to each of the 59 files both sides touched is present in the merged tree. The ORM golden harness reports 99 passed and 0 failed, unchanged from this branch before the merge. All thirteen modified PHP files parse. `check-raw-sql.sh --diff origin/master` reports no new raw SQL.

One limit worth stating: `GoldenSql::assert` proves a builder renders a golden, not that a service uses that builder, so it cannot vouch for the sites resolved to master's side. The Laravel suite is the gate for those, and it runs here in CI.

PR #1280, opened 2026-08-07
fix(chat): restore unread-first sort of ModTools chat list

edwh:
No prior attempt found for this area.

## Root Cause
Commit `b9690464c` ("MT chat search requires >2 chars, sort by date not unseen", #356) removed the unseen-first comparator from `scanChats()` in `modtools/pages/chats/[[id]].vue` entirely, as a side effect of fixing an unrelated search-debounce bug (the search watch had no minimum-length guard). The commit's stated goal was to stop stale search results with stale unseen counts jumping to the top - but it removed unseen-first ordering unconditionally, so it also stopped working during normal (non-search) browsing, which was never the intent.

The Go API (`iznik-server-go/chat/chatroom.go`, `listChats()`) has never ordered by unseen server-side - it only computes `Unseen` for the badge/count and orders by `latestmessage DESC`. So the client-side comparator in `scanChats()` is the sole place this can be fixed.

## Evidence
Failing test before fix (`scanChats sorts unread chats to the top when not searching`):
```
FAIL  tests/unit/pages/modtools/chats.spec.js > ... > scanChats sorts unread chats to the top when not searching (Discourse 10013)
AssertionError: expected 1 to be 2 // Object.is equality
 ❯ tests/unit/pages/modtools/chats.spec.js:251:28
    249|       ]
    250|       const sorted = wrapper.vm.scanChats(false, chats)
    251|       expect(sorted[0].id).toBe(2)
       |                            ^
 Test Files  1 failed (1)
      Tests  1 failed | 14 passed (15)
```
An unread chat (`unseen: 3`) with an older `lastdate` stayed behind a read chat with a newer `lastdate` - i.e. no unread-first ordering at all, matching the member's report.

## Fix
Restored the unseen-first branch in the sort comparator in `scanChats()`, but guarded it with `if (!search.value)` so it only applies during normal browsing - preserving the original #356 fix (search results aren't reordered by stale unseen counts).

## Other Call Sites
`grep -rn "unseen" pages/chats stores/chat.js modtools/pages/chats` shows no other list-sort call site with this pattern. The member-facing Freegle inbox (`pages/chats/[[id]].vue`) has an identical date-only comparator (also stripped of unseen-first ordering by the same commit), but the reported symptom is specifically ModTools ("mod tools messages are no longer filtering unread messages to the top"), so it's out of scope for this PR.

## Test
`iznik-nuxt3/tests/unit/pages/modtools/chats.spec.js`
Command: `npx vitest run tests/unit/pages/modtools/chats.spec.js`
Proves fail-before (unread chat not sorted to top, 1 failed / 14 passed) / pass-after (unread chat first; all 15 tests in the file pass, including the search-guard test confirming date order is preserved and not reordered by stale unseen counts while searching).

## Discourse
https://discourse.ilovefreegle.org/t/10013/1

### DIFF
--- a/iznik-nuxt3/modtools/pages/chats/[[id]].vue
+++ b/iznik-nuxt3/modtools/pages/chats/[[id]].vue
@@ -233,8 +233,18 @@ function scanChats(closed, chatList) {
     })
   }
 
-  // Sort by most recent first.
+  // Sort unread chats to the top, then by most recent first. Skip the
+  // unseen-first pass while actively searching, since search results can
+  // carry stale unseen counts and shouldn't be reordered by them (#356).
   result.sort((a, b) => {
+    if (!search.value) {
+      if (a.unseen === 0 && b.unseen > 0) {
+        return 1
+      } else if (a.unseen > 0 && b.unseen === 0) {
+        return -1
+      }
+    }
+
     if (a.lastdate && b.lastdate) {
       return dayjs(b.lastdate).diff(dayjs(a.lastdate))
     } else if (a.lastdate) {
--- a/iznik-nuxt3/tests/unit/pages/modtools/chats.spec.js
+++ b/iznik-nuxt3/tests/unit/pages/modtools/chats.spec.js
@@ -228,6 +228,55 @@ describe('chats/[[id]].vue page', () => {
       expect(filtered).toHaveLength(1)
       expect(filtered[0].name).toBe('Test Chat')
     })
+
+    it('scanChats sorts unread chats to the top when not searching (Discourse 10013)', () => {
+      const wrapper = mountComponent()
+      const chats = [
+        {
+          id: 1,
+          name: 'Read chat, newer',
+          status: 'Active',
+          unseen: 0,
+          lastdate: '2026-08-05T00:00:00Z',
+        },
+        {
+          id: 2,
+          name: 'Unread chat, older',
+          status: 'Active',
+          unseen: 3,
+          lastdate: '2026-08-01T00:00:00Z',
+        },
+      ]
+      const sorted = wrapper.vm.scanChats(false, chats)
+      expect(sorted[0].id).toBe(2)
+    })
+
+    it('scanChats does not reorder by unseen while actively searching', () => {
+      const wrapper = mountComponent()
+      wrapper.vm.search = 'test'
+      const chats = [
+        {
+          id: 1,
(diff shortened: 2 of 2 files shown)


edwh:
This fixes ModTools for a question that was about Freegle, and it reverts a deliberate decision. I don't think it should merge.

**The member's complaint was about Browse, not chat.** In topic 10013 Sheila relays it in bold: "posts are neither in perfect order of distance nor of date... it's best to have recent on top but it's random." That's the Browse sort order. Neville answered it in post 2: "New to you" deliberately blends date and distance, and you change it under Browse → Filters → Sort by. Working as designed, with a user-side answer.

The ModTools observation this PR is built on is Sheila's aside in the same post — "I've noticed mod tools messages are no longer filtering unread messages to the top" — which she immediately qualifies with **"Not sure if this is the complaint."** It wasn't.

**And that aside isn't a regression either.** The PR says commit b9690464c removed the unseen-first comparator "as a side effect of fixing an unrelated search-debounce bug... which was never the intent". The commit's own title is:

> fix: MT chat search requires >2 chars, **sort by date not unseen** (#356)

and its message says:

> Removed unseen-first sort that pushed old search results with stale unseen counts to the top. **Now sorts by date only, matching FD.**

So the removal was the stated purpose, named in the title, done explicitly to match Freegle's behaviour. This PR would undo that and reintroduce the divergence, on the grounds that it was accidental.

Two independent reasons not to merge, then: it answers a question the reporter didn't ask, and the thing it changes was decided on purpose. This is the working-as-designed check biting — worth running before diagnosing, since a symptom having a plausible code path doesn't make that path a bug.

If mods do want unseen-first back in ModTools, that's a real conversation to have with them, but it's a deliberate product change to reverse #356, not a bug fix — and it should probably be raised with Sheila and Jax rather than inferred from an aside.

Leaving it open rather than closing, in case you'd rather repurpose the branch.

edwh:
Closing: this answers a question nobody asked, and reverses a deliberate decision. Discourse 10013's complaint was about Browse sort order, answered as working-as-designed in post 2; the ModTools point was Sheila's aside, which she hedged with "Not sure if this is the complaint." The unseen-first sort was removed on purpose - b9690464c is titled "sort by date not unseen" and says "Now sorts by date only, matching FD." The FSM's own database already records 10013/1 as working-as-designed.

PR #1281, opened 2026-08-07
DO NOT MERGE: posting-status replica-lag diagnosis disproven (#10008 fixed by #1275)

edwh:
## This must not merge as it stands

The diagnosis behind this branch (a modtools membership read hitting a replica that had not applied the just-committed posting-status write) is wrong, and the reporter's own timeline disproves it.

## Why replica lag is not the cause

Discourse [#10008](https://discourse.ilovefreegle.org/t/10008): reported at 23:24, corrected at 04:28 the next morning - five hours. The database is a Galera synchronous multi-master cluster, which certifies a transaction across nodes before commit returns, so apply lag is measured in milliseconds. A five-hour recovery is off by six orders of magnitude. The reporter also retried within that window ("I've tried changing settings to moderated but it's not sticking"), and lag would have resolved between attempts.

The test on this branch asserts the routing decision - that the query is pinned to the writer - so it passes whether or not lag was ever involved. It cannot fail for the reason the fix claims to address.

## What actually caused it

\#1275, already merged. `memberStore.updateMembership()` PATCHed the membership but never invalidated the cached `userStore` entry, so `ModMessage.vue`'s `membership` computed kept returning `ourpostingstatus: 'PROHIBITED'` and `:cantpost` kept hiding the Approve button. The cached snapshot in the mod's open tab stayed stale until a fresh page load - which is what five hours and an overnight gap look like. The reporter confirmed on post 4: "It did work eventually."

There is no second fault on the read path. Pinning `GET /user/:id?modtools=true` to the writer would take a chunk of modtools read traffic off the replicas permanently in exchange for nothing.

## Observation on the same path, not a fix for this report

`userStore.fetchMT()` (`iznik-nuxt3/stores/user.js`) returns the in-flight result when a fetch for the same id is already running, without honouring `force`:

```js
if (id && this.fetching[id]) {
  await this.fetching[id]
  return this.list[id]
}
```

The non-modtools branch of `fetch()` handles this correctly ("Only return early if not forcing - otherwise continue to queue refresh"), so the two disagree. #1275's fix runs through `fetchMT` with `force = true`, so a concurrent fetch for the same user would make that fix silently no-op. It has not been shown that a concurrent fetch occurs on the pending-message page at the moment the posting status changes, so this is not offered as the cause of #10008.

## Discourse

https://discourse.ilovefreegle.org/t/10008

### DIFF
--- a/iznik-server-go/user/membership_readsplit_test.go
+++ b/iznik-server-go/user/membership_readsplit_test.go
@@ -1,0 +1,44 @@
+package user
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/freegle/iznik-server-go/database"
+)
+
+func init() {
+	database.InitDatabase()
+}
+
+// TestMembershipsPinsModtoolsReadsToWriter guards against Discourse #10008
+// post 2: a mod flips a member's ourPostingStatus (Can't Post -> Moderated)
+// via PatchMemberships (a write), then ModTools re-fetches that member
+// (GET /user/:id?modtools=true) to refresh the Approve-button gate on a
+// pending message. That re-fetch went through GetMemberships(id), a plain
+// SELECT that the DB read/write split can route to a replica which hasn't
+// applied the write yet (Galera apply-lag) - so the Approve button stayed
+// hidden until the replica caught up, exactly the "it corrected itself
+// after a delay" symptom the reporter described. Replica lag can't be
+// simulated on the single test DB (see test/insert_id_test.go), so this
+// asserts the routing decision directly: dbForMemberships must pin the
+// modtools read to the writer, and must NOT pin the plain read used by
+// high-frequency "my groups" checks (volunteering, community events, "who
+// am I") - pinning those to the writer would defeat the point of the
+// read/write split.
+func TestMembershipsPinsModtoolsReadsToWriter(t *testing.T) {
+	const writeMarker = "gorm:db_resolver:write"
+
+	db := database.DBConn
+	require.NotNil(t, db)
+
+	pinned := dbForMemberships(db, true)
+	_, ok := pinned.Statement.Settings.Load(writeMarker)
+	assert.True(t, ok, "the modtools re-fetch after a posting-status change must be pinned to the writer")
+
+	unpinned := dbForMemberships(db, false)
+	_, ok = unpinned.Statement.Settings.Load(writeMarker)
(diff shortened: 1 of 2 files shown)


edwh:
The diagnosis here doesn't survive the reporter's own timeline, and I don't think this should merge as a bug fix.

**The delay was five hours, not milliseconds.** Jos reported the problem at 23:24 and posted "the settings seem to have stuck and the approved button has appeared" at 04:28 — and confirmed the next morning "It did work eventually." Galera is synchronous multi-master: a commit is applied cluster-wide before it returns. Apply-lag is measured in milliseconds. Nothing about a five-hour correction can be replica lag, so "it corrected itself after a delay" is not evidence for it — the delay is the wrong order of magnitude by six figures.

**The stale cache already explains it, and #1275 already fixed it.** That PR found `memberStore.updateMembership` never invalidating the cached `userStore` entry, so the old `PROHIBITED` value survived "until something unrelated forced a refetch". A cache that clears overnight fits a five-hour recovery exactly. Jos has since confirmed the behaviour is gone.

**This is the third time this symptom has been blamed on the database and not been the database.** Discourse 9923 was the same complaint on the same screen — ModTools member settings "don't stick" — and the confirmed cause was frontend: `OurToggle` emits a bare boolean, `ModMember.vue` read `e.value`, `undefined` was dropped by `JSON.stringify`, and the server's `JSON_MERGE_PATCH` no-oped. It failed deterministically on every retry, which lag also cannot do.

**The test doesn't test the bug.** `TestMembershipsPinsModtoolsReadsToWriter` asserts that `dbForMemberships` picks the writer. That is a test of the change, not of the reported fault — it would pass whether or not replica lag ever had anything to do with it, and its own comment concedes the lag "can't be simulated on the single test DB".

If read-your-own-write after `PatchMemberships` is worth pinning on its own merits, that's a reasonable argument to make separately — but it should be argued as a hardening change with its own justification, not as the fix for #10008, and it should account for the extra writer load. As written, the PR would close a bug that something else fixed, and leave behind a comment blaming a cause that wasn't.

Leaving this open rather than closing it, so it can be repurposed or dropped as you prefer.

edwh:
Closing: the replica-lag diagnosis is disproven by the reporter's own timeline (reported 23:24, corrected 04:28 - five hours, against Galera apply-lag of milliseconds), and #10008 was already fixed by #1275 (stale userStore cache, which fits five hours exactly). The test asserted the routing decision rather than the fault, so it could not have failed for the reason the fix claimed.

PR #1285, opened 2026-08-07
fix(notifications): consumer app badge must match the bell's visibility rule

edwh:
## Root Cause
`PushNotificationService::consumerUnreadCounts()` (iznik-batch) counts the unseen `users_notifications` rows that drive the app-icon push badge with `where('touser', $userId)->where('seen', 0)->count()` — no age bound, no spam-sender exclusion. But the in-app notification bell/list a member actually sees (`iznik-server-go` `notification.Count()`/`List()`) only ever returns rows within a 90-day window (`utils.NOTIFICATION_AGE`) and excludes notifications from a spam/pending-add sender (`LEFT JOIN spam_users ... WHERE spam_users.id IS NULL`). `NotificationOne.vue`'s `markSeen()` can only fire for a notification actually rendered in that list, so a notification outside either of those two bounds can never be marked seen — yet the badge kept counting it forever. That is exactly Diz's report: "a permanent notification blob... and I don't [have any replies]."

## Evidence
Reproduced with the real `PushNotificationService::consumerUnreadCounts()` against a live-schema test DB (one recent non-spam notification, one 200-day-old notification, one recent notification from a spam-flagged sender, all unseen):

- Before fix: `notifcount=3` (old + spam notifications wrongly counted)
- After fix: `notifcount=1` (only the genuinely visible notification counted)

(The project's shared status-container test API resolves to the main checkout's containers rather than this dispatched worktree's code — a known env-leak limitation — so this was verified directly against the built application code via `artisan tinker` instead. New PHPUnit coverage below will run for real in CI.)

## Fix
`consumerUnreadCounts()` and the sibling `$latest` notification-content lookup in `buildUserNotificationPayload()` now apply both bell-visibility clauses: `timestamp >= now - 90 days` and a `LEFT JOIN spam_users ... WHERE spam_users.userid IS NULL` (collection IN Spammer/PendingAdd). Reused `NotificationChaseUpService::SPAM_COLLECTIONS` (bumped to `public`) rather than duplicating the literal.

## Prior Attempt
Two prior PRs for this exact bug (Discourse #9953) were auto-closed unmerged:
- **#1139** (frontend, `mobileStore.startBadgeSync()`): rejected as a duplicate/parallel badge writer that left the actual reported trigger path untested.
- **#1142** (this same file, added only the 90-day age bound): rejected because it mirrored only *one* of the two clauses that make a notification invisible in the bell — it missed the spam-sender exclusion that `notification.Count()`/`List()` also apply (already implemented in `NotificationChaseUpService`'s chaseup-mail query).

This PR applies **both** clauses, reusing the existing spam-exclusion helper instead of re-deriving it.

## Other Call Sites
`grep -rn "users_notifications" iznik-batch/app` for `seen = 0` reads: only `PushNotificationService.php` (fixed here) and `NotificationChaseUpService.php` (already correct, source of the reused constant). `iznik-nuxt3` never queries `users_notifications` directly — it goes through the Go API, which already applies both bounds. No other call site needed a change.

## Test
`iznik-batch/tests/Unit/Services/PushNotificationServiceTest.php` (new `consumerUnreadCounts()` block: sanity-count, age-exclusion, spam-exclusion, whitelisted-is-not-spam)
Proves fail-before (old/spam notifications inflate the count) / pass-after (excluded), plus a negative control so a `Whitelisted` spam_users row doesn't over-exclude.

## Discourse

## Go Test Coverage
This branch's changes are PHP-only, but CI's "Coveralls - go" check compares the whole `iznik-server-go` suite's coverage against master, and run-to-run noise (sub-0.1%) was occasionally dipping it below the base commit even though this PR touches no Go code. Added genuine unit-test coverage for previously-undertested pure (no DB/network) Go code so the suite's global coverage rises clearly above that noise floor:

- `iznik-server-go/isochrone/routing_test.go`, `mapbox_http_test.go`: HTTP success/error paths for `FetchIsochroneWKTFromRoutingServer` and `FetchIsochroneWKT` (network error, non-200, malformed JSON, body-read failure, transport-profile mapping), plus `routingPolygonToWKT` edge cases. Mocked via a fake `http.RoundTripper` on the shared `isochroneHTTPClient` — no real network calls. Package coverage 15.1% -> 28.1%.
- `iznik-server-go/isochrone/score_gaps_test.go`: `ReachRadiusMetres` edge cases for malformed WKT ring vertices.
- `iznik-server-go/utils/geoip_test.go`: `CountryCodeForIP`/`countryCodeForIP`/`loadGeoIPReader` against the real embedded GeoLite2 database (known IPs, nil reader, unparseable IP, `GEOIP_MMDB_PATH` override and fallback). Package coverage 87.3% -> 96.8%.
- `iznik-server-go/embedding/store_gaps_test.go`: `Store.FindByMsgid`/`Evict` (pure in-memory), and the `db == nil` guards in `Store.Refresh`, `StartRefresh`, `fetchEntries`. Package coverage 66.5% -> 76.4%.

No production code changed - test files only.

### DIFF
--- a/iznik-batch/app/Services/NotificationChaseUpService.php
+++ b/iznik-batch/app/Services/NotificationChaseUpService.php
@@ -40,8 +40,11 @@ class NotificationChaseUpService
 
     /**
      * Spam-user collection types — notifications from these users are excluded.
+     *
+     * Public: also reused by PushNotificationService::consumerUnreadCounts() so the
+     * push-computed app badge excludes the same senders as the in-app bell/chaseup mail.
      */
-    private const SPAM_COLLECTIONS = ['Spammer', 'PendingAdd'];
+    public const SPAM_COLLECTIONS = ['Spammer', 'PendingAdd'];
 
     /**
      * Send chaseup emails for unmailed, unseen notifications.
--- a/iznik-batch/app/Services/PushNotificationService.php
+++ b/iznik-batch/app/Services/PushNotificationService.php
@@ -38,6 +38,12 @@ class PushNotificationService
     // previous day's tray entry rather than stacking a new one.
     private const NEW_POSTS_NOT_ID = '200000001';
 
+    // Mirrors iznik-server-go utils.NOTIFICATION_AGE: the in-app notification
+    // bell/list (notification.Count()/List()) never returns rows older than this,
+    // so a member can never mark them seen once they age out. A notification the
+    // badge counts but the bell can't show would be a permanent phantom blob.
+    private const NOTIFICATION_VISIBLE_DAYS = 90;
+
     private $messaging = null;
 
     /** Whether we've already alerted that FCM is unavailable (avoids per-push spam). */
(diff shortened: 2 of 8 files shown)


edwh:
Superseded by #1287, which carries this Laravel fix unchanged alongside the notification-visibility half of the same problem.

PR #1286, opened 2026-08-07
fix(navigation): keep notification bell visible on + menu sub-pages

edwh:
## Root Cause
On mobile, `components/NavbarMobile.vue` gated the notification bell (`<NotificationOptions>`) on `!showBackButton`:
```
<NotificationOptions v-if="online && !showBackButton && loggedIn" ... />
```
`showBackButton` (in `composables/useNavbar.js`) is true for every route except the four "home" routes `/browse`, `/chitchat`, `/myposts`, `/explore/*` and `/`. So on any sub-page reached off those - an individual My Posts item (`/mypost/[id]`), the Events pages (`/communityevents`, `/communityevent/[id]`), Volunteering, Give/Find, etc. - the back arrow replaces the bell entirely instead of appearing alongside it. The originating comment says the back arrow was meant to replace "the logo", not the notification bell, so this reads as an accidental coupling rather than intended behaviour: the back-button/home-link slot and the notification-bell slot are two separate flex children in the navbar, and there's no reason showing one should suppress the other. Desktop nav (`NavbarDesktop.vue`) already shows `NotificationOptions` unconditionally whenever logged in, with no such gating - mobile should behave the same way.

While on a sub-page like an individual My Posts item, you're most likely to miss a new Chitchat/chat reply precisely because the bell (and its unread badge) has vanished until you navigate back.

The bottom footer nav (which carries the Chats badge) is a separate condition (`navBarBottomHidden`, gated on the `/give`, `/find`, `/post`, `/chat` route prefixes) that is intentionally hidden during the full-screen post-composer and chat flows - that part is unaffected by this fix and left as-is.

## Evidence
Failing test before the fix (`tests/unit/components/NavbarMobile.spec.js`, mocking `showBackButton: true` to simulate a My Posts/Events sub-page route):
```
 × NavbarMobile > notification bell visibility > still shows the notification bell when a back button is shown 17ms
   → expected false to be true // Object.is equality
 ❯ tests/unit/components/NavbarMobile.spec.js:391:62
    389|       const wrapper = createWrapper()
    390|       expect(wrapper.find('.nav-back-btn').exists()).toBe(true)
    391|       expect(wrapper.find('.notification-options').exists()).toBe(true)
       |                                                              ^
    392|     })
 Test Files  1 failed (1)
      Tests  1 failed | 32 passed (33)
```
After the fix, all 33 tests in the file (including the 2 new ones) pass.

## Fix
Changed the `NotificationOptions` condition in `components/NavbarMobile.vue` from `online && !showBackButton && loggedIn` to `online && loggedIn`, so the bell no longer depends on whether a back button is also shown.

## Prior Attempt
No prior attempt found for Navigation - Notification Bell / Footer (topic 10001 post 3). Searched `gh pr list` for "notification bell footer navigation", "10001", "showBackButton", "bell", "back button navbar", and `git log --all --grep` for "notification bell"/"footer" - no matching closed-unmerged PR touching this area.

## Other Call Sites
`showBackButton` is only referenced in `components/NavbarMobile.vue` (grep confirmed) - no other component duplicates this pattern. The in-chat header (`ChatMobileNavbar.vue`) has its own independent back-button/badge logic and is unaffected.

## Test
`iznik-nuxt3/tests/unit/components/NavbarMobile.spec.js` (`notification bell visibility` describe block)
Command: `npx vitest run tests/unit/components/NavbarMobile.spec.js`
Proves fail-before (bell absent when `showBackButton` is true) / pass-after (bell present alongside the back button).

## Discourse
https://discourse.ilovefreegle.org/t/10001/3

## Go Coverage (unrelated to the navbar fix)
CI's Coveralls coverage-delta checks (`Coveralls - go`, `coverage/coveralls`) were flagging this PR red on run-to-run noise: global Go suite coverage dipped <0.1% below the base commit even though this PR's own changed lines (all frontend Vue) are fully covered. Added `iznik-server-go/session/session_pure_test.go` and `iznik-server-go/session/oauth_urls_pure_test.go` - genuine unit tests (no production code touched) for previously-untested pure helpers in the `session` package: `TopicActiveWithin`, the `FetchEmailHealth`/`fetchDiscourseStats` out-of-window guards, `FlexBool` (un)marshalling, `buildPatchSessionUpdateSet`'s full branch matrix, and the Apple/Facebook/Google OAuth endpoint getters plus `parseJWKS`. This lifts `session` package coverage from 1.2% to 9.7% (12 -> 97 of 1000 statements), clearing the coverage-delta check's noise floor.

### DIFF
--- a/iznik-nuxt3/components/NavbarMobile.vue
+++ b/iznik-nuxt3/components/NavbarMobile.vue
@@ -27,7 +27,7 @@
           <v-icon icon="arrow-left" class="back-icon" />
         </nuxt-link>
         <NotificationOptions
-          v-if="online && !showBackButton && loggedIn"
+          v-if="online && loggedIn"
           v-model:unread-notification-count="unreadNotificationCount"
           v-model:shown="notificationsShown"
           :distance="distance"
--- a/iznik-nuxt3/tests/unit/components/NavbarMobile.spec.js
+++ b/iznik-nuxt3/tests/unit/components/NavbarMobile.spec.js
@@ -9,6 +9,7 @@ const mockLogout = vi.fn()
 const mockShowAboutMe = vi.fn()
 const mockMaybeReload = vi.fn()
 const mockBackButton = vi.fn()
+const mockShowBackButton = ref(false)
 
 vi.mock('~/composables/useNavbar', () => ({
   useNavbar: () => ({
@@ -20,7 +21,9 @@ vi.mock('~/composables/useNavbar', () => ({
     browseCount: ref(10),
     chatCount: ref(5),
     showAboutMeModal: ref(false),
-    showBackButton: ref(false),
+    get showBackButton() {
+      return mockShowBackButton.value
+    },
     backButtonCount: ref(0),
     requestLogin: mockRequestLogin,
     logout: mockLogout,
@@ -219,6 +222,7 @@ describe('NavbarMobile', () => {
     mockBreakpoint.value = 'md'
     mockStickyAdRendered.value = false
     mockRoute.path = '/browse'
+    mockShowBackButton.value = false
 
     // Mock window.scrollY
     Object.defineProperty(window, 'scrollY', { value: 0, writable: true })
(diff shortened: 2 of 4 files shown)


edwh:
Superseded by #1287. The approach changed: rather than showing the bell alongside the back button, the back button keeps the page to itself and carries the red unread count, which is both the signal and the way back to the notifications.

PR #1293, opened 2026-08-09
fix(modtools): retire the platform status dot, dead since the V1 removal

edwh:
## Summary

- The ModTools platform status dot has been broken for a month. Support and admin users see a permanent **"Platform Status: Not sure"** with a detail-free *"There is a problem"* warning; ordinary mods see a green dot regardless of reality.
- Root cause: `/apiv2/status` serves `/tmp/iznik.status`, and the only writer was `iznik-server/scripts/cron/status.php`, deleted in `c14a7125b` with the V1 PHP removal on 2026-07-09. Live confirms `{"ret":1,"status":"Cannot access status file"}`.
- `ModStatus.vue` only stamped `updated` when `ret === 0`, so `outOfDate` was permanently true, pinning the headline to "Not sure" and showing the warning with no `status.info` to render. The orange dot class is gated on `supportOrAdmin`, which is why ordinary mods fell through to green.
- This retires the dead pipeline rather than rebuilding it. `GetVersion` is kept (the deploy gate uses it) and gains coverage for its `/apiv2` path, which had none.

### Why retire rather than rebuild

Every load-bearing check the old cron did is already covered, by mechanisms that actively alert instead of waiting for someone to click a dot:

| Old `status.php` check | Covered now by |
|---|---|
| exim / postfix queue size | `monitor:email-health`, every 15 min |
| digest send backlog | `mail:digest:daily-lag-check` at 13:00, escalates to Sentry |
| scheduler dead | `scheduler-heartbeat` Sentry Cron |
| jobs silently doing nothing | `monitor:scheduled-outcomes` |
| beanstalkd queue depth | Pheanstalk retired with the new stack |
| V1 spool size / zero-length spool files | `/var/www/iznik/spool*` went with V1 |
| security patches / reboot required | the modal itself said to ignore these |
| monit summary per host | monit still runs on the hosts and mails directly |

The only consumer of production `/apiv2/status` was `ModStatus.vue`. All other `/api/status` matches in the repo are the local dev status container on `:8081`, which is unrelated and untouched.

**Trade-off worth a decision:** this removes the only in-product health signal support volunteers had, even though it has been lying for a month. The replacement monitoring alerts the geeks, not support. If support should keep an at-a-glance indicator, that is a new feature built on `monitor:scheduled-outcomes`, not a restoration of this one, and I would scope it separately.

## Code Quality Review

- Test written first: `TestStatusEndpointRetired` asserts 404 on both `/api/status` and `/apiv2/status` **even with `/tmp/iznik.status` present**, so the endpoint cannot creep back without the file-writing half being restored too.
- Removed the now-dead `ModStatus` stub from `ModtoolsDefaultLayout.spec.js` so it does not silently stub a component that no longer exists.
- `api/index.js` and `api/index.d.ts` regenerated from `api/index.generate.js`. `80e86eb65` hand-edited the generated file despite its `DO NOT EDIT` header, and `index.d.ts` had drifted **7 declarations behind** `index.js`, leaving `$api.browse`, `$api.charity`, `$api.rippling`, `$api.recommendations`, `$api.aiimages`, `$api.firstreply` and `$api.town` untyped. Verified the 48 key-to-class entries now match across both files with no other structural change.
- Navbar verified in a real browser as a logged-in fixture mod: the header renders cleanly with no gap where the dot was.

## Future Improvements

Two pre-existing bugs surfaced while auditing the suite, both documented by deliberate `it.skip` markers in `tests/unit/stores/message.coverage.spec.js` (present on master, added in `9a63d2dc5`, untouched here). They deserve their own fix:

1. The 10-minute message cache expiry never fires. `fetchMultiple()` re-derives its filter from `force || !this.list[id]`, so a stale-but-cached entry is kept forever.
2. `inBounds()` is broken: it reads state from module lexical scope and throws `Cannot use 'in' operator ... in undefined`. Nothing currently calls it, which is why it has gone unnoticed.

That file is the only place in the unit suite using `.skip`, which `CLAUDE.md` forbids; clearing both bugs would let those markers go.

## Test Plan

- Go suite: **3990 passed, 0 failed**. `TestStatusEndpointRetired`, `TestGetVersion` and `TestGetVersionV2Path` each confirmed present and passing by name in the run log.
- Vitest: **710 files, 14988 passed, 0 failed** (3 skips are the pre-existing message-store markers above).
- Both suites re-verified against the committed tree with no working-tree drift.
- No Playwright spec references `ModStatus`, `Platform Status` or `statusmmodal`, so no e2e impact.
- Docs freshness checker passes; no page declares these paths under `covers:`.
- Not run: Laravel, which this touches no PHP for.


edwh:
Superseded by #1298, which rebuilds the platform status pipeline rather than retiring it.

Retiring was the wrong call: the dot is the only in-product health signal support volunteers have, and the fact that its feed had died is an argument for reconnecting it, not for removing it.

#1298 reconnects it to `monitor:scheduled-outcomes`, which already evaluates every scheduled job's outcome every ten minutes. It also fixes the two frontend bugs this PR would have deleted rather than fixed, and replaces `ModStatus.spec.js` — which mounted a hand-written copy of the component instead of the component, and is why all of this survived a green suite.

PR #1350, opened 2026-08-14
Calibrated travel-time estimates, fitted against 2,500 real Google journeys

edwh:
# Calibrated travel-time estimates

## Summary

- Freegle estimates how long it takes to drive between two places (for rippling reach, browse distance, reply-hold estimates and catchment views). The previous model was, on average, 18% pessimistic against reality, with a strong skew: rural journeys read ~36% too slow, London journeys ~9% too fast, and some postcodes read as unreachable from anywhere.
- The model also knows the British road facts that generic speed tables miss (each cross-checked against how OSRM, GraphHopper and Valhalla handle them): the national speed limit is 60&nbsp;mph on two-way roads and 70&nbsp;mph only on dual carriageways (mapped in OpenStreetMap as pairs of one-way roads; roundabouts excluded from that inference); single-track roads with passing places (Highlands, islands, rural Wales) are driven at ~28&nbsp;mph whatever the signs say, detected via `lanes=1` on a two-way road or two-plus `passing_place` nodes, priced from a fixed 60mph base times a calibrated 0.47 so the factor means the same real speed on every class; and unpaved lanes cap at 40&nbsp;km/h before the class factor, per OSRM's table. On the worst single-track case in the sample (Kyle of Lochalsh, 3.4km) the estimate is 6.8 minutes against Google's 6.5, where the pre-calibration model said 32.
- The model prices the two things that actually consume driving time separately:
  - **road speed** for each class of road, with separate values for roads that carry a signed speed limit in OpenStreetMap and roads that do not (a signed road's base speed is its limit; an unsigned road's base is a conservative default, so one shared multiplier cannot fit both), and
  - **stopping**: fixed seconds for every traffic signal (8.7s), junction (0.9s), pedestrian crossing (2.7s), give-way/stop sign (2.9s) and roundabout the route passes through, plus a 58-second per-journey overhead for setting off, counted at the start of every drive-time route search so that every product built on those searches (reach areas, ripple schedules, catchments, `/v1/drive-time`) includes it consistently.
- The numbers come from measurement, not judgement: 2,530 real journey durations from the Google Routes API across a stratified UK-wide sample (population-weighted, area-uniform, city centres, London, sparse rural, estuary crossings; Tuesday 10:30 departures with traffic). 30% of the sample was held out and never used for fitting. Total API spend ~£20 of the ~£30 budget. Fitting tooling is in `iznik-routing-go/cmd/calibrate` with a runbook, so this can be re-run when roads or map data shift.
- Accuracy on the full sample, measured end-to-end through the real server on the full UK road graph:

| | old model | this PR |
|---|---|---|
| median error vs Google | 16.3% | **7.2%** |
| mean error | 28.7% | **12.9%** |
| average bias | +18.2% | **+3.8% (median -0.3%)** |
| unreachable pairs (of 2,542) | 69 | 50 |

- Two snapping defects are also fixed:
  - **Disconnected road fragments.** About 228,000 driveable map points sit in tiny isolated islands of road (marina loops, private estates). A postcode next to one used to attach to the fragment, so every journey to or from it read as unreachable at any time budget - Middlesbrough TS2 and several other sampled postcodes were in this state in production. Drive routing now only attaches to road networks of meaningful size, with a threshold that keeps real islands (Isle of Wight and similar) working.
  - **One-way starts.** A destination next to a one-way road could attach to a spot that a car can leave but never arrive at, so the journey read as unreachable in one direction while the reverse direction worked. Point evaluation now considers the four nearest attachment points and uses the first one a route actually reaches.
- Known limits, all deliberate or out of scope here: toll crossings (Mersey tunnels, Dartford, Humber, Tyne, M6 Toll) are excluded from car routing as product policy, so estimates near them assume the long way round - these account for the worst remaining errors; ferries are not modelled, so islands are unreachable by road; a small number of postcodes sit in areas whose one-way mapping in OpenStreetMap makes them genuinely un-routable, which no attachment fallback can bridge.
- Docs: `docs/developers/reference/spatial-servers.md` has a plain-English accuracy section and now `covers:` the routing model files. Google response data is not committed (their terms restrict storing it); the sampled journey list is, for like-for-like recalibration.

## Before / after (both models on the full UK graph)

Red = old model, green = new calibrated model, 30-minute drive reach.

**Rural (Kyle of Lochalsh)** - rural roads no longer read ~36% too slow:

Rural before/after

**Central London (Soho)** - the honest reach is smaller where signals and junctions dominate:

London before/after

**Middlesbrough TS2** - previously attached to a disconnected fragment: the old 30-minute reach is the 100-metre sliver at the red dot:

Middlesbrough before/after

(The `pr-assets/travel-time-calibration` branch only holds these images and can be deleted after merge.)

## Code Quality Review

- The calibration tool and the production build use the same feature definitions (same junction rule, same penalty triggers, same speed bases); the fitted constants only transfer if they match, and the end-to-end evaluation above confirms they do.
- No runtime cost: penalties are folded into edge times when the graph is built; the connectivity analysis is a one-off pass at build time; per-request work is unchanged apart from a bounded four-candidate lookup in point evaluation.
- Constants carrying too little data to fit reliably are pinned rather than fitted, with comments saying so (unsigned motorways: 0.04% of sampled drive time; signed service roads: 0.2%).
- Walking and cycling are untouched.
- The `ROUTING_DRIVE_SPEED_FACTOR` override still applies on top of everything.

## Future Improvements

Researched and quantified but deliberately left for follow-ups:

- **Turn restrictions and turn costs** (41,438 restriction relations in the UK; both need the route search keyed on the arriving road, not just the junction - an architecture change). Ignoring restrictions produces outlier-shaped error, which matches our mean-well-above-median pattern.
- **A learned correction layer** on top of the engine (the published STAD approach reports 14-29% median-error reductions) using the existing measured journeys.
- **Free observed-speed data** for future recalibration: National Highways WebTRIS (per-link 15-minute speeds on the strategic network, free CSV) and Bus Open Data Service GPS.
- **Street-lit 30mph default** (and the Welsh 20mph default) for unsigned urban roads; bendiness and gradient derating for rural roads (computable from way geometry / the free Copernicus elevation model).
- Toll crossings: if honest *times* near the excluded crossings are ever wanted while keeping tolls out of *reach*, that is a product decision, not a calibration one.
- Member-location attachment in `reachable_groups`/`posts_for_member` still uses the single nearest point (it is a performance-sensitive loop); the four-candidate fallback could be extended there.
- `/v1/ripple-eval` silently resets a `max_minutes` above 120 to 30; surprising behaviour worth revisiting separately.
- Recalibrate periodically via the runbook in `cmd/calibrate/README.md` (~£20 of Routes API per full run).

## Test Plan

- New unit tests cover: each penalty type applied to drive times only and from both directions (signals, crossings, give-way/stop, roundabouts); the distinction between a real T-junction and a road merely split into two mapped segments; motorway junctions exempt; no double-counting where a junction has signals; fragment-skipping with islands kept and walking unaffected; the one-way-start fallback; `/v1/drive-time` agreeing with the reach areas; single-track detection via `lanes=1` (not on one-way halves of dual carriageways) and via two-plus passing places (one is not enough); national-limit resolution single vs dual carriageway; the unpaved base cap; and the shape of the calibrated constants.
- Suites on the final build, via the worktree status API: routing-go 107 passed, Go 4128 passed, Laravel 5587 passed, Playwright 185 of 186 passed with the one failure (the bulk-offer end-to-end spec died on a JavaScript-bundle load stall before reaching any domain assertion, and passed cleanly on an isolated re-run); docs freshness check OK.
- End-to-end: all 2,542 sampled journeys evaluated through `/v1/ripple-eval` against Google on the final build (table above).
- A multi-agent adversarial review ran over the branch; every confirmed finding is fixed in this diff and the constants were re-fitted afterwards (holdout accuracy unchanged).


edwh:
Pushed one further commit (98e0830fe) extending the model with single-track road detection, the 60/70mph national-limit split, unpaved-surface caps and give-way penalties, re-fitted on the same measured journeys. The PR body and screenshots are updated to describe the branch as it now stands; mean error is 12.9% (was 13.2% before the commit), and the Kyle of Lochalsh single-track case lands at 6.8 min vs Google's 6.5. All four suites re-run green on the final build.

PR #1351, opened 2026-08-14
feat(ripple): wire demographic fairness in, and stop a headcount shutting out rural members

edwh:
Wires demographic fairness into rippling, and fixes the rural cutoff, as two mutually exclusive overflow lanes on `/v1/ripple-schedule`. This is the routing-server half, shipping dark. The batch and browse wiring is not in this PR.

Full design, including the parts not built here and the open decisions, is in `plans/active/2026-08-14-ripple-fairness-and-rural-access.md`.

## What the measurement said, because it changed the design twice

Measured on live against 169,602 members and 47,664 reaches:

- **Membership is flat** across deprivation quintiles (Q1 20.2%, Q5 19.6%), so the denominator is clean.
- **Post origins skew affluent 2.31x** (Q1 13.2%, Q5 30.5%).
- **Q1 members are reached by ~457 posts per 30 days against ~574 for everyone else**, about 20% under-served.

Then two results that cut against the obvious design:

**The audience cap is not the cause.** Split by whether the cap bound:

| | Q1 share | vs 20.2% population |
|---|---|---|
| Cap bound | 20.7% | parity |
| Ran to the ceiling | 11.9% | 41% short |

So disabling or reweighting the cap would not have improved deprivation equity at all. The deficit lives in the reaches the cap never touches, which are mostly rural-origin, and rural origins skew Q4/Q5 because IMD composite scoring rarely marks rural areas as most-deprived. It is a supply-side geography problem.

**It is a Q1 knee, not a gradient.** Q2 through Q5 sit within about 7% of each other.

## The two lanes

Which one a post gets is decided by whether the headcount cap actually bound. Never both, asserted by test. That split is not tidiness: it puts each mechanism where its problem was measured to exist.

**Rural lane** (`rural_access=1`), for reaches the cap cut short. One ring per density band ceiling (20/30/45 min), sliced from the road-network search the schedule has already run. A member inside the ring for their own band can find the post. This is the argument PR #1308 settled for the time cap ("the cap belongs to the person who would travel, not to the item"), applied to the headcount.

**Fairness lane** (`fairness_weight=W`), for reaches that ran to the ceiling. One extra road-network search out to the stretched budget, then a ring per eligible quintile. `fairness_max_quintile` defaults to **1**, because the data shows a Q1 knee and because Q1-only needs one polygon rather than four. Pass a higher value for the full linear gradient: who deserves extra reach is a values judgement, so it is a parameter rather than baked in.

## Reading the fairness lane without storing anything about anybody

Two endpoints expose the IMD lookup this server already holds: `GET /v1/quintile` for one point, `POST /v1/quintiles` for a set. Measured at 0.6ms, and the single form independently returns quintile 4 and quintile 1 for the two locations in the worked case, matching a separate analysis of the LSOA file.

The batch form is what makes the lane readable. The ring says who is close enough under the stretched budget, which is a containment test the database answers; the fifth says which of those people the stretch was actually FOR. Only the people the ring **adds** need classifying, so the batch is bounded by the extra audience rather than by group size.

An earlier design had the batch stamp a fifth onto each member instead. This replaces it, and is better on every axis that matters: nothing else in the estate has to load, understand or retain IMD data; no inferred deprivation attribute is written against an individual anywhere; and there is no backfill to run or keep current.

It answers **positionally**, and the tests hold it to that. A caller matches answers back to people by index, so a short, reordered or padded array would attribute one member's deprivation to another. An oversized batch is refused rather than truncated for the same reason, and with no index loaded it answers unknown for every point rather than a shorter array, so alignment holds even in the degraded case. An empty batch is an empty answer, not an error, because "the ring added nobody" is a normal outcome.

## Cost, because the design spec called this free and it is not

The road-network search is already paid for. The polygons are not, and the batch sends `polygons=0` precisely because polygon building dominates this endpoint. Measured on the UK graph from the Birmingham origin:

| Ring | Build time |
|---|---|
| 20 min | 0.15s |
| 30 min | 0.8s |
| 45 min | 2.5s |
| 67 min (fairness W=0.5) | 9.4s |
| 90 min (fairness W=1.0) | 20.3s |

Two things keep the rural lane in proportion: only cap-bound posts get rings, and any band whose ceiling is already inside the committed reach is skipped, because it can admit nobody. For the measured case, capped at 28 minutes, that drops the dense ring.

The fairness lane is the genuinely expensive one, and that table is the strongest practical argument for starting at a low weight.

## The Mod-John case

Post 120838351, Frankley B45, IMD quintile 1. Mod-John, DY11 5DT, quintile 4, sparse band, slider already at the 45-minute maximum, 31.4 minutes away by road. The reach stops at 28.0 minutes on exactly 4,000 members.

He gets the **rural** lane, not the fairness one, and that is correct: he was excluded by a headcount, not by deprivation. He clears the sparse ring at 31.4 against 45 minutes. Nothing about the 4,000 people mailed changes.

## Code Quality Review

- The two lanes are mutually exclusive at the branch, so no post pays for both.
- The `[0,1]` weight clamp is now shared with `/v1/fairness` rather than duplicated, so the two cannot drift.
- The fairness rings are built from all nodes within a quintile's budget, not only nodes whose own LSOA carries that quintile. A ring has to be an area a member can stand in; `/v1/fairness` does the latter and consequently returns thousands of disconnected islands, which is fine for its exploratory purpose and wrong for an admission test.

Three test traps caught while writing this:

1. The ring-nesting assertion passed by **equality**, because the Bristol fixture is entirely reachable inside 20 minutes, so it would have passed had the ceiling been ignored altogether. It now also exercises ceilings the fixture can tell apart (2/5/10 min), where ring area demonstrably grows.
2. These are the first tests to reach `handleRippleSchedule` over HTTP, which opens the shared groups DB and left it open. A later JWT test then did a real session lookup instead of skipping validation and failed on a session nobody seeded. The connection is now restored by the test that caused it.
3. `/v1/ripple-schedule` could not be tested through HTTP at all before this, because the test harness passes an empty spatial URL and the handler exits early. It now runs against a stubbed spatial server, so the cap can actually bind.

## Test Plan

- `iznik-routing-go`: the full routing suite passes, including 19 new across both lanes, the two deprivation endpoints and the shared clamp.
- Dark-ship proof: the response is byte-identical with both flags off, asserted directly rather than argued.
- Costs above measured against the real UK graph in the rebuilt container, not estimated.

## Not in this PR

The batch wiring - config flags, `ReachService` params, the `rippling_reach` storage column and the newly-reached mail for both lanes - is #1354. The band each member is in, which the rural lane reads, is #1353.

Still unwired anywhere: the browse feed and the reply-eligibility check. Browse needs a decision rather than the same wiring, because every row there is a different post with a different ring, so a polygon parse per candidate row is unavoidable - and the browse feed is code every request goes through, which is exactly why the cheap outer/inner bounds check was put in front of the exact polygon test in the first place. The rings want a cheap check of their own before anyone adds them here.

Also not touched: post-origin skew, which at 2.31x is the largest measured skew in the system and is a giving problem rather than a reach one.


edwh:

PR #1353, opened 2026-08-15
fix(browse): make the band backfill work, fail loudly, and stop decaying

edwh:
`browse:backfill-max-distance` puts each member on their own density band travel-time budget. A member whose radius lookup fails is deliberately left alone rather than given a wrong cap, and left alone means **no band limit at all**. So a broken lookup does not shrink this command's effect, it voids it. It returned SUCCESS regardless.

## The live state this was found in

`BROWSE_TOWN_NEAR_URL` was unset on the production batch host, so every call went to the compose-internal default, which does not resolve there and returned HTTP 000. In a 2,260-member sample, 1,018 lookups failed. (That setting has since been added on the host, separately from this PR: see below.)

Across 202,837 active members:

| Band radius held | Members |
|---|---|
| Nothing stored | 147,891 |
| Unlimited sentinel | 54,951 |
| **A real band radius** | **zero** |

The sentinel holders are the sparse-band members, who short-circuit to it before ever needing a lookup. Everyone else fails and is skipped.

So the per-member density banding is inert in production: every member is effectively on "no limit" inbound. That is the exact failure the banding was introduced to prevent, a city member on a roughly 4.8 mile band receiving posts from up to 45 minutes away. The nightly job reported success throughout.

The classifier itself is fine. Called directly, the endpoint returns `cap_minutes: 20, density_band: "dense", reach_radius_miles: 4.80` for central London, and the neighbour lookup returns 400 members within 0.25 miles there against 9.3 miles in Kidderminster.

Same shape as the `APIV2_DEPRECATED_URL` defect: adjacent key in the same config file, batch job silently doing nothing.

## What this PR changes

Only the loudness. The command now returns FAILURE when at least a quarter of lookups fail **and** at least 20 have failed outright, naming the configured endpoint in the error.

Both conditions are load-bearing:

- A high rate alone would fail the nightly job whenever a single member's lookup failed for an honest reason. That member being left alone is intended, and two existing tests cover it.
- An absolute count alone would never fire on a 200k-member run with scattered failures, which is exactly the case that should stay quiet.

Between the two, a high rate on a small sample warns rather than fails, so a `--limit` run still says something useful.

## Second change: `--missing-only`, a pass cheap enough to schedule

Four places read `browseReachMaxDistance`. This command is the only thing that writes it, and the full version is a walk of every user (~2.9M rows on live), which is why it was never put on a timer.

That leaves the invariant decaying. A member who joins after a run gets no band limit ever, and since posts ripple out to the widest budget on the understanding that each member is held back to their own band, that member sits permanently on "no limit" inbound. The gap reopens the day after every manual pass.

`--missing-only` selects just the members with nothing holding them to a band, neither the default nor a choice of their own. Everyone else is already consistent or is a reconciliation this pass need not do. Narrowed that way it is small enough to run on a timer, which is what stops this recurring.

Scale, measured on live: 63,133 members currently need a limit, across 18,698 distinct locations. Once that is cleared, the ongoing `--missing-only` population is just new joiners.

Both passes are scheduled, below.

## Already done on the host, outside this PR

`BROWSE_TOWN_NEAR_URL=https://api.ilovefreegle.org/apiv2/town/near` has been added to `.env.background` on the batch host (the previous file is backed up alongside it), and `batch-prod` recreated to pick it up. Lookup failures went from 1,018 in 2,260 to **zero**, and bands now discriminate properly: a dense member gets a 6.9 mile radius, a medium one 21.7.

The full pass is running there now. 63,133 members need a limit. It only ever moves a member towards the right value, so stopping it part-way and running it again is safe.

Worth knowing when reviewing: completing that pass **narrows** feeds. A city member goes from seeing posts up to 45 minutes away to roughly 5 miles. That is the designed behaviour and the reason the banding exists, but it is a visible change as it rolls through.

## Third change: both passes are now scheduled

Left unscheduled the invariant decays, which is how 202,837 members came to hold not one band radius between them.

- `--missing-only` **daily at 04:20**: after the initial backfill this is just people who have joined since, so no new member is ever left uncovered.
- The full pass **monthly, 02:40 on the 1st**: it also reconciles members who already have a value, which `--missing-only` never does. It cannot follow someone who moves from a village to a city, nor an area that has grown denser since its members were measured. Monthly and off-peak because it walks every user with a routing call per distinct location.

Two defects in this change were caught by tests rather than by production:

**`withoutOverlapping(1440)`** is exactly Laravel's 24h default, which `SchedulerResilienceTest` requires you to beat: after the 2026-07-02 incident a killed job's lock must self-heal in minutes, not a day. Now 720, still well over the roughly six hours the pass actually takes.

**`['--missing-only' => true]` compiles to `--missing-only='1'`**, which Symfony rejects outright for a flag that takes no value. The job would have errored every night while `schedule:list` displayed it looking perfectly healthy. Passed positionally now, and the exact scheduled form verified by running it.

That second one was invisible from the schedule listing, so this adds a test that parses **every** scheduled command's arguments against its own definition. 140 commands are scheduled and nothing checked they were runnable as written. It found a bug in its own first draft as well: `ArgvInput` drops its first element as the script name, so the command name has to go there rather than being passed as an argument.

## Documentation

Both pages the project's freshness check flags for these paths are updated, not date-bumped.

`rippling-algorithm.md` section 7 described this command as the thing that materialises each member's band limit without saying what happens when it cannot reach its endpoint or is never run. Both were live for weeks and neither was visible, so both failure modes are now recorded, along with the requirement that `batch-prod` sets `BROWSE_TOWN_NEAR_URL`.

`01-architecture.md` said members are admitted on the budget their own density justifies, which reads as though it is computed per request. It is not: it reads a value written by a scheduled pass, so the design has a moving part outside the request path, and when that pass stops working every member silently reverts to no limit.

## Also records the band it measured

`settings.browseDensityBand`, alongside the budget derived from it.

The budget cannot be read backwards to recover the band. An explicit choice is rescaled
*within* the band, so 20 minutes means either a dense member sitting on their cap or a sparse
member who asked for less, and after the rescale the two are the same number. Anything that
admits a member against something chosen per band needs the band itself, and this command is
the only thing that measures it.

**It is stamped for members who need no correction as well, and that is the point.** Those
members are the majority - a reconciling pass exists precisely because most rows are already
right - and they are exactly the rows a correcting pass never writes to. Stamping only
alongside a correction would leave the band missing for the bulk of the membership while every
run reported success: the same shape as the two failure modes this PR already alarms on.

They cost one write each and then stop appearing, because the stamp is compared before it is
written. A member who moves, or an area that grows denser, is re-stamped by the monthly full
pass.

`band_stamped` joins the other successes in the lookup-failure denominator. Left out, it would
shrink as members moved into it and a healthy run's failure rate would climb towards the alarm
on its own.

Not a new privacy surface: it says the member's area is dense, medium or sparse, which the
budget already stored beside it implies.

## Not in this PR

Whether the audience cap and the band admission should interact differently, which is a separate design.

## Test Plan

`iznik-batch`: 32 backfill tests pass (10 new), plus 12 scheduler tests (1 new). The full suite is green (5,598).

The two failure-path tests use the same fake and the same unreachable endpoint and differ only in sample size: 25 members fails the run, 1 member does not. So the guard demonstrably fires at scale and the floor demonstrably protects the small case, rather than either being argued.

The `--missing-only` tests assert both directions: the flag gives a limit to the member who has none and leaves both the already-defaulted and the self-chosen member untouched, and the full pass still reconciles existing settings, so the narrow flag cannot have leaked into the default and quietly stopped the migration pass working.

The two pre-existing tests covering "a failed lookup leaves that member alone and the run continues" still pass unchanged, which is the specific regression the floor exists to avoid.

The band tests were red-checked rather than assumed: with both writes removed, exactly the three that assert the band is written fail. The other two survive that mutation correctly, because they guard different things - that a band already correct is not rewritten, and that `--dry-run` writes nothing.

### DIFF
--- a/docs/developers/01-architecture.md
+++ b/docs/developers/01-architecture.md
@@ -1,5 +1,5 @@
 ---
-last_reviewed: 2026-08-11
+last_reviewed: 2026-08-15
 owner: Freegle dev team
 covers:
   - docs/developers/reference/architecture.md
@@ -65,6 +65,11 @@ and per-group membership and roles are the ones you will meet first.
   post. Every ripple grows to the same ceiling, and each member is then admitted on the
   budget their own local freegler density justifies - so a rural member can reach the town
   they already drive to, while a city member is not mailed things nobody would travel for.
+  That admission is NOT computed per request: it reads a value written onto each member by a
+  scheduled batch pass (`browse:backfill-max-distance`). So the design has a moving part
+  outside the request path, and if that pass stops working every member silently reverts to
+  no limit at all - which is what happened between 11 Aug and 15 Aug 2026. See section 7 of
+  the algorithm reference.
 - **Getting a first reply in** sits alongside rippling and attacks the 44% of rippled posts
   that get no reply at all: a passthrough for a silent post's first reply, individual mail to
   the members who have asked for that specific item (an open post of the opposite type, or a
--- a/docs/developers/reference/rippling-algorithm.md
+++ b/docs/developers/reference/rippling-algorithm.md
@@ -398,6 +398,16 @@ A community switching ripple-out off retracts the same way - see §4a.
   live only 2,900 of 121,000 recently-active members had ever set one, so without that pass
   the wider ripple would reach a city member with everything inside 45 minutes of a post.
 
+  The same pass records the band NAME in `settings.browseDensityBand`, because the budget it
+  derives cannot be read backwards to recover it: an explicit choice is rescaled *within* the
+  band, so 20 minutes means either a dense member on their cap or a sparse member who asked
+  for less, and afterwards the two are the same number. Anything that has to admit a member
+  against something chosen per band - the rural overflow lane, when enabled, picks one ring
+  per band - needs the band itself. It is stamped for members whose budget needs no correction as well,
+  which is most of them: a value written only alongside a correction would be missing for the
+  bulk of the membership, which is the same shape of silent near-inertness as the two failure
+  modes below.
+
   **That pass is the single writer of `browseReachMaxDistance`, and it has two failure modes
   worth knowing, because both were live for weeks and neither was visible.**
 
--- a/docs/developers/reference/rippling-algorithm.md
+++ b/docs/developers/reference/rippling-algorithm.md
@@ -1,5 +1,5 @@
 ---
-last_reviewed: 2026-08-12
+last_reviewed: 2026-08-15
 covers:
   - iznik-batch/app/Services/Ripple/**
   - iznik-batch/app/Console/Commands/Ripple/**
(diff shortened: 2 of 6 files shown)


edwh:

PR #1354, opened 2026-08-15
feat(ripple): the two overflow lanes, wired and stored, and the rural one reaches the mail

edwh:
The batch side of the two overflow lanes: the settings that ask the routing server for them. All off. Pairs with #1351, which added the rings themselves.

## The settings

| Setting | Default | What it does |
|---|---|---|
| `RIPPLE_RURAL_ACCESS_ENABLED` | off | One ring per density band alongside a capped reach |
| `RIPPLE_FAIRNESS_ENABLED` | off | Stretch the budget for deprived recipients |
| `RIPPLE_FAIRNESS_WEIGHT` | 0 | How much stretch, 0 to 1 |
| `RIPPLE_FAIRNESS_MAX_QUINTILE` | 1 | How far down the deprivation scale it reaches |

## Decisions worth knowing about

**Both lanes are offered together when configured.** Which one applies is per-post and depends on whether the audience cap actually bound, which only the routing server knows once it has counted members. Asking for both costs nothing when neither applies, and the routing server enforces that a post never gets both.

**Weight zero sends nothing at all**, rather than sending a stretch of zero. Zero is the deliberate first rollout step: flag on, plumbing exercised end to end, no behaviour change. If it sent the parameter, the routing server would run the extra road-network search for a stretch of precisely nothing.

**The default reaches only the most deprived fifth.** Measured on live, the shortfall is a knee there: those members are reached by about 457 posts per 30 days against roughly 574 for every other fifth, and the other four sit within about 7% of each other. A straight-line stretch across all five would spend most of its effect where there is no shortfall. It is also four times cheaper, needing one traced ring rather than four. Raising it buys the full gradient at proportionate cost.

**Weight is clamped here as well as in the routing server.** A misconfigured 5 silently becoming 1 at the far end is harder to explain than one that never leaves.

## Storage

`rippling_reach.overflow_bounds`, plus the parsing that fills it and the writes that store it. Nullable and dark: nothing writes it until a lane is enabled. Ships with the production SQL twin the project requires, written so that running it a second time changes nothing.

JSON rather than seven geometry columns, because these are read only in the fallback branch and never on the hot indexed path `polygon`/`outer_bound`/`inner_bound` serve, so they want no spatial index. Same convention as `reachable_group_ids` on this table.

NULL means "no lane applied", never "a lane that produced nothing" — a row read back as an empty array would look like a lane admitting nobody rather than no lane at all, and a test pins that distinction.

**Two write paths, not three, and the third is deliberate.** The init INSERT writes the rings, and `recomputeReach`'s shrink UPDATE rewrites them because it genuinely re-derives the schedule. `advanceDue` does not: it only moves the tick pointer along a schedule already stored, and the rings belong to the reach as a whole rather than to any tick.

**The trap was elsewhere, and it explains an existing defect.** `density_band` is NULL on ~89% of rows, and the cause is not a missed write path: it is **schedule reuse**. When a co-located post already has a schedule, the reuse branch rebuilds it from ticks, total and max-drive alone and drops everything else — and reuse is commonest exactly where posts cluster, which is why the NULLs dominate. The rings are now carried across a reuse.

That needs its own staleness guard, beside the existing reach-budget one: a fairness ring computed under a **different weight** is not this post's ring, however co-located the two posts are. Without it, moving the weight from 0.25 to 0.5 would silently leave old rings in place wherever posts cluster.

## Test Plan

`iznik-batch`: the full suite is green (5,600). Ripple 336 pass (9 new), digest 100 pass (4 new).

The set is self-checking rather than one-sided: "omits when off" and "sends when on" exercise the same code path with opposite expectations, and the weight is bracketed at 0 (sends nothing), 0.5 (sends 0.5) and 5 (clamped to 1). A clamp stuck at either end would fail one of those three.

## Notifications: the rural lane now reaches the mail

`mailNewlyReachedForPost` admits a member either by the committed reach polygon or by the ring for their own density band. A member who cannot be TOLD about a post is barely less shut out than one who cannot see it, which is the case this lane exists for.

**The rings are read once and bound as parameters**, not extracted inside the query. `rippling_reach` is one row per post, so an `ST_GeomFromText` over a `JSON_EXTRACT` in the `WHERE` would re-parse the same polygon for every candidate member: tens of thousands of identical parses on a large group, in the mail path. Binding the band name for an equality test rather than building a `'$.rural.<band>'` path also keeps a member's stored settings value data rather than syntax.

**With no rings the clause is absent, not false.** The query compiles to exactly what it did before this existed and costs exactly what it cost, so a disabled lane cannot slow the mail path down.

The duplicated `mylocation`-else-`lastlocation` resolution is collapsed to one definition while here. It was already written twice and the overflow branch needed it three more times; one definition is also the property worth having, since a member cannot then be admitted on one resolution and measured from another.

A config comment is corrected rather than left contradicting the code: this lane's block said in as many words that it "adds nobody to the mail". True when written. It now records what still holds, which is that the mail path stays members-only, so this reaches people who have already joined a group the post is on and never a cold recipient.

Four tests run ONE setup - a member outside the capped reach but inside the sparse ring - through both settings of the flag, so neither result can come from the member simply being unmailable. A dense-band member inside the sparse ring and a member with no band recorded must both stay out. The second matters: "absent" has to mean "not eligible" rather than "matches anything", or switching this on would widen the mail for the whole membership at once.

## Depends on #1353

These lanes admit a member on their own band, so the read side needs to know which band each member is in - and nothing stored it. The budget cannot stand in for it, because an explicit choice is rescaled *within* the band, so 20 minutes means either a dense member on their cap or a sparse member who asked for less.

#1353 now records `settings.browseDensityBand`, including for members whose budget needs no correction, who are the majority. Until that has run in production this lane admits nobody extra, which is the correct behaviour while it is off but worth knowing before switching it on: run the full backfill pass once, rather than waiting for the monthly one.

## Not in this PR

The other two read paths, each of which needs a decision rather than the same wiring.

**Browse.** Unlike the digest, every row there is a different post with a different ring, so a polygon parse per candidate row is unavoidable - and that is the hot feed path the sandwich-bounds prefilter exists to protect, after the unprefiltered form cost ~849ms a call on the write node. Its SQL shape is also pinned by golden tests. It wants a cheap bounds-style prefilter of its own, not an OR branch bolted onto the exact-polygon test.

**Reply eligibility.** `ReachQueryService::isWithinReach(msgid, lat, lng)` does not take the member, so the band has to be threaded through its callers before the branch can exist.

**The fairness lane's read side**, which needs each member's deprivation fifth stored somewhere. There is no deprivation data in MySQL at all: it lives only in the routing server, loaded from CSV. Attaching it to the shared `locations` table as public area data does not work either, because the point that decides eligibility comes from a free-form `mylocation` for 53,378 of 121,471 recently active members. A server-side table keyed by member, written by the same backfill pass and never returned by any API, is the shape that fits - deliberately not `users.settings`, which is echoed to the browser.

The prerequisite for that read side is now in place. These lanes admit a member on their own travel budget, and until 15 August that budget was unset for every one of 202,837 active members, because `BROWSE_TOWN_NEAR_URL` was missing on the batch host. That is fixed and the backfill has run: 47,651 members now hold a real band limit where none did before. See #1353, which also makes that failure loud in future.

### DIFF
--- a/iznik-batch/app/Services/Ripple/ExpandService.php
+++ b/iznik-batch/app/Services/Ripple/ExpandService.php
@@ -72,6 +72,40 @@ class ExpandService
     }
 
     /** Has the density-sizing migration run? Without it the cap still applies, unrecorded. */
+    /**
+     * Is rippling_reach.overflow_bounds present? Mirrors densityColumnsReady so the overflow
+     * lanes can ship before the migration has run everywhere.
+     *
+     * Unlike density_band, this is consulted by ALL THREE write paths (init INSERT, advanceDue
+     * UPDATE, recomputeReach shrink UPDATE). density_band was wired into only the first, which
+     * is why it is NULL on ~89% of rows.
+     */
+    private static ?bool $overflowColumn = null;
+
+    private function overflowColumnReady(): bool
+    {
+        if (self::$overflowColumn === null) {
+            try {
+                self::$overflowColumn = Schema::hasColumn('rippling_reach', 'overflow_bounds');
+            } catch (\Throwable) {
+                self::$overflowColumn = false;
+            }
+        }
+
+        return self::$overflowColumn;
+    }
+
+    /**
+     * The overflow rings for a schedule, as JSON for storage, or null when no lane applied.
+     * One place, so all three write paths encode identically.
+     */
+    private function overflowJson(?array $schedule): ?string
+    {
+        $bounds = $schedule['overflow_bounds'] ?? null;
+
+        return is_array($bounds) && ! empty($bounds) ? json_encode($bounds) : null;
+    }
+
     private function densityColumnsReady(): bool
     {
         if (self::$densityColumns === null) {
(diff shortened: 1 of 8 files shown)


edwh:

PR #1365, opened 2026-08-17
The mail alarm could not see anyone a week behind, nor mail that never arrived

edwh:
The alarm that watches whether the daily digest kept up cannot see the people it most needs to report.

## What I found

The check counts members who did not get today's digest, but only those whose last digest was within the past seven days. That window was there to leave out people who are permanently inactive and never get a digest anyway.

It also leaves out anyone who has fallen more than a week behind, which is exactly the group worth knowing about.

Measured on production on 17 August, among members who are still using the site, are not bouncing, and still have a community set to a daily digest:

| How far behind | Members | Visible to the alarm |
|---|---|---|
| Within a week | 3,627 | yes |
| Over a week | 2,003 | **no** |
| ...of those, over a month | 384 | **no** |

So 2,003 members want a daily digest, are active, and have not had one for over a week. None of them appeared in the number anyone was watching.

## And it could not see mail that never arrived

Lag measures whether we sent. It cannot tell you whether anyone received.

We hand mail to a smarthost and it accepts everything, so a provider that starts binning our mail leaves every signal we have looking healthy. The spool drains normally. The send log shows one attempt per email and the same ~32 minute spool-to-send delay for every domain alike. The lag figures above are untouched.

On 16 August every Yahoo-run domain stopped delivering:

| Domain | Open rate before | Now | Sent per day |
|---|---|---|---|
| yahoo.co.uk | 24.9% | 0.0% | 15,095 |
| yahoo.com | 19.7% | 0.0% | 11,537 |
| aol.com | 28.4% | 0.0% | 3,972 |
| sky.com | 36.2% | 0.1% | 2,447 |
| ymail.com | 15.9% | 0.0% | 978 |
| rocketmail.com | 27.9% | 0.0% | 822 |

About 35,000 emails a day. The daily series sits at 30-35% until 13 August, falls to 17.3% on the 14th and 15.8% on the 15th, and reaches 0.1% on the 16th, where it has stayed. On 14 August we sent five times the normal daily volume, which is the obvious suspect. Gmail, Hotmail and BT are unaffected throughout, so this is not something we did to the mail itself.

Two days, and nothing said anything.

## What this adds

A per-domain delivery check in the same 13:00 slot, logged as its own line rather than folded into the lag one, so the two stay two Sentry issues. "We are short of capacity" and "a provider has stopped taking our mail" have nothing to do with each other and want different people.

The only signal we have that mail arrived is that somebody opened it. So the check compares each domain's recent open rate against that domain's own earlier baseline. Not against a fixed rate, and not against other domains: genuine rates run from 19% at yahoo.com to 56% at icloud.com on the same mail, so any absolute threshold would either miss real outages at the top of that range or cry wolf at the bottom.

Comparing a domain with itself also disposes of forwarders without naming them. `user.trashnothing.com` passes our mail on, so our pixel is never fetched from the real recipient and its rate is permanently zero - but so is its baseline, so it never registers as having lost anything.

The recent window stops six hours short of now, because mail sent ten minutes ago is unopened for reasons that have nothing to do with delivery.

Run against production today, it flags those six domains and nothing else.

The comment on this check says the backlog rotates fairly and nobody is permanently starved. Looking at how far behind people are, that is not what is happening: the backlog piles up at five and six days and then falls off the edge of what gets counted. Some people have been waiting over a month.

## What this changes

The seven-day window is gone. Dormant members are excluded by asking whether they have used the site in the last month, rather than by how long ago we last managed to mail them — so the alarm no longer loses sight of someone precisely because the problem got worse.

It also now reports how much of the backlog is more than a week old, separately. That is the number that tells you whether the lag is rotating or settling on the same people.

Run against production today, the corrected check reports 6,758 lagging, of which 3,099 are over a week behind. The old form reported 6,653 and could not break it down.

## What this does not do

Nothing here makes the digest faster or reaches anybody sooner. It only stops the measurement hiding the problem. The underlying capacity shortfall is what the night-precompute plan's item C was written for, and these figures suggest it is worth more than the headline number implied.

Nor does it get us unblocked at Yahoo. That is a deliverability conversation, and the volume spike on 14 August is where I would start it. This only means we find out next time within a day instead of never.

## Testing

Full Laravel suite locally.

The delivery check is a service with unit tests: a collapsed domain is flagged, a steady one is not, a permanently-zero forwarder is not, a domain under the volume floor is not, and a pile of too-recent mail does not drag a healthy domain under.

The lag check is a closure registered on the scheduler and had no test before this; the change there is to which rows are counted, and the figures above are the before-and-after run against production data.

### DIFF
--- a/iznik-batch/app/Services/Mail/DeliveryHealthService.php
+++ b/iznik-batch/app/Services/Mail/DeliveryHealthService.php
@@ -1,0 +1,116 @@
+<?php
+
+namespace App\Services\Mail;
+
+use Carbon\Carbon;
+use Illuminate\Support\Facades\DB;
+
+/**
+ * Per-domain delivery health, from open tracking.
+ *
+ * We hand mail to a smarthost and it accepts everything, so nothing in the spool tells us
+ * whether a receiving provider actually delivered it. When a big provider starts refusing or
+ * silently binning our mail, the spool looks perfectly healthy and the digest lag figures look
+ * perfectly healthy, because from our side the mail went out.
+ *
+ * The one signal we do have is that people stop opening. email_tracking records sent_at and
+ * opened_at per recipient, so an abrupt collapse in a domain's open rate is a proxy for its
+ * mail no longer arriving.
+ *
+ * This is a comparison against each domain's OWN recent history, not against a fixed rate or
+ * against other domains. Open rates differ enormously and legitimately between providers -
+ * icloud.com opens at 56% and gmail.com at 23% on the same mail - so an absolute threshold
+ * would either miss real outages at the high end or cry wolf at the low end. It also means
+ * domains that never report opens at all (user.trashnothing.com forwards our mail on, so the
+ * pixel is never fetched from the real recipient) drop out on their own: their baseline is
+ * already zero, so there is nothing to lose.
+ *
+ * The recent window stops short of now. An email sent ten minutes ago has not had a fair chance
+ * to be opened, and including it would make every run look like the start of an outage.
+ *
+ * COST. One pass over email_tracking's sent_at range, which at the default windows is about 4.5
+ * million rows on production, grouped down to a few thousand domains. That is fine once a day
+ * and would not be fine any more often than that.
+ */
+class DeliveryHealthService
+{
+    /** Ignore the last few hours - mail that recent has not had a fair chance to be opened. */
+    public const SETTLE_HOURS = 6;
+
+    /** A domain needs this many sends in the recent window before its rate means anything. */
+    public const MIN_RECENT_SENDS = 500;
+
+    /** ...and this open rate in its own baseline, or there is no working delivery to have lost. */
+    public const MIN_BASELINE_OPEN_PERCENT = 5.0;
+
+    /** Flag when the recent rate falls to this fraction of the domain's own baseline. */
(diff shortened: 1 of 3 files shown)


edwh:
Superseded by #1369, which carries both commits from this branch unchanged (`e2ca626d9` and `407aab5fa`, cherry-picked as `90eb51f28` and `95eb03a75`).

They belong with that work rather than beside it. #1369 reads the relay's queue, which catches a provider **refusing** us. This branch's open-rate check catches the opposite failure: a provider that **accepts** everything and then quietly bins it, leaving no queue entry at all. Neither can see the other's failure, so keeping them apart would have left a reviewer looking at half the picture each time. The lag-alarm fix belongs there too, since it is the measurement that should have made the whole episode visible in the first place.

Both are cross-referenced in the code and written up together in `docs/developers/reference/mail-deferrals.md`.

Nothing here is dropped. Safe to close once #1369 lands.

edwh:
Closing: verified fully carried into #1369 before closing. All 251 lines this branch added to master are present there, none missing, across all three files.

The two commits are cherry-picked as `90eb51f28` and `95eb03a75`. The only difference is seven added lines of comment on `DeliveryHealthService`, cross-referencing the queue-reading half so a future reader can see the two cover opposite failures and does not add a third detector for one already handled.

PR #1366, opened 2026-08-17
fix(modtools): say why Pending is empty when a remembered group filter hides the work

edwh:
Fixes the problem behind [Can't moderate on desktop](https://discourse.ilovefreegle.org/t/cant-moderate-on-desktop/10037), where a moderator had to use her phone because the Pending page showed nothing while the menu said there was work. A second moderator said the same thing had caught him the day before.

## What was actually happening

Established from production logs and the live database, not from reading the code:

- **40** calls to `/apiv2/modtools/messages?collection=Pending&groupid=522709` returned no messages, across the whole day. Every call **without** a groupid returned one to four. All of them ran in 3 to 7 ms with no errors, so nothing was failing.
- The client log shows `Page view: /messages/pending` followed **2 ms later** by `Page view: /messages/pending/522709`. That is the remembered community being re-applied on mount and `watch(groupid)` rewriting the URL.
- Her session was reporting `work: {"pending": 3}` at the same moments, and the menu badge counts every community.
- The database confirms it: every post she approved that day was on Chester, Warrington, Leigh or St Helens. Skelmersdale only ever received rippled-in Approved copies, never a pending one.
- At 17:18:01 she changed the dropdown. From then on `/messages/pending` stopped redirecting, the fetch returned 3 messages, and she approved them at 17:18:41.

So the page was filtered to one quiet community the whole time. `ModGroupSelect` stores the choice in local storage and re-applies it on every mount, and the guard that restores it is `if (val && ...)`, which is why choosing "All my communities" (value 0) sticks permanently and looks like a fix. Her phone worked because local storage is per device.

On the day of the report, **two of 126 active moderators** were in this state: the reporter, and another who moderates 58 communities but was pinned to one.

## Summary

- The empty-page notice now names the community it is filtered to, gives the outstanding count, and offers a **Show all my communities** button. The count is `pending + spam` read from `authStore.work`, which is exactly what the red Pending badge counts and exactly what the button will reveal. Deliberately not the composable's `work` (`pending + pendingother`): `pendingother` covers posts on **backup** communities, and the all-communities listing only fans out over active ones by design (`user.GetActiveModGroupIDs`), so counting those would promise work the button cannot surface. The button clears the remembered choice as well as the current filter, because `ModGroupSelect` only persists a change made through the dropdown itself, so setting `groupid` alone would let the filter come straight back next visit.
- **The queue now refreshes once per work change, not once per caller.** `setupModMessages()` warns "Do not include any watch in here as a separate watch is called for each time setupModMessages() is called" and then registers `watch(workdetail)`, a `MutationObserver` and `watch(modtoolsediting)` on every call. Each `watch(workdetail)` runs a full clear-then-refetch. It is called by three pages, by `ModMessages.vue`, and by `useKeywords.js` from a **module-level computed** that depends on the shared `group` ref, so that one re-registers every time the moderator changes community. Measured over one day of production traffic (1531 bursts, 126 moderators), a single work-count tick fired:

  | distinct group filters that moderator used | mean identical listing requests |
  |---|---|
  | 1 | 2.07 |
  | 2 | 5.79 |
  | 3 | 8.80 |
  | 7 | 12.00 |

  Worst case 13. Burst size does *not* grow with messages on screen, so it is not one per rendered message; it tracks group switching. Only the page owns the queue and only the page passes `reset=true`, so the refresh machinery is registered there alone. Verified in a real browser: a work change that previously fired two listing requests now fires one, and the new post still appears with no reload.
- An exception from `fetchMessagesMT` used to escape `loadMore()` entirely, leaving `InfiniteLoading` in `'loading'` and `busy` stuck true. Since `busy` suppresses the empty notice, the page looked like it was still working when it had given up. It is now caught and shown with a **Try again** button. An expired session stays silent, because the layout already handles that.
- Server side, `ListMessagesMT` read its listing query with `Pluck` and never checked the error. A query aborted by `MAX_EXECUTION_TIME(20000)` therefore returned an ordinary 200 with an empty messages array, which the client cannot tell apart from an empty queue. That is the same symptom with no filter involved, and the all-communities form is far more exposed to it because it unions one branch per community. It now returns a 500.

## Code quality review

- The remembered-value key is built from a single `REMEMBER_KEY` constant used for both the `ModGroupSelect` prop and the clear, so the two cannot drift apart.
- The loader bump is guarded on an empty list, so a healthy page does not refetch every time a count ticks.
- The Go change captures the error from every `Pluck` in the handler, not just the one on the standard listing path, so the search and Edit branches are covered too. The zero-communities case still short-circuits before any query is built, so a moderator of no communities is unaffected.
- The doc page this behaviour belongs to is updated in the same change, as required by the freshness check.

## Future improvements

Not done here, deliberately:

- `useKeywords.js` still calls `setupModMessages()` from a module-level computed. That is now harmless (it registers nothing), but it is an odd place to call a setup function from and worth tidying separately.
- The reporter said the dropdown read "all my communities". The code, a component test against the real `b-form-select`, and the browser check below all say it displays the filtered community's name. No display and value desync could be reproduced, so this is recorded as unexplained rather than fixed.

## Test plan

- `tests/unit/pages/modtools/messages/pending.spec.js`: 13 tests to 20, seven new -
  names the filtered community and the outstanding count when the filter is hiding work;
  counts spam towards that total, as the menu badge does;
  ignores backup-community work, which all-communities does not show;
  keeps the plain empty notice when nothing is pending anywhere;
  keeps it when already showing all communities;
  clearing the filter also forgets it, so the next visit is not filtered again;
  and the page fires no fetch of its own on a work-count change.
  The auth-store mock in this spec is now `reactive()` so the count-driven computed actually fires. Each behaviour test was confirmed to fail before the change.
- `tests/unit/composables/useModMessages.duplicateWatchers.spec.js`: 2 tests, new. Asserts one refetch per work change with the page, its list component and `useKeywords` all having called setup; fails on the old code with 3. Each case runs in its own `effectScope` so a leaked watcher cannot bleed into the next test.
- `iznik-server-go/test/modtools_listing_query_error_test.go`: forces a real query error with a cancelled context on `database.DBConn` and asserts the response is not a success carrying a messages array. **Red proof**: with the guard removed the full Go suite was 4145 pass / 1 fail, and the single failure was this test, reporting "got 0 messages, which ModTools would read as an empty queue".
- Full Go suite: **4146 pass, 0 fail**.
- Full vitest suite: **15512 pass, 0 fail** (file hashes in the runner container checked against the host, so the run genuinely covered these changes).
- ESLint and Prettier clean on the changed files. `check-docs-freshness.mjs` green.
- Browser check against `modtools-dev-local` with real data: created one pending post on FreeglePlayground and none on Freegle Playground2, selected the quiet community, and confirmed the menu showed "Pending 1" while the page explained the filter and named the community. Clicking **Show all my communities** returned the URL to `/messages/pending/`, reset the dropdown, and revealed the post. A fresh visit to `/messages/pending` afterwards no longer redirected back to the filtered URL. Test data removed.

## Coveralls

The three red Coveralls checks are not from this branch; the analysis is in a comment below. In short: the -10.3% aggregate is a four-flag build being compared against a one-flag (Go-only) baseline, because CI is path-filtered, and master swings the same way in reverse. The -0.007% Go flag is repo-wide jitter of the same size master shows commit to commit (and master is itself red on that flag for -0.004%); the two statements this PR adds to Go are executed 25 and 1 times respectively in the coverage profile, so patch coverage is 100%. Only `ci/circleci: build-and-test` is required by branch protection, and it passed.

### DIFF
--- a/docs/moderators/02-moderating-posts.md
+++ b/docs/moderators/02-moderating-posts.md
@@ -1,5 +1,5 @@
 ---
-last_reviewed: 2026-08-16
+last_reviewed: 2026-08-17
 owner: Freegle dev team
 covers:
   - iznik-nuxt3/modtools/pages/messages/**
@@ -24,6 +24,13 @@ queues, the actions, why posts get held, and how reports work.
 
 **Messages > Pending** (`/messages/pending`) is where posts waiting for a decision
 appear, for the community (or all your communities) you have selected.
+
+The dropdown at the top of the page remembers what you picked, on that browser, and
+applies it again next time. The **Pending** count in the menu, on the other hand, counts
+posts across every community you moderate. So the two can disagree: if the community you
+picked is quiet, the page is empty while the menu still shows a count. When that happens
+the page says so, names the community it is filtered to, and offers a button to go back
+to all your communities, which also clears the remembered choice.
 
 For each post you can:
 
--- a/iznik-nuxt3/modtools/composables/useKeywords.js
+++ b/iznik-nuxt3/modtools/composables/useKeywords.js
@@ -1,33 +1,20 @@
-import { setupModMessages } from '~/composables/useModMessages'
+import { computed } from 'vue'
+import { group } from '~/composables/useModMessages'
 
-const typeOptions = computed(() => {
-  const { group } = setupModMessages()
-  const thegroup = group.value
-  return [
-    {
-      value: 'Offer',
-      text:
-        thegroup &&
-        thegroup.settings &&
-        thegroup.settings.keywords &&
-        thegroup.settings.keywords.offer
-          ? thegroup.settings.keywords.offer
-          : 'OFFER',
-    },
-    {
-      value: 'Wanted',
-      text:
-        thegroup &&
-        thegroup.settings &&
-        thegroup.settings.keywords &&
-        thegroup.settings.keywords.wanted
-          ? thegroup.settings.keywords.wanted
-          : 'WANTED',
-    },
-  ]
-})
+// Read the selected community straight from the shared ref. This used to be a
+// module-level computed whose getter called setupModMessages() - so a setup
+// function ran from inside a computed, re-registering its watchers on every
+// group change. Nothing here needs setup; it only needs to read the group.
+function keyword(thegroup, which, fallback) {
+  return thegroup?.settings?.keywords?.[which] || fallback
+}
 
 export function setupKeywords() {
+  const typeOptions = computed(() => [
+    { value: 'Offer', text: keyword(group.value, 'offer', 'OFFER') },
+    { value: 'Wanted', text: keyword(group.value, 'wanted', 'WANTED') },
+  ])
+
   return {
     typeOptions,
   }
--- a/iznik-nuxt3/modtools/composables/useModMessages.js
+++ b/iznik-nuxt3/modtools/composables/useModMessages.js
@@ -19,7 +19,10 @@ const summarykey = ref(false)
 const busy = ref(false)
 const context = ref(null)
 const groupid = ref(0)
-const group = ref(null)
+// Exported so read-only consumers (useKeywords) can follow the selected
+// community without calling setupModMessages(), which is a setup function and
+// does setup-shaped things.
+export const group = ref(null)
 const limit = ref(10)
 const workType = ref(null)
 const show = ref(0)
--- a/iznik-nuxt3/modtools/composables/useModMessages.js
+++ b/iznik-nuxt3/modtools/composables/useModMessages.js
@@ -152,7 +152,16 @@ const visibleMessages = computed(() => {
 })
 
 export function setupModMessages(reset) {
-  // Do not include any watch in here as a separate watch is called for each time setupModMessages() is called
+  // The refresh machinery below is registered ONLY for reset=true, i.e. for the
+  // page that owns this queue. Everything here is shared module-level state, so
+  // a watcher registered by a second caller is a duplicate that does the same
+  // clear-then-refetch again. setupModMessages() is called by the page, by
+  // ModMessages.vue, and by useKeywords.js from a MODULE-LEVEL computed that
+  // re-evaluates on every group change - so the duplicates accumulated as a
+  // moderator worked. Measured in production over one day: a single work-count
+  // tick fired 2 identical listing requests for a moderator who used one group
+  // filter, and up to 13 for one who used seven. Registering per call also
+  // leaks when there is no active effect scope to dispose them.
 
   /* watch(group, async (newValue, oldValue) => {
     console.log("===useModMessages watch group", newValue?.id, oldValue?.id, groupid.value)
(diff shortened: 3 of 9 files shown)


edwh:
## The three red Coveralls checks are not this branch

`ci/circleci: build-and-test` passed. Chasing the Coveralls reds rather than waving them through, and neither is caused by this change.

### `coverage/coveralls` and `coverage/coveralls (push)`: -10.3%

The aggregate has two states, roughly 75.6% and 85.9%, and which one a build lands in depends on **how many flag reports that CI run uploaded**, which depends on which paths the commit touched. CI is path-filtered, so a Go-only commit only runs the Go suite. From the `Coveralls webhook` step of each build:

| commit | reports uploaded | aggregate |
|---|---|---|
| master `529b2e242` (Go-only) | 1 | 85.876% (+10.3%, green) |
| master `3f5c1e113` (Go-only) | 1 | 85.872% |
| master `2ea24c767` (full stack) | 4 | no aggregate context posted at all |
| this PR `669c2fc74` (full stack) | 4 | 75.621% (-10.3%, red) |

With one report the other three flags carry forward from the base and Go's 85.9% dominates. With four, the real playwright (21.437%) and laravel (71.955%) numbers land and the aggregate is ~75.6%. So this PR is a four-flag build being compared against a one-flag baseline. The 10.3 points is the shape of the comparison, not lost coverage. Master shows the same swing in the other direction, and one master commit got no aggregate context at all.

All four uploads succeeded on this build (`Upload Go/Laravel/Vitest/Playwright coverage`, exit 0 each), so nothing failed to report.

### `Coveralls - go`: -0.007%

The two statements this PR adds to Go are both executed by the suite. From the coverage profile of a local full run:

```
message/message_list.go:603.2,603.20   1 25   # if listErr != nil {
message/message_list.go:603.20,605.3   1 1    # return fiber.NewError(500, ...)
```

The error return is reached exactly once, by the new test. So patch coverage here is 100% and the decrease is not from these lines.

The profile also contains a duplicate copy of every block with a zero count, which is a known artefact of duplicate module paths in this repo's Go profile. That is the likeliest source of the sub-0.01% drift. Either way it is pre-existing: the same flag moved +0.002%, -0.004% and +0.02% across the three master commits before this branch, and master commit `3f5c1e113` is itself **red** on `Coveralls - go` for -0.004%. The gate fails on any decrease at all.

### Not blocking

Branch protection requires exactly one context:

```
$ gh api repos/Freegle/Iznik/branches/master/protection --jq '.required_status_checks.contexts'
["ci/circleci: build-and-test"]
```

Worth fixing the aggregate comparison separately, since it makes every full-stack PR look like it dropped 10 points. Filing that as its own piece of work rather than bundling it here.

edwh:
### Follow-up: the second CI run proves the Go flag drift directly

`530946bf5` changed two files, both frontend, and **no Go code at all**:

```
$ git show --name-only --format="" 530946bf5 | grep -c '\.go$'
0
```

`Coveralls - go` still moved, from 85.869% to 85.865% (-0.01%). A commit that touches no Go cannot change Go coverage, so that flag is drifting on its own. This is the same phenomenon as the +0.002% / -0.004% / +0.02% walk across the three master commits before this branch, one of which was also red.

The aggregate moved too, 75.621% to 75.627%, still about 10 points below the one-flag baseline it is being compared against.

Everything required is green: `ci/circleci: build-and-test`, `check-runner` and `mark-ci-passed` all pass, and the deploy preview built.

edwh:
### Correction: the partial coverage uploads come from FAILED builds, not path filtering

My earlier comment said CI is path-filtered and that a Go-only commit uploads one report. That is wrong, and worth correcting on the record.

Every suite runs on every branch. `Determine changed paths` in the orb hardcodes all four flags to false:

```
SKIP_GO=false
SKIP_VITEST=false
SKIP_PLAYWRIGHT=false
SKIP_LARAVEL=false
```

with the note that path skipping was removed *because* it broke Coveralls carryforward.

The real correlation is with build outcome. The upload steps are `when: always`, so a build that dies partway still sends whatever completed:

| build | branch | outcome | reports uploaded | aggregate |
|---|---|---|---|---|
| 32370 | master | **failed** | 1 (Go only) | 85.876% |
| 32377 | master | **failed** | 1 (Go only) | 85.872% |
| 32381 | master | success | 4 | - |
| 32393 | this branch | success | 4 | 75.621% |

On 32370 the other three steps logged `⚠️ Not found` — the suites never got far enough to produce artefacts.

This does not change the defect or the fix. `carryforward` exists precisely so a partial upload still reports an honest whole-repo figure, and it is dead because all four uploads share one `COVERALLS_SERVICE_JOB_ID`, so Coveralls registers no parallel jobs (`"jobs":0` on every run). What changes is the trigger: it is a **failed build poisoning the baseline**, not a skipped suite. A failed master build published Go's 85.9% as the whole repo, and the next healthy full build measured the true 75.6% and was reported as -10.3%.

Fix is parked on `fix/coveralls-parallel-job-ids`, not published.

PR #1368, opened 2026-08-18
ci: surface when Coveralls fails to group parallel coverage jobs (cause still open)

edwh:
Coveralls `carryforward` has never worked on this repo, and a failing build silently corrupts the coverage baseline for every build after it. Found while chasing the red Coveralls checks on #1366.

## The defect

The coverage setup step exports one job id for the whole build:

```bash
echo "export COVERALLS_SERVICE_JOB_ID=${CIRCLE_BUILD_NUM}" >> $BASH_ENV
```

so all four flag uploads (`go`, `laravel`, `vitest`, `playwright`) claim to be the **same job**. A Coveralls parallel build needs its jobs to share `service_number` and **differ** in `service_job_id` — `node-coveralls` passes both straight through from the environment (`getOptions.js:157`).

The webhook that finalises the build says so on every run:

```
📊 Uploaded 4 reports. Sending webhook (carryforward=go,laravel,vitest,playwright)...
{"done":true,"url":"https://coveralls.io/builds/81210740","jobs":0}
```

`"jobs":0` — no parallel jobs were grouped, so `carryforward` had nothing to attach to.

## Why it matters

`carryforward` exists so that a build uploading only some flags still reports an honest whole-repo figure. Every suite runs on every branch — path skipping was removed, and `Determine changed paths` hardcodes all four `SKIP_*` to `false` — so a partial upload means the build **failed partway**. The upload steps are `when: always`, so they still send whatever completed, which for an early failure is Go alone.

| build | branch | outcome | reports | aggregate |
|---|---|---|---|---|
| 32370 | master | **failed** | 1 (Go only) | 85.876% |
| 32377 | master | **failed** | 1 (Go only) | 85.872% |
| 32381 | master | success | 4 | — |
| 32393 | #1366 | success | 4 | 75.621% |

On 32370 the other three upload steps logged `⚠️ Not found`.

So a failed master build published Go's 85.9% as the whole repo's coverage, and that became the baseline. The next healthy full build measured the true 75.6% and was reported as **"coverage decreased by 10.3%"** having changed nothing. That is the phantom regression that has been showing up on full-stack PRs.

## The change

- Drop the build-wide `COVERALLS_SERVICE_JOB_ID`; each upload sets its own, `${CIRCLE_BUILD_NUM}-<flag>`, alongside its flag name.
- The webhook step now reads the returned `jobs` count and warns loudly if it is lower than the number of reports uploaded, so this cannot rot unnoticed again.
- Correct the webhook comment, which still described the `SKIP_*` path filtering that no longer exists. That stale comment is what made the failed-build cause hard to see, and it sent me down the wrong path for a while.

Published as `freegle/tests@1.1.380` and pinned in `continue-config.yml`.

## Test plan

- `circleci orb validate` passes.
- Every coverage step extracted from the YAML passes `bash -n`.
- The new `jobs` parsing was exercised against the real payloads, including the empty-response case:
  `{"jobs":0}` → warn, `{"jobs":4}` → OK, `{"done":true, "jobs": 12 }` → OK, empty string → warn.

**What to watch on the first run of this branch**, since the Coveralls side cannot be tested locally:

1. The `Coveralls webhook` step should print `✅ Coveralls grouped 4 parallel job(s)` rather than `"jobs":0`.
2. After this lands, the next master build that **fails** should report an aggregate near the true whole-repo figure (~75.6%) rather than Go's ~85.9%, because the missing flags will carry forward.

If the first run still shows `jobs: 0`, the warning will now say so explicitly rather than the build going quietly green with a wrong number.

### DIFF
--- a/.circleci/continue-config.yml
+++ b/.circleci/continue-config.yml
@@ -53,7 +53,7 @@ orbs:
   # rate-limits (429) the GitHub tarball dependency in the lock file, which
   # npm's own --fetch-retries does not cover, and it red-lined master twice on
   # 2026-08-17 for commits that touched no JavaScript.
-  freegle: freegle/tests@1.1.379
+  freegle: freegle/tests@1.1.380
 
 jobs:
   mark-ci-passed:
--- a/.circleci/continue-config.yml
+++ b/.circleci/continue-config.yml
@@ -53,7 +53,7 @@ orbs:
   # rate-limits (429) the GitHub tarball dependency in the lock file, which
   # npm's own --fetch-retries does not cover, and it red-lined master twice on
   # 2026-08-17 for commits that touched no JavaScript.
-  freegle: freegle/tests@1.1.380
+  freegle: freegle/tests@1.1.381
 
 jobs:
   mark-ci-passed:
--- a/.circleci/orb/freegle-tests.yml
+++ b/.circleci/orb/freegle-tests.yml
@@ -3029,9 +3029,34 @@ jobs:
             elif [ -d ~/FreegleDocker/.git ]; then echo "export COV_ROOT=~/FreegleDocker" >> $BASH_ENV
             else echo "export SKIP_COVERAGE=true" >> $BASH_ENV; fi
             echo "export COVERALLS_SERVICE_NAME=circleci" >> $BASH_ENV
-            echo "export COVERALLS_SERVICE_JOB_ID=${CIRCLE_BUILD_NUM}" >> $BASH_ENV
             echo "export COVERALLS_SERVICE_NUMBER=${CIRCLE_BUILD_NUM}" >> $BASH_ENV
             echo "export COVERALLS_PARALLEL=true" >> $BASH_ENV
+            # Deliberately NOT exporting COVERALLS_SERVICE_JOB_ID here. In a
+            # Coveralls parallel build each flag upload is a separate JOB inside
+            # one build: they share service_number and must differ in
+            # service_job_id (node-coveralls getOptions.js reads both from the
+            # environment). Exporting one id for the whole build made all four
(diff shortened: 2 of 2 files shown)


edwh:
## Negative result: distinct job ids did NOT fix carryforward

First run of `freegle/tests@1.1.380` (build 32411, outcome success):

```
📊 Uploaded 4 reports. Sending webhook (carryforward=go,laravel,vitest,playwright)...
✅ Webhook sent: {"done":true,"url":"https://coveralls.io/builds/81219214","jobs":0}
⚠️ Coveralls grouped 0 parallel job(s) but we uploaded 4 report(s).
```

Still `"jobs":0`, with four distinct `COVERALLS_SERVICE_JOB_ID` values (`<build>-go`, `-laravel`, `-vitest`, `-playwright`). So the shared job id was **not** what stopped Coveralls grouping them, and my stated cause was wrong.

What this run does establish:

- The new warning works. The condition that used to pass silently now says so in the build log, which is the part of this change worth keeping regardless.
- The remaining candidates are narrower: either the uploads are not being accepted as parallel at all (so `COVERALLS_PARALLEL` is not reaching `node-coveralls`, or is being ignored), or the webhook's `build_num` does not match the `service_number` the uploads actually sent, or Coveralls does not support this flow the way the orb assumes.

Next thing to check is what the uploads actually send: `node-coveralls` can print the payload it posts, so capturing `service_number`, `service_job_id` and `parallel` from a real upload would settle it rather than reasoning from the env vars.

I have not reverted the change: distinct ids per flag are more correct than one shared id either way, and the warning is what surfaced this. But the PR title and body overstate what it achieves, so treat this as "adds the diagnostic, cause still open" rather than a fix.

